PR #2, opened 2020-04-17
Add locking to kustomization apply

stefanprodan:
Prevent concurrent apply of the same kustomization using tmp file locking.

### DIFF
--- a/controllers/kustomization_controller.go
+++ b/controllers/kustomization_controller.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"path"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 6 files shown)


controllers/kustomization_controller.go
```diff
@@ -150,10 +169,9 @@ func (r *KustomizationReconciler) sync(
 	}
 
 	// apply kustomization
-	cmd = fmt.Sprintf("cd %s && kubectl apply -f %s.yaml", tmpDir, kustomization.GetName())
+	cmd = fmt.Sprintf("cd %s && kubectl apply -f %s.yaml --timeout=5m", tmpDir, kustomization.GetName())
```
hiddeco, line 172: Do we want this timeout to be configurable in the future? If so, I would add a `// TODO` to keep them visible.

stefanprodan: Going to use the `interval` since overlapping is not desirable.

PR #3, opened 2020-04-17
Implement garbage collection

stefanprodan:
Changes:
- add `GarbageCollect` predicate to `KustomizationReconciler`
- remove the Kubernetes objects previously applied on the cluster, based on `spec.prune` label select
- add garbage collection spec to docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # kustomize-controller
 
+[![e2e](https://github.com/fluxcd/kustomize-controller/workflows/e2e/badge.svg)](https://github.com/fluxcd/kustomize-controller/actions)
+
 The kustomize-controller is a Kubernetes operator that applies kustomizations in-cluster.
 
 ![overview](docs/diagrams/fluxcd-kustomize-source-controllers.png)
@@ -19,7 +21,7 @@ A kustomization object defines the source of Kubernetes manifests by referencing
 the path to the kustomization file, 
 and a label selector used for garbage collection of resources removed from the Git source.
 
-Specification:
+### Specification
 
 ```go
 // KustomizationSpec defines the desired state of a kustomization.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -44,9 +44,26 @@ type KustomizationSpec struct {
...
 
+### Garbage collection
+
+Garbage collection means that the Kubernetes objects that were previously applied on the cluster
+but are missing from the current apply, will be removed. Garbage collection is also performed when a Kustomization
+object is deleted, triggering a removal of all Kubernetes objects previously applied on the cluster.
+
+When garbage collection is enabled, all Kubernetes objects must have a common label that matches the `prune` selector.
+
+For example, `prune: env=dev` requires a kustomization.yaml with:
```
hiddeco, line 59:
```suggestion
For example, `prune: env=dev` requires a `kustomization.yaml` with:
```

api/v1alpha1/kustomization_types.go
```diff
@@ -28,7 +28,7 @@ type KustomizationSpec struct {
 	// +required
 	Path string `json:"path"`
 
-	// Label selector used for prune operations, e.g. env=staging.
+	// Label selector used for garbage collection.
 	// +kubebuilder:validation:Pattern="^.*=.*$"
 	// +optional
 	Prune string `json:"prune,omitempty"`
```
hiddeco, line 34: Is a single label sufficient, or would a `[]string` be better?

stefanprodan: label selectors are not an array but an an expression e.g.:
`prune: "environment in (production),tier in (frontend)"`

stefanprodan: I've added the Kubernetes docs link to the readme.

PR #4, opened 2020-04-17
Add e2e tests

stefanprodan:
End-to-end workflow with Kubernetes Kind:
- build and deploy kustomize-controller
- deploy source-controller
- apply sample kustomizations and git repositories
- check that kustomization apply was successful

### DIFF
--- a/controllers/kustomization_controller.go
+++ b/controllers/kustomization_controller.go
@@ -80,6 +80,17 @@ func (r *KustomizationReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 		return ctrl.Result{}, err
 	}
 
+	if source.GetArtifact() == nil {
+		msg := "Source is not ready"
+		kustomization = kustomizev1.KustomizationNotReady(kustomization, kustomizev1.ArtifactFailedReason, msg)
+		if err := r.Status().Update(ctx, &kustomization); err != nil {
+			log.Error(err, "unable to update Kustomization status")
+			return ctrl.Result{Requeue: true}, err
+		}
+		log.Info(msg)
+		return ctrl.Result{}, nil
+	}
+
 	// try git sync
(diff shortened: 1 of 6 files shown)


controllers/kustomization_controller.go
```diff
@@ -80,6 +80,17 @@ func (r *KustomizationReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
...
 
+	if source.GetArtifact() == nil {
+		msg := "Source is not ready"
+		kustomization = kustomizev1.KustomizationNotReady(kustomization, kustomizev1.ArtifactFailedReason, msg)
+		if err := r.Status().Update(ctx, &kustomization); err != nil {
+			log.Error(err, "unable to update Kustomization status")
+			return ctrl.Result{Requeue: true}, err
+		}
+		log.Info(msg)
+		return ctrl.Result{}, nil
```
hiddeco, line 91: Any specific reason for using `Requeue: true` in the `if`-statement above this but not here?

stefanprodan: Yes, requeueing here is useless since the artifact is not ready, while a status update error could be caused be a transient network error

hiddeco: "Not ready" gives me the idea that it is transient as well, and the resource is still initializing?

stefanprodan: It can take forever

hiddeco: That’s why it uses a exponential backoff strategy?

stefanprodan: Maybe it’s not obvious from these changes but once the git repo is initialized and the artifact is ready, the kustomizion is triggered right away, there is no reason to spam the api server with calls while source-controller is cloning the repo.

hiddeco: Ah, I forgot about that gear wheel. Ack.

PR #50, opened 2020-06-29
Check suspend status before progressing

hiddeco:
When a release is marked as suspended no operations should be performed
for the release except acknowledging the reconciler has noticed it has
been suspended. This means the suspend check should happen _before_ the
Kustomization is marked as progressing (which should be an indicator
reconciliation is enabled for the Kustomization).

### DIFF
--- a/controllers/kustomization_controller.go
+++ b/controllers/kustomization_controller.go
@@ -66,14 +66,8 @@ func (r *KustomizationReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 
 	log := r.Log.WithValues(strings.ToLower(kustomization.Kind), req.NamespacedName)
 
-	kustomization = kustomizev1.KustomizationProgressing(kustomization)
-	if err := r.Status().Update(ctx, &kustomization); err != nil {
-		log.Error(err, "unable to update Kustomization status")
-		return ctrl.Result{Requeue: true}, err
-	}
-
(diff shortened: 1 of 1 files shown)


controllers/kustomization_controller.go
```diff
@@ -66,12 +66,6 @@ func (r *KustomizationReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
...
 	log := r.Log.WithValues(strings.ToLower(kustomization.Kind), req.NamespacedName)
 
-	kustomization = kustomizev1.KustomizationProgressing(kustomization)
-	if err := r.Status().Update(ctx, &kustomization); err != nil {
-		log.Error(err, "unable to update Kustomization status")
-		return ctrl.Result{Requeue: true}, err
-	}
-
 	if kustomization.Spec.Suspend {
 		msg := "Kustomization is suspended, skipping execution"
```
hiddeco, line 70: Maybe

```suggestion
		msg := "Kustomization is suspended, skipping reconciliation"
```

Given the discussion we had about "sync" and "execution" being synonyms of "reconciliation" @stefanprodan .

stefanprodan: Can you please amend the commit with the above change

hiddeco: As you please, done.

Issue #49, opened 2020-06-26
Unit testing


stefanprodan:
Currently the kustomize controller testing is done with Kubernetes Kind and a series of end-to-end tests. We should create unit tests and mock the source controller artifacts server.

The artifacts mock server could be built on top of Go `httptest` with the following features:
- `func new() (port int)`  creates a httptest server and returns the allocated port
- `func artifactFromYaml(path, yaml string) (url string)` creates a tar.gz with the path e.g. `dev/manifests.yaml` and returns the URL in the format `hash(path).tar.gz`

A basic unit test could look like:
- init mock server
- create artifact from an in-memory namespace yaml
- create a GitRepository object with the artifact url set to the one received from the mock server
- create Kustomization object referencing the above source
- start Kustomize reconciler
- wait for the namespace to be created

hiddeco:
Given that the [source-controller](https://github.com/fluxcd/source-controller/) already has [a set of test servers](https://github.com/fluxcd/source-controller/tree/master/internal/testserver), it would likely be best to move the package there from `internal` to `pkg`, and add another mocking server for artifacts there while making use of some of the logic from the [`Storage` utility](https://github.com/fluxcd/source-controller/blob/master/controllers/storage.go#L126).

This way it can be re-used by all other components when they need to mock source artifacts.

PR #56, opened 2020-07-06
Add reconciliation test

stefanprodan:
Changes:
- Setup unit tests with ginkgo
- Add namespace reconciliation test
- Update source-controller to v0.0.2

Fix: #49


Makefile
```diff
@@ -14,6 +14,7 @@ all: manager
 
 # Run tests
 test: generate fmt vet manifests api-docs
+	curl -s https://raw.githubusercontent.com/fluxcd/source-controller/v0.0.2/config/crd/bases/source.fluxcd.io_gitrepositories.yaml > config/crd/bases/gitrepositories.yaml
```
hiddeco, line 17: I would make this a dedicated target and chain it in.

PR #70, opened 2020-07-20
Upgrade CRDs to apiextensions.k8s.io/v1

stefanprodan:
Ref: https://github.com/fluxcd/toolkit/issues/111


.github/workflows/e2e.yaml
```diff
@@ -38,6 +38,7 @@ jobs:
       - name: Check if working tree is dirty
         run: |
           if [[ $(git diff --stat) != '' ]]; then
+            git diff
```
hiddeco, line 41: Not sure if required (as I successfully used it without `--no-pager` on other projects, likely due to no pager being configured on GitHub actions), but the notification-controller PR includes `--no-pager` here.

PR #82, opened 2020-08-17
Make use of dedicated fluxcd/pkg/testserver module

hiddeco:
This includes a dependency change from github/fluxcd/source-controller
to github.com/fluxcd/source-controller/api (v0.0.9) to decrease the
amount of indirect dependencies.


stefanprodan:
@hiddeco can you update source-controller in config and makefile please.

Issue #94, opened 2020-09-01
Add support for Mozilla SOPS decryption


stefanprodan:
Ref: https://github.com/fluxcd/toolkit/discussions/156

PR #95, opened 2020-09-01
Implement Mozilla SOPS decryption

stefanprodan:
This PR implements Mozilla SOPS decryption fix: #94 

### Example

In order to store secrets safely in a public or private Git repository, you can use [Mozilla SOPS](https://github.com/mozilla/sops) and encrypt your Kubernetes Secrets data with OpenPGP keys. 

Generate a GPG key **without passphrase** using [gnupg](https://www.gnupg.org/) then use sops to encrypt a Kubernetes secret:

```sh
sops --pgp=FBC7B9E2A4F9289AC0C1D4843D16CEE4A27381B4 \
--encrypt --encrypted-regex '^(data|stringData)$' --in-place my-secret.yaml
```

Commit and push the encrypted file to Git.

> **Note** that you should encrypt only the `data` section, encrypting the Kubernetes secret
> metadata, kind or apiVersion is not supported by kustomize-controller.

Create a secret in the `gitops-system` namespace with the OpenPGP private key:

```sh
gpg --export-secret-keys --armor FBC7B9E2A4F9289AC0C1D4843D16CEE4A27381B4 |
kubectl -n gitops-system create secret generic sops-gpg \
--from-file=sops.asc=/dev/stdin 
```

Configure decryption by referring the private key secret:

```yaml
apiVersion: kustomize.toolkit.fluxcd.io/v1alpha1
kind: Kustomization
metadata:
  name: my-secrets
spec:
  interval: 5m
  path: "./"
  sourceRef:
    kind: GitRepository
    name: my-secrets
  decryption:
    provider: sops
    secretRef:
      name: sops-pgp
```


docs/spec/v1alpha1/kustomization.md
```diff
@@ -16,6 +16,10 @@ type KustomizationSpec struct {
 	// +optional
 	DependsOn []string `json:"dependsOn,omitempty"`
 
+	// Decrypt Kubernetes secrets before applying them on the cluster.
+	// +optional
+	Decryption Decryption `json:"decryption,omitempty"`
```
hiddeco, line 21:
```suggestion
	Decryption *Decryption `json:"decryption,omitempty"`
```

controllers/kustomization_decryptor.go
```diff
@@ -0,0 +1,120 @@
...
+	return nil
+}
+
+func (kd *KustomizeDecryptor) gpgImport(path string) error {
+	cmd := exec.Command("gpg", "--import", path)
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("gpg import error: %s", string(out))
+	}
+	return nil
```
hiddeco, line 119: This leaks any imported key to other `Kustomization` resources, which is probably not something we want to do.

docs/spec/v1alpha1/kustomization.md
```diff
@@ -60,6 +64,21 @@ type KustomizationSpec struct {
...
+```go
+type Decryption struct {
+	// Provider is the name of the decryption engine.
+	// +kubebuilder:validation:Enum=sops
+	// +required
+	Provider string `json:"provider"`
+
+	// The secret name containing the private OpenPGP keys used for decryption.
+	// +optional
+	SecretRef corev1.LocalObjectReference `json:"secretRef,omitempty"`
```
hiddeco, line 78:
```suggestion
	SecretRef *corev1.LocalObjectReference `json:"secretRef,omitempty"`
```

controllers/kustomization_decryptor.go
stefanprodan: Yeah that's the tradeoff for running the decryption in the same container for all tenants. How can we isolate the keys per kustomization with nupg?

hiddeco: By using a clean context for the command to run in, and setting the `GNUPGHOME` for the duration of the decryption process to a temporary directory.

Other option would be to provide the `--homedir` argument to the `gpg` command, and set the `SOPS_GPG_EXEC` variable when running the `sops` command (https://github.com/mozilla/sops/blob/master/pgp/keysource.go#L54), but I think this is less foolproof.

hiddeco (approved):
This looks good to go to me for now. In a future iteration we should have a look at getting rid of the `gpg` binary as we now have full control over the process, and can now make it so that it does not have to fallback.

:100:

PR #96, opened 2020-09-02
Publish amd64/arm64 images to GHCR

stefanprodan:
Ref: https://github.com/fluxcd/toolkit/issues/194


hiddeco (approved):
:1st_place_medal: :100:

PR #103, opened 2020-09-09
Use annotation and predicate from fluxcd/pkg

squaremo:
The reconcileAt annotation and the predicate that recognises it are now in fluxcd/pkg/runtime. This removes (near) duplicate definitions in favour of using those.


go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/fluxcd/pkg/lockedfile v0.0.5
 	github.com/fluxcd/pkg/logger v0.0.1
 	github.com/fluxcd/pkg/recorder v0.0.6
+	github.com/fluxcd/pkg/runtime v0.0.0-20200909163337-e7e634246495
```
stefanprodan, line 12:
```suggestion
	github.com/fluxcd/pkg/runtime v0.0.1
```

stefanprodan (changes requested):

stefanprodan (approved):
LGTM

Issue #98, opened 2020-09-04
Check pruning objects with finalizers behaves reasonably


stealthybox:
During pruning, an object with remaining finalizers can cause deletes to hang, especially when the controllers for those objects are uninstalled first or if they are down/disabled.

We should check what happens in our apply controllers:
Some good behavior could look like:
- object is not deleted
- does not affect creates/updates of other objects
- does not affect deletes of other objects
- create/update/delete of other objects is timely or done concurrently
- an object specific error about finalizers is returned for `PruneFailedReason`
- multiple errors are aggregated

Alternatively, if pruning is done first, maybe it's best not to continue on with anything if it fails.

Inspired by this user-reported issue in Fluxv1,
figured we should make sure we do it right:
> I figured it out, after way more time than i hoped for.  git was all ok since fluxctl release --dry-run knew exactly what needed to be done (it was able to do the diff with git), it just couldn't execute the change.  the reason is that one of the kube resources could not be deleted that was related to some recently removed resource.  because of that, flux seems to abort everything and just time out, not just in that namespace, for for ANYTHING including changing the tags on images.  it just becomes paralyzed for all subsequent gitops work.  I wish it
let me know what resource or action it timed out on
it doesn't shut down all other unrelated syncs
For the record, resources can become undeletable if they have finalizers on them (like the new istio-operator does).  then you have to patch the resource to remove the finalizer, and then remove it.
> https://cloud-native.slack.com/archives/CLAJ40HV3/p1599242087142400?thread_ts=1599175627.127600&cid=CLAJ40HV3

stefanprodan:
> Alternatively, if pruning is done first, maybe it's best not to continue on with anything if it fails.

We can't run prune before apply because it will result in downtime if you rename a deployment.

The kustomize-controller GC is very different to Flux v1. Unlike Flux, we store a snapshot of the previous applied Kinds in the status sub-resource, so we know what needs to be pruned after we apply a new revision. 

I think we should detect stuck finalizers and skip pruning those objects on the 2nd GC run. This way we don't block dependants, and would make GC reliable as it will remove newly resources masked for deletion. 

@stealthybox do you think you can help out improving the GC when finalizers are stuck?

PR #104, opened 2020-09-10
Refactor garbage collector

stefanprodan:
Changes:
- rewrite GC with controller-runtime client instead of kubectl shell-outs
- track group version kinds instead of kinds (GC snapshot API breaking change)
- ignore objects that are no longer present on the cluster
- delete namespaced CRs before CRDs
- mark for deletion objects with finalizers
- ignore objects stuck in finalization

Fix: #98


hiddeco (approved):
LGTM :1st_place_medal:

PR #115, opened 2020-09-17
ci: Add sanity check for container images

stefanprodan:
Fail build if container images have been pushed with corrupted layers and enable buildx debug logs, ref: https://github.com/fluxcd/toolkit/issues/241

PS. This needs to be ported on all the toolkit controllers


crazy-max:
Nice LGTM :)

hiddeco (approved):

PR #118, opened 2020-09-19
Send event when reconcile succeeds with update metadata

phillebaba:
Related to:
https://github.com/fluxcd/notification-controller/issues/45


stefanprodan (approved):
LGTM

Thanks @phillebaba

PR #123, opened 2020-09-29
Implement namespace overriding

stefanprodan:
Changes:
- Add `TargetNamespace` field to API and docs
- Implement namespace overriding

This PR allows setting or overriding the namespace in the` kustomization.yaml`, making possible for cluster admins to onboard external apps into existing namespaces. This will also open the path to create ephemeral namespaces for PR reviews.


hiddeco (approved):
LGTM, :+1: for including validation rules

Think you need to rerun `make test` however to make CI happy.

PR #132, opened 2020-10-04
Update health check docs

phillebaba:
Add a notice regarding using health checks with Helm charts containing a single replica.
This notice should be removed when https://github.com/helm/helm/pull/8671 is merged.

Fixes https://github.com/fluxcd/helm-controller/issues/81


stefanprodan (approved):
LGTM

Thanks @phillebaba

PR #138, opened 2020-10-09
Implement Prometheus instrumentation

stefanprodan:
This PRs adds Prometheus instrumentation for the Kustomization reconciliation duration and ready status in the same fashion as [kube-state-metrics](https://github.com/kubernetes/kube-state-metrics) does for Deployment conditions. 

Ready status example:
```
# HELP gotk_reconcile_condition The current condition status of a GitOps Toolkit resource reconciliation.
# TYPE gotk_reconcile_condition gauge
gotk_reconcile_condition{kind="Kustomization",name="podinfo",namespace="gotk-system",status="Deleted",type="Ready"} 0
gotk_reconcile_condition{kind="Kustomization",name="podinfo",namespace="gotk-system",status="False",type="Ready"} 0
gotk_reconcile_condition{kind="Kustomization",name="podinfo",namespace="gotk-system",status="True",type="Ready"} 1
gotk_reconcile_condition{kind="Kustomization",name="podinfo",namespace="gotk-system",status="Unknown",type="Ready"} 0
```

Ready alert example:
```yaml
groups:
- name: GitOpsToolkit
  rules:
  - alert: ReconciliationFailure
    expr: gotk_reconcile_condition{type="Ready",status="False"} == 1
    for: 10m
    labels:
      severity: page
    annotations:
      summary: '{{ $labels.kind }} {{ $labels.namespace }}/{{ $labels.name }} reconciliation has been failing for more than ten minutes.'
```

Reconciliation duration example:

```
# HELP gotk_reconcile_duration The duration in seconds of a GitOps Toolkit resource reconciliation.
# TYPE gotk_reconcile_duration histogram
gotk_reconcile_duration_bucket{kind="Kustomization",name="podinfo",namespace="gotk-system",le="1e-08"} 0
gotk_reconcile_duration_bucket{kind="Kustomization",name="podinfo",namespace="gotk-system",le="1e-07"} 0
gotk_reconcile_duration_bucket{kind="Kustomization",name="podinfo",namespace="gotk-system",le="1e-06"} 0
gotk_reconcile_duration_bucket{kind="Kustomization",name="podinfo",namespace="gotk-system",le="9.999999999999999e-06"} 0
gotk_reconcile_duration_bucket{kind="Kustomization",name="podinfo",namespace="gotk-system",le="9.999999999999999e-05"} 0
gotk_reconcile_duration_bucket{kind="Kustomization",name="podinfo",namespace="gotk-system",le="0.001"} 0
gotk_reconcile_duration_bucket{kind="Kustomization",name="podinfo",namespace="gotk-system",le="0.01"} 0
gotk_reconcile_duration_bucket{kind="Kustomization",name="podinfo",namespace="gotk-system",le="0.1"} 0
gotk_reconcile_duration_bucket{kind="Kustomization",name="podinfo",namespace="gotk-system",le="1"} 1
gotk_reconcile_duration_bucket{kind="Kustomization",name="podinfo",namespace="gotk-system",le="10"} 2
gotk_reconcile_duration_bucket{kind="Kustomization",name="podinfo",namespace="gotk-system",le="+Inf"} 3
gotk_reconcile_duration_sum{kind="Kustomization",name="podinfo",namespace="gotk-system"} 36.929779548000006
gotk_reconcile_duration_count{kind="Kustomization",name="podinfo",namespace="gotk-system"} 3
```

Ref: https://github.com/fluxcd/toolkit/issues/329


internal/metrics/recorder.go
```diff
@@ -0,0 +1,62 @@
...
+	readyGauge *prometheus.GaugeVec
+}
+
+func NewRecorder() *Recorder {
+	return &Recorder{
+		readyGauge: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Namespace: "",
+				Subsystem: "",
+				Name:      "gitops_toolkit_ready_condition",
```
seaneagan, line 24: Are we sure we don't want to parameterize this metric on the condition type? Even if it is decided not to expose the other condition types yet, this would at least allow for that possibility in the future. Which would then be useful for building generic dashboards for all condition types without having to hard code each separate metric name for each condition type.

controllers/kustomization_controller.go
```diff
@@ -142,6 +148,7 @@ func (r *KustomizationReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 			log.Error(err, "unable to update status")
```
seaneagan, line 148: Github won't let me comment on L136 (after the ready condition is set to false due to progressing reason) but shouldn't we record readiness there since its value has potentially changed?

controllers/kustomization_controller.go
```diff
@@ -768,6 +779,24 @@ func (r *KustomizationReconciler) event(kustomization kustomizev1.Kustomization,
...
+
+	objRef, err := reference.GetReference(r.Scheme, &kustomization)
+	if err != nil {
+		r.Log.WithValues(
+			strings.ToLower(kustomization.Kind),
+			fmt.Sprintf("%s/%s", kustomization.GetNamespace(), kustomization.GetName()),
+		).Error(err, "unable to record readiness metric")
+		return
+	}
+	if rc := meta.GetCondition(kustomization.Status.Conditions, meta.ReadyCondition); rc != nil {
```
seaneagan, line 795: From the kubernetes [API conventions](https://github.com/evankanderson/k8s-community/blob/master/contributors/devel/sig-architecture/api-conventions.md):

> The absence of a condition should be interpreted the same as Unknown.

So even if the ready condition is not found yet shouldn't we still record the readiness as unknown?

controllers/kustomization_controller.go
stefanprodan: Good catch! Fixed, thank you.

controllers/kustomization_controller.go
stefanprodan: I don't see how this is possible with the current code but I've done as you asked.

internal/metrics/recorder.go
stefanprodan: We could make it `gitops_toolkit_condition` with the a label `type` or you had something different in mind?

seaneagan: That works for me. If we want something more specific could be e.g. `gitops_toolkit_condition_status_equals`.

internal/metrics/recorder.go
```diff
@@ -0,0 +1,60 @@
...
+				}
+			case corev1.ConditionFalse:
+				if status == string(condition.Status) {
+					value = 1
+				}
+			case corev1.ConditionUnknown:
+				if status == string(condition.Status) {
+					value = 1
+				}
+			}
```
hiddeco, line 55: Doesn't this just equal to:

```suggestion
			if string(condition.Status) == status {
				value = 1
			}
```

stefanprodan: You are right, I've moved things around and the switch is useless now.

hiddeco (approved):
Awesome :medal_sports:

PR #141, opened 2020-10-14
Fix status reporting when the source is not found

stefanprodan:
When the source is not found the status wasn't updated accordingly.

### DIFF
--- a/controllers/kustomization_controller.go
+++ b/controllers/kustomization_controller.go
@@ -31,6 +31,7 @@ import (
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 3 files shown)


controllers/kustomization_controller.go
```diff
@@ -147,12 +147,7 @@ func (r *KustomizationReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 
 	// resolve source reference
 	source, err := r.getSource(ctx, kustomization)
-	if err != nil {
-		return ctrl.Result{}, err
-	}
-
-	// check source readiness
-	if source.GetArtifact() == nil {
+	if err != nil || source.GetArtifact() == nil {
```
hiddeco, line 150: Looking at the logic of `getSource`, the result of this is that a "not found" or transient error (they are both the same there, instead of relying on `apierr.IsNotFound()`) is equal to the source not being ready. I am not sure if this is what you are after.

stefanprodan: Good point 👍 Please check the latest push.

hiddeco (approved):

PR #148, opened 2020-10-20
Improve remote cluster documentation

hiddeco:
To highlight the fact that it works with just a KubeConfig, and not just
CAPI.


stefanprodan (approved):
Thanks @hiddeco

PR #149, opened 2020-10-20
Add images attribute to set application image name/tag dynamically

takutakahashi:
This implementation is a feature porting from `kustomization.yaml` .
https://kubectl.docs.kubernetes.io/pages/app_management/container_images.html

It becomes to change image name/tag dynamically with patching `Kustomization` resources.


api/v1beta1/kustomization_types.go
```diff
@@ -92,6 +93,11 @@ type KustomizationSpec struct {
 	// +optional
 	Timeout *metav1.Duration `json:"timeout,omitempty"`
 
+	// Images override image definitions in manifests
+	// It can set the same configuration parameters with kustomization.yaml
+	// +optional
+	Images []kustypes.Image `json:"images,omitempty"`
```
stefanprodan, line 99: We need to define our own Image type and not make the API package depend on kustomize.

controllers/kustomization_generator.go
```diff
@@ -80,6 +80,8 @@ func (kg *KustomizeGenerator) WriteFile(dirPath string) (string, error) {
 		kus.Namespace = kg.kustomization.Spec.TargetNamespace
 	}
 
+	kus.Images = kg.kustomization.Spec.Images
```
stefanprodan, line 83: This removes images even if the override is for some of them. We need to merge the two arrays and override only the ones specified in `kustomization .Spec.Images`.

stefanprodan (commented):
@takutakahashi can you please create an API change proposal and explain why is this needed in https://github.com/fluxcd/toolkit/discussions

takutakahashi:
@stefanprodan 
Thank you for your advice. I will send a proposal to this discussion board!

stefanprodan:
Implemented in #194

PR #150, opened 2020-10-21
Use controller-runtime utils for finalizer and health checks

hiddeco:


main.go
```diff
@@ -64,6 +66,7 @@ func main() {
 
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&eventsAddr, "events-addr", "", "The address of the events receiver.")
+	flag.StringVar(&healthAddr, "health-addr", ":9440", "The address the health endpoint binds to.")
```
stefanprodan, line 69: We should change the the probes in the deployment manifest.

stefanprodan (commented):
Please run make

stefanprodan (approved):
LGTM

PR #151, opened 2020-10-21
Suppress health check events when no changes made

ordovicia:
When no Kubernetes resources are created or configured for a `Kustomization` and it has no dependencies, the health check status for the `Kustomization` will not change.
In this case kustomize-controller does not have to deliver health check events.
What do you think?

### DIFF
--- a/controllers/kustomization_controller.go
+++ b/controllers/kustomization_controller.go
@@ -753,7 +753,7 @@ func (r *KustomizationReconciler) checkHealth(statusPoller *polling.StatusPoller
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


stefanprodan:
@ordovicia I think it's ok to suppress the events no matter the dependencies. If a dependency is failing, it will never get to the health check part.

ordovicia:
Ah, right. Thank you.
I removed the `len(DependsOn) > 0` condition.

stefanprodan (approved):
LGTM

Thanks @ordovicia 🏅

PR #152, opened 2020-10-22
Use annotation helpers

squaremo:
Per https://github.com/fluxcd/pkg/issues/39.


stefanprodan (approved):
LGTM

Thanks @squaremo

PR #153, opened 2020-10-27
Move dedicated watchers to in-controller watches

hiddeco:
This prevents the resources from getting annotated, and instead uses
the `handler.EnqueueRequestsFromMapFunc` to queue requests based on
changes to the source objects.


stefanprodan (approved):
LGTM

Thanks @hiddeco 🍰

PR #157, opened 2020-10-29
Release v0.2.0

stefanprodan:
Ref: https://github.com/fluxcd/toolkit/issues/388


hiddeco (commented):
`config/default/kustomization.yaml` requires an update too.

hiddeco (approved):
:flying_saucer: :dash:

PR #161, opened 2020-10-30
Update k8s to 1.19 + kustomize 0.6.4

relu:
Kubernetes 1.19 and Kustomize patch version bump.

Note k8s 1.19 brings in klog v2, hence the update on the logr and respectively zapr (indirect, just for compatibility reasons) packages.


stefanprodan:
Hi @relu 

The Kubernetes package upgrade should come from [fluxcd/pkg](https://fluxcd/pkg), we need to bump the version in `apis/meta` and `runtime`, do a release of both packages then update all controller.

cc @hiddeco

relu:
Thanks, @stefanprodan! I missed the fact that these dependencies are pulled in through pkg. I will follow-up with a PR there and update the versions. Regardless, this set of PRs on the components is probably still going to be relevant with the addition that the pkg deps will also need to be updated once their new versions will be released.

stefanprodan (approved):
LGTM

Thanks @relu

Issue #120, opened 2020-09-21
Allow kustomize plugins to be optionally enabled


polivbr:
It would be very useful to allow the new-ish kustomize plugin functionality to be optionally enabled. I understand that there are potential security issues around plugins and that it's an alpha-level feature, so I wouldn't enable by default, but in a tightly controlled environment, they can be extremely useful.

stefanprodan:
I don't think this is possible due to the fact that the controller uses kustomize as library and the Go plugins will faild the compilation conditions, more details here https://kubernetes-sigs.github.io/kustomize/guides/plugins/goplugincaveats/

As for exec plugins, those should work if you install them inside the container at build time, this is not possible at runtime, since the controller rootfs is readonly due to security reasons.

polivbr:
If you allow the plugin home directory to be configurable (which I see can be done through the api), it could be pointed at a directory inside of the source, which would allow plugins to be stored in source control alongside the kustomizations. This would get around the issue of the readonly rootfs.

stefanprodan:
> If you allow the plugin home directory to be configurable (which I see can be done through the api), it could be pointed at a directory inside of the source

Ah yes this would the best approach  💯

seaneagan:
Note that kustomize plugins [can now be defined](https://github.com/kubernetes-sigs/kustomize/pull/2597) as [KRM config functions](https://github.com/kubernetes-sigs/kustomize/blob/master/cmd/config/docs/tutorials/function-basics.md) which are container based, and thus don't require the plugin home directory mapping, and should just come for free (assuming plugins are enabled, new enough version of kustomize, etc). It's [likely](https://github.com/GoogleContainerTools/kpt/issues/712) that the other plugin runtimes will be deprecated in favor of KRM config functions, in which case it may be better to avoid adding support for them.

hiddeco:
@seaneagan `kpt` is a client side tool, that can perform different magic tricks than controllers living within Kubernetes. See my comment here: https://github.com/fluxcd/helm-controller/issues/104#issuecomment-704392154

seaneagan:
@hiddeco good point, configuring docker-in-docker (in kubernetes) should be possible, but much trickier than simply running plugins in an earlier build step before Kubernetes is involved. Note that KRM config functions are not just a `kpt` feature, they are also targeted to `kustomize` (currently behind a flag).

hiddeco:
[...]

PR #163, opened 2020-11-03
Optionally enable external plugins

polivbr:
External plugins are a very useful and power feature of kustomize, but
due to their current alpha status, are not not enabled by default.

This commit allows an optional pluginHome path to be specified,
relative to the source root, which will enable the plugin functionality.


polivbr:
Addresses issue #120

stefanprodan:
Hmm so people are expected to place Linux binaries in their repo then execute them inside the controller container. I wander if this is a miss use of Git, let's discuss this at the next Flux meeting.

polivbr:
There's nothing saying that the executables have to be linux binaries. They could be shell scripts, which to me is no different than having CI build scripts in your git repo. Even if someone wanted to use binaries, that should be the decision of the repo owner and shouldn't be dictated by the tooling.

zhech2:
My team is interested in this change. Is there anything we can do to help?

mhaley-tignis:
Any way we can push this forward? Our team is using the `--enable-helm` option with `kustomize` when running locally. I would like to use the same manifests in our cloud clusters managed by fluxv2 but currently I get a reconciliation error:

```
{"level":"error","ts":"2021-09-14T15:15:13.101Z","logger":"controller.kustomization","msg":"Reconciliation failed after 42.506436ms, next try in 10m0s","reconciler group":"kustomize.toolkit.fluxcd.io","reconciler kind":"Kustomization","name":"aipc","namespace":"flux-system","revision":"main/d4e9662ac29fb45bb9a61bba662e8ba815b60a4f","error":"kustomize build failed: accumulating resources: 2 errors occurred:\n\t* accumulateFile error: \"accumulating resources from '../../components/cert-manager-installation': read /tmp/aipc237354456/infrastructure/components/cert-manager-installation: is a directory\"\n\t* accumulateDirector error: \"couldn't make target for path '/tmp/aipc237354456/infrastructure/components/cert-manager-installation': json: unknown field \\\"helmCharts\\\"\"\n\n"}
```

Being able to set the [plugin config](https://github.com/fluxcd/kustomize-controller/blob/74f08c3f1b3ec8ab8725c7775da8d32903996835/controllers/kustomization_generator.go#L402) based into build would solve our problem. There is an enabled plugin config already (see [here](https://github.com/kubernetes-sigs/kustomize/blob/api/v0.9.0/api/types/pluginconfig.go#L26-L33)). Can we an option to use that configuration instead of the default?

I don't see this changing security threat model at all since the helm binary being executed has to be baked into the container. As long as it is trusted, and works as intended, things should be fine.

khrisrichardson:
@mhaley-tignis This PR looks to be more specific to running exec plugins. A patch I reference in https://github.com/fluxcd/kustomize-controller/issues/323#issuecomment-913767721 enables both the KRM config function plugins and the Helm generator. I agree that those could be handled by kustomize-controller command line flags, with the caveat that if someone wishes to use the KRM config functions, they'll need to provide a `gotk-patches.yaml` file to mount `/var/run/docker.sock`, which would be a security concern, albeit not germane to your request to enable the Helm generator. The other caveat that is germane to your request is that the kustomize-controller container image doesn't ship with the Helm binary ... another reason I went the custom image route.

mhaley-tignis:
You are right. I think we are just going to switch to running `helm template` ourselves on the few charts we rely on, checking the resulting manifest in, then just using kustomize like we do for everything else, and patching the manifest if/when needed. When the external chart is updated, our upgrade path should be as simple as `helm template` again (minus all the other complications of live updates :) ).

HerrmannHinz:
> Hmm so people are expected to place Linux binaries in their repo then execute them inside the controller container. I wander if this is a miss use of Git, let's discuss this at the next Flux meeting.

hey @stefanprodan  - do you remember what was the outcome of the meeting? or, can we look at some docs/discussion/meeting-recording regards this? 
TBH im totally new to the concept of plugins aka binaries as kustomize plugins.

akokshar:
Are there any plans to merge this? 
Not being able to use self-written exec plugins is a blocking obstacle for us to adopt fluxcd :/

stefanprodan:
I will close this PR when we get OCI artifacts support into Flux. Using OCIRepository instead of GitRepository will enable a workflow where you run `kustomize build` and `flux push` in CI, then you reconcile the final manifests with a Flux Kustomization that uses the OCI artifact as source.

The workflow goes like this:
- you commit a change to Git
- CI runs `kustomize build --enable-alpha-plugins --enable-exec --enable-helm . > ./app-deploy/final.yaml`
- CI runs `flux push artifact ghcr.io/org/app-config:latest --path ./app-deploy`

On the cluster you would reconcile the app with:

```yaml
apiVersion: source.toolkit.fluxcd.io/v1beta2
kind: OCIRepository
metadata:
  name: my-app
  namespace: apps
spec:
  interval: 10m
  url: ghcr.io/org/app-config
  ref:
    tag: "latest"
---
kind: Kustomization
metadata:
  name: my-app
  namespace: apps
spec:
  targetNamespace: apps
  interval: 10m
  path: "./app-deploy"
  prune: true
  sourceRef:
    kind: OCIRepository
    name: my-app
  wait: true
  timeout: 2m
```

Ref: https://github.com/fluxcd/kustomize-controller/pull/684

PR #164, opened 2020-11-04
Use latest generation when updating final status

stefanprodan:
To avoid stale status writes, we use the latest object generation before updating the final status sub-resource.


hiddeco (approved):
With a note that I am not yet sure this is the most idiomatic way of dealing with this issue, as I am currently in the process of finding this out myself for the Helm Controller.

PR #174, opened 2020-11-17
Refactor to adopt k8s standardized Condition type

relu:
Updates to use metav1.Condition type and removes references for
deprecated corev1.Condition* constants and uses the new k8s api/meta
helpers in place of the old pkg/apis/meta types.


stefanprodan (commented):
Can you please bump source-controller to v0.3.0 in:
-  go.mod
- Makefile
- config/default/kustomization.yaml

stefanprodan (approved):
LGTM 💯

PR #176, opened 2020-11-18
Add docs for excluding non-k8 YAML files in kustomization generation

### LINKED ISSUE #175: Kustomization creation fails when dir contains non-k8 yaml files
First of all, thanks for the amazing work on flux v2! 

When testing migration from v1, I've noticed an issue with kustomization generation: Given a directory, which does not contain kustomization.yaml, but containing other yaml files not related to kubernetes objects kustomization creation fails with:

`kustomize create failed: failed to decode Kubernetes YAML from /tmp/flux-system630708581/.sops.yaml: error unmarshaling JSON: while decoding JSON: Object 'Kind' is missing in '{"creation_rules":[{"XXX":"projects/XXX/locations/global/keyRings/XXX/cryptoKeys/XXX","path_regex":"\\.secret\\.yaml$"}]}'`

This can be triggered by having YAML files like `.sops.yaml` or ci workflows definitions e.g. `.gitlab-ci.yml` in root directory and not having `kustomization.yaml`

Flux v1 ignored non-k8 yaml files thanks to [this func ](https://github.com/fluxcd/flux/blob/ba54af3787709cd7468d6d9f2496e3f615431ece/pkg/cluster/kubernetes/resource/load.go#L151), whereas v2 and the kustomization flow requires all of yaml files to be k8 by using [SliceForBytes](https://github.com/kubernetes/cli-runtime/blob/master/pkg/kustomize/k8sdeps/kunstruct/factory.go#L46) for validation.

Possible workaround: simply adding kustomization.yaml in the affected directories, but this requires action and makes migration a bit more demanding for new users. 

I'd be happy to contribute a fix or to document the workaround in case this gets considered as working as intended.

kowalczykp:
resolves #175

### DIFF
--- a/docs/spec/v1beta1/kustomization.md
+++ b/docs/spec/v1beta1/kustomization.md
@@ -214,7 +214,25 @@ Source supported types:
 
 If your repository contains plain Kubernetes manifests, the `kustomization.yaml`
 file is automatically generated for all the Kubernetes manifests
-in the `spec.path` and sub-directories.
+in the `spec.path` and sub-directories. This expects all YAML files present under that path to be valid kubernetes manifests
+and needs non-kubernetes ones to be excluded using `.sourceignore` file or `spec.ignore` on `GitRepository` object.
+
+Example of excluding gitlab ci workflows and sops rules creation files:
+
+```yaml
+apiVersion: source.toolkit.fluxcd.io/v1beta1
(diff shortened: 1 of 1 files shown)


stefanprodan (approved):
LGTM

Thanks @kowalczykp

PR #177, opened 2020-11-19
Use DeletionTimestamp for prune and readiness

hiddeco:


stefanprodan (approved):
LGTM

Thanks @hiddeco 🍰

PR #181, opened 2020-11-20
Release v0.3.0

stefanprodan:
This is the third MINOR prerelease. It introduces two breaking changes:
* the `ServiceAccount` field has been removed and replaced by `ServiceAccountName`; it is no longer possible to specify a namespace for a service account, the namespace is inferred from the Kustomization namespace 
* the status condition type has changed to the type introduced in Kubernetes API machinery `v1.19.0`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,28 @@
 
 All notable changes to this project are documented in this file.
 
+## 0.3.0 (2020-11-20)
+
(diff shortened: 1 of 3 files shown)


hiddeco:
Maybe include a note about the `namespace` no longer being supported as part of the `ServiceAccount` field change?

hiddeco (approved):

PR #185, opened 2020-11-26
Remove default value from suspend field

stefanprodan:


api/v1beta1/kustomization_types.go
```diff
@@ -77,11 +77,9 @@ type KustomizationSpec struct {
 	SourceRef CrossNamespaceSourceReference `json:"sourceRef"`
 
 	// This flag tells the controller to suspend subsequent kustomize executions,
-	// it does not apply to already started executions. Defaults to false.
-	// +kubebuilder:default:=false
-	// +kubebuilder:validation:Optional
+	// it does not apply to already started executions.
```
hiddeco, line 80:
```suggestion
	// it does not apply to already started executions. Defaults to false.
```

The behavior still defaults to false :-)

hiddeco (approved):

Issue #203, opened 2020-12-14
Throttle reconciliation in case of error


raffis:
If there is an error it looks like the configured interval does not get considered.

I have a `10m` configured interval and one error in a manifest and see a reconciliation every ~10s.
Which also leads to a slack alert for each. (Also the slack alerts are useless because of #190, I don't see the actual error in the notification nor the log because of #202).

```
{"level":"error","ts":"2020-12-14T10:23:53.606Z","logger":"controllers.Kustomization","msg":"unable to update status after reconciliation","controller":"kustomization","request":"flux-system/devops-k8s","error":"Kustomization.kustomize.toolkit.fluxcd.io \"devops-k8s\" is invalid: status.conditions.message: Invalid value: \"\": status.conditions.message in body should be at most 32768 chars long"}
{"level":"error","ts":"2020-12-14T10:23:53.606Z","logger":"controller","msg":"Reconciler error","reconcilerGroup":"kustomize.toolkit.fluxcd.io","reconcilerKind":"Kustomization","controller":"kustomization","name":"devops-k8s","namespace":"flux-system","error":"Kustomization.kustomize.toolkit.fluxcd.io \"devops-k8s\" is invalid: status.conditions.message: Invalid value: \"\": status.conditions.message in body should be at most 32768 chars long"}
{"level":"error","ts":"2020-12-14T10:24:21.658Z","logger":"controllers.Kustomization","msg":"unable to update status after reconciliation","controller":"kustomization","request":"flux-system/devops-k8s","error":"Kustomization.kustomize.toolkit.fluxcd.io \"devops-k8s\" is invalid: status.conditions.message: Invalid value: \"\": status.conditions.message in body should be at most 32768 chars long"}
{"level":"error","ts":"2020-12-14T10:24:21.658Z","logger":"controller","msg":"Reconciler error","reconcilerGroup":"kustomize.toolkit.fluxcd.io","reconcilerKind":"Kustomization","controller":"kustomization","name":"devops-k8s","namespace":"flux-system","error":"Kustomization.kustomize.toolkit.fluxcd.io \"devops-k8s\" is invalid: status.conditions.message: Invalid value: \"\": status.conditions.message in body should be at most 32768 chars long"}
[...]

Issue #202, opened 2020-12-14
Error hidden with info log level and status body too long


raffis:
This is partially related to #190.

If many manifests are in a repository and one of these manifest got an error the error is not visible in info log level because you end up seeing:
```
{"level":"error","ts":"2020-12-14T09:52:40.719Z","logger":"controllers.Kustomization","msg":"unable to update status after reconciliation","controller":"kustomization","request":"flux-system/k8s","error":"Kustomization.kustomize.toolkit.fluxcd.io \"k8s\" is invalid: status.conditions.message: Invalid value: \"\": status.conditions.message in body should be at most 32768 chars long"}
```

However if the log level is `debug` the real error is visible before that in the debug msg `Normal`.
The error should be visible in all log levels despite seeing `body too long error`.

Issue #190, opened 2020-12-01
Ability to control verbosity of events


invke:
The output of the events of the kustomize-controller quite quickly get too long on medium-sized clusters. This then causes the output in the slack provider from the notification-controller to exceed the [slack message limit](https://api.slack.com/changelog/2018-04-truncating-really-long-messages), truncating the event. This is a pretty easy limit to hit, I'm hitting it with ~200 resources. With any errors located at the end, the most useful information is lost from the message. This relates to an [issue made on the notification-controller](https://github.com/fluxcd/notification-controller/issues/81) but probably better belongs here, as it's caused by the payload of the event sent by the kustomize-controller.

I originally thought it might be a good idea to configure the verbosity of the underlying `kubectl apply` that feeds into the info/error events, but that still doesn't prevent the output of the unchanged resources that floods the output.

Would it be appropriate to filter out lines that have `unchanged`? There are probably better solutions, like some kind of "smarter" message as mentioned by @phillebaba [here](https://github.com/fluxcd/notification-controller/issues/81#issuecomment-736564387), but this to me seems the simplest, albeit depending on the output of `kubectl apply`.

stefanprodan:
> Would it be appropriate to filter out lines that have unchanged?

I think this could work but we should filter the output only when the apply fails.

Issue #167, opened 2020-11-09
Add kubectl error code to error messages


sylr:
It happens that the controller logs an error with an empty message,e.g.:

```json
{"level":"error","ts":"2020-11-09T13:55:26.336Z","logger":"controller","msg":"Reconciler error","reconcilerGroup":"kustomize.toolkit.fluxcd.io","reconcilerKind":"Kustomization","controller":"kustomization","name":"aks20-euw-infra-main-m
onitoring","namespace":"flux","error":"apply failed: "}
```

In that case the error message should mention that kubectl did not write anything on stderr and print the error code.

PR #205, opened 2020-12-14
Refactor apply error reporting

stefanprodan:
Changes:
- filter kubectl apply output and extract errors
- limit apply output to 20K charts (avoid reaching max etcd size)
- log kubectl exit code when the process is killed

Fix: #167
Fix: #190
Fix: #202
Fix: #203

### DIFF
--- a/controllers/utils.go
+++ b/controllers/utils.go
@@ -1,0 +1,71 @@
+/*
+Copyright 2020 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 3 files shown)


controllers/utils.go
```diff
@@ -0,0 +1,68 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import "strings"
+
+// parseApplyOutput extracts the objects and the action
+// performed by kubectl (created, configured, unchanged)
```
hiddeco, line 22: Can you add an example of the format of (a line from) the apply output, that would help me to better understand how it is parsed.

hiddeco (approved):

Issue #209, opened 2020-12-15
GC seems to abandon objects


nairb774:
Looking at https://github.com/fluxcd/kustomize-controller/blob/8296b8e1f1cf497b694b2fd98a0a6658e78c3a77/controllers/kustomization_gc.go#L68 it seems that the GC logic only searches for objects with a known prior checksum, but this seems to leave some objects behind. I unfortunately don't know how it happened, but I definitely have objects that match the `kustomize.toolkit.fluxcd.io/name` and `kustomize.toolkit.fluxcd.io/namespace` labels but do not have the correct `kustomize.toolkit.fluxcd.io/checksum`. I noticed this when we attempted to clean up some previously deployed resources, but they weren't removed.

```
$ kubectl get statefulset -l kustomize.toolkit.fluxcd.io/namespace=flux-system,kustomize.toolkit.fluxcd.io/name=deploy-ns-kafka -L kustomize.toolkit.fluxcd.io/checksum -A
NAMESPACE   NAME                 READY   AGE   CHECKSUM
kafka       connect              3/3     63d   626af1c4a8d9c2ac041f8e6db6a56d53d936c94a
kafka       connect-mongo-test   0/0     40d   e70b56648e95ab4befdf1c82f49ec2c4ff7c53fc
```

(The connect-mongo-test was scaled to 0 long ago, and was only recently cleaned up)

The relevant Kustomization:

```yaml
kind: Kustomization
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"kustomize.toolkit.fluxcd.io/v1beta1","kind":"Kustomization","metadata":{"annotations":{},"labels":{"managed-namespace":"kafka","shell-operator":"006-flux-provision"},"name":"deploy-ns-kafka","namespace":"flux-system","ownerReferences":[{"apiVersion":"v1","kind":"Namespace","name":"kafka","uid":"fc483992-68fe-4aa5-8c8d-440607a0fcc6"}]},"spec":{"dependsOn":[{"name":"cluster"}],"interval":"1m","path":"$REMOVED","prune":true,"serviceAccountName":"deploy-ns-kafka","sourceRef":{"kind":"GitRepository","name":"k8s"},"suspend":false}}
  creationTimestamp: "2020-12-14T18:57:02Z"
  finalizers:
  - finalizers.fluxcd.io
  generation: 4
  labels:
    managed-namespace: kafka
    shell-operator: 006-flux-provision
[...]

PR #210, opened 2020-12-16
Refactor garbage collection

stefanprodan:
Delete stale objects if their checksum doesn't match the last applied revision. This accounts for controller restarts during reconciliation, previously, if GC was killed while running, stale objects would never be removed.

Pin controller-runtime to v0.6.3 (mitigation until we upgrade to v0.7.0), ref: #214, fix: #209


controllers/kustomization_gc.go
```diff
@@ -120,8 +122,13 @@ func (kgc *KustomizeGarbageCollector) Prune(timeout time.Duration, name string,
 	return changeSet, true
 }
 
-func (kgc *KustomizeGarbageCollector) matchingLabels(name, namespace, checksum string) client.MatchingLabels {
-	return gcLabels(name, namespace, checksum)
+func (kgc *KustomizeGarbageCollector) isStale(obj unstructured.Unstructured) bool {
+	itemChecksum := obj.GetLabels()[fmt.Sprintf("%s/checksum", kustomizev1.GroupVersion.Group)]
```
hiddeco, line 126: I would move the `fmt.Sprintf` to a dedicated function as it is also used in the `gcLabels` function, or move the various label names to constants, and create something like:

```go
const gcNameLabel = "name"

func gcLabelName(name string) string {
	return fmt.Sprintf("%s/%s", kustomizev1.GroupVersion.Group, name)
}
```

hiddeco (approved):

PR #219, opened 2020-12-17
Emit healthcheck event when Kustomization was not ready

ordovicia:
Healthcheck can temporarily fail due to slow application startup, network failure, etc.
In this case, Flux emits an event of the failed healthcheck.
But when the healthcheck passes in another reconciliation, the current implementation does not emit a healthcheck event because the revision has not changed.

This PR fixes this behavior.
Flux will always emit a healthcheck event if a Kustomization was not ready in the previous reconciliation.


stefanprodan (approved):
LGTM

Thanks @ordovicia

Issue #217, opened 2020-12-16
Support for impersonation when running GC pass


nairb774:
After upgrading to v0.5.2 to pick up https://github.com/fluxcd/kustomize-controller/issues/209 I noticed GC still wasn't working - no errors, no messages. I think I've narrowed it down to the lack of impersonation support as noted here: https://github.com/fluxcd/kustomize-controller/blob/37e1c9847335016808d3003a61fe4e5322aac275/controllers/kustomization_controller.go#L995

Mostly filing the ticket to raise awareness of the missing feature, and hopefully get eyes towards having it addressed. Thanks!

stefanprodan:
> I noticed GC still wasn't working - no errors, no messages.

This is not related to impersonation, it should work regardless of the service account. Can you please provide an example?

stefanprodan:
@nairb774 GC should be fixed in flux2 v0.5.7, to clean up any stale objects you have to make a change to any YAML in Git and the when the new revision is reconciled will trigger a GC sweep. If you have any more issues with GC, you can now enable debug logging with `flux install/bootstrap --log-level=debug`.

nairb774:
Running v0.5.3 with debug logs, I get the following log entries when GC is triggered:

```json
{
  "msg": "gc query failed for DestinationRule: destinationrules.networking.istio.io is forbidden: User \"system:serviceaccount:flux-system:flux-deployer\" cannot list resource \"destinationrules\" in API group \"networking.istio.io\" in the namespace \"kafka\"",
  "level": "info",
  "kustomization": "flux-system/deploy-ns-kafka",
  "logger": "controllers.Kustomization",
  "ts": "2020-12-20T19:11:03.821Z"
}
{
  "msg": "gc query failed for Sidecar: sidecars.networking.istio.io is forbidden: User \"system:serviceaccount:flux-system:flux-deployer\" cannot list resource \"sidecars\" in API group \"networking.istio.io\" in the namespace \"kafka\"",
  "level": "info",
  "kustomization": "flux-system/deploy-ns-kafka",
  "logger": "controllers.Kustomization",
  "ts": "2020-12-20T19:11:03.822Z"
}
{
  "msg": "gc query failed for ServiceEntry: serviceentries.networking.istio.io is forbidden: User \"system:serviceaccount:flux-system:flux-deployer\" cannot list resource \"serviceentries\" in API group \"networking.istio.io\" in the namespace \"kafka\"",
  "level": "info",
[...]

Issue #214, opened 2020-12-16
controller-runtime v0.6.4: v1.ListOptions is not suitable for converting


stefanprodan:
After upgrading from controller-runtime v0.6.3 to v0.6.4 listing HelmReleases is no longer possible:

```
v1.ListOptions is not suitable for converting to \"helm.toolkit.fluxcd.io/v2beta1\" in scheme \"pkg/runtime/scheme.go:101
```

This breaks kstatus heath checks and pruning.

stefanprodan:
Fix in controller-runtime v0.7.0 https://github.com/kubernetes-sigs/controller-runtime/pull/1238

We need to upgrde fluxcd/pkg/runtime before we can fix it in kustomize-controller

PR #221, opened 2020-12-17
Upgrade controller-runtime to v0.7.0

stefanprodan:
This PR upgrades the [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime) dependency to `v0.7.0`,  fix: #214.

Changes:
- Loggers have been removed from the reconcilers and are now retrieved from the `context.Context` passed to the `Reconcile` method and downwards functions.
- Logger configuration flags are now bound to the flag set using `BindFlags` from `runtime/logger`, ensuring the same contract across GitOps Toolkit controllers, and the `--log-json` flag has been deprecated in favour of the `--log-encoding=json` default.
- The `ChangePredicate` from `runtime` has changed to a `ReconcilateAtChangedPredicate`, and is now chained with the `GenerationChangedPredicate` from `controller-runtime` using `predicate.Or`.
- Signatures that made use of `runtime.Object` have changed to `client.Object`, removing the requirement to e.g. call `runtime.Object#Object`.
- The `client.MatchingField` function was deprecated, and has been replaced with `client.MatchingFields{}`.
- The `leader-election-role` was changed, as leader election now works via the `coordination/v1` API.
- The `SourceIndexKey` constant has been split out into `{GitRepository,Bucket}IndexKey` constants.
- The `util.ObjectKey` was added to easily construct a `client.ObjectKey` / `types.NamespacedName` from a `metav1.Object`.

Impersonation changes:

- The impersonation was refactored into `KustomizeImpersonation`.
- Run GC and health checks using service account impersonation, fix: #217.

Ref: https://github.com/fluxcd/flux2/issues/613


controllers/kustomization_controller.go
```diff
@@ -688,10 +703,10 @@ func (r *KustomizationReconciler) getServiceAccountToken(kustomization kustomize
 	return token, nil
 }
 
-func (r *KustomizationReconciler) apply(kustomization kustomizev1.Kustomization, dirPath string) (string, error) {
+func (r *KustomizationReconciler) apply(ctx context.Context, kustomization kustomizev1.Kustomization, dirPath string) (string, error) {
 	start := time.Now()
 	timeout := kustomization.GetTimeout() + (time.Second * 1)
-	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	applyCtx, cancel := context.WithTimeout(context.Background(), timeout)
```
phillebaba, line 709: Why not use the context passed to the function to create the timeout context?

controllers/kustomization_controller.go
```diff
@@ -749,43 +761,41 @@ func (r *KustomizationReconciler) apply(kustomization kustomizev1.Kustomization,
...
+func (r *KustomizationReconciler) applyWithRetry(ctx context.Context, kustomization kustomizev1.Kustomization, revision, dirPath string, delay time.Duration) (string, error) {
+	changeSet, err := r.apply(ctx, kustomization, dirPath)
 	if err != nil {
 		// retry apply due to CRD/CR race
 		if strings.Contains(err.Error(), "could not find the requested resource") ||
 			strings.Contains(err.Error(), "no matches for kind") {
-			r.Log.Info("retrying apply",
-				"error", err.Error(),
-				"kustomization", fmt.Sprintf("%s/%s", kustomization.GetNamespace(), kustomization.GetName()))
+			(logr.FromContext(ctx)).Info("retrying apply", "error", err.Error())
```
phillebaba, line 770: I am wondering if this is safe to do, as FromContext will return nil if no logger is found in the context? Or is it just fine to assume that the context will always be right.
https://github.com/go-logr/logr/blob/4fa77cb7175ca349567983fd9cc9fd94462f41fa/logr.go#L201

Using FromContextOrDiscard might be a bit better as it returns a discarding logger in the worst case scenario.

main.go
```diff
@@ -73,13 +72,18 @@ func main() {
 			"Enabling this will ensure there is only one active controller manager.")
 	flag.IntVar(&concurrent, "concurrent", 4, "The number of concurrent kustomize reconciles.")
 	flag.DurationVar(&requeueDependency, "requeue-dependency", 30*time.Second, "The interval at which failing dependencies are reevaluated.")
-	flag.StringVar(&logLevel, "log-level", "info", "Set logging level. Can be debug, info or error.")
-	flag.BoolVar(&logJSON, "log-json", false, "Set logging to JSON format.")
 	flag.BoolVar(&watchAllNamespaces, "watch-all-namespaces", true,
 		"Watch for custom resources in all namespaces, if set to false it will only watch the runtime namespace.")
+	flag.Bool("log-json", false, "Set logging to JSON format.")
+	flag.CommandLine.MarkDeprecated("log-json", "Please use --log-level=json instead.")
```
phillebaba, line 78: Where is the `--log-level` flag defined?

controllers/kustomization_controller.go
stefanprodan: Because we want to respect the timeout set in spec.timeout for apply, otherwise it could hand forever

controllers/kustomization_controller.go
stefanprodan: If the context doesn’t have a logger, then it should panic, note that we set the logger in main.go so the possibility for this is zero.

main.go
stefanprodan: It is a typo, it should be log format, these flags come from our pkg/runtime

cc @hiddeco this is what i meant that flags from packages are magically confusing :)

phillebaba: I was suspecting that the flag was set in pkg. I would have to agree that it is confusing. Sure it is nice that it happens auto magically but its better at times to be explicit in what is happening.

controllers/kustomization_controller.go
phillebaba: Ok just wanted to check that it was an active choice. I have never been a fan of passing loggers in a context but it seems to be the way a lot of projects are going.

controllers/kustomization_controller.go
phillebaba: I think i did a bad job explaining my thoughts.

Instead of doing the first why not do the seconds?
```
applyCtx, cancel := context.WithTimeout(context.Background(), timeout)
```
vs.
```
applyCtx, cancel := context.WithTimeout(ctx, timeout)
```

stefanprodan: Ahh yes that makes sense, thanks 👍

controllers/kustomization_controller.go
hiddeco: The logger is not passed from the context, but the metadata for the logger is collected from the context.

main.go
hiddeco: > Sure it is nice that it happens auto magically but its better at times to be explicit in what is happening.

Not if all controllers are expected to follow the same contract, the magic that happens here assures they do and are always aligned. Which makes the maintainer burden of thinking about the magic more justified to me than having misaligned flags across controllers.

api/v1beta1/kustomization_types.go
```diff
@@ -238,9 +238,9 @@ func (in *Kustomization) GetStatusConditions() *[]metav1.Condition {
 }
 
 const (
-	// SourceIndexKey is the key used for indexing kustomizations
+	// GitRepositoryIndexKey is the key used for indexing kustomizations
 	// based on their Git sources.
-	SourceIndexKey string = ".metadata.source"
+	GitRepositoryIndexKey string = ".metadata.git"
```
hiddeco, line 243: I would use the full name here (`.metadata.gitRepository`).

main.go
phillebaba: As long as there is a good reason I think it is fine to decide that the pros outweigh the cons. My thought is just that the log options will probably not change often or ever, but I may be wrong. Now that I know that this happens it wont affect me.

config/rbac/leader_election_role.yaml
```diff
@@ -4,23 +4,15 @@ kind: Role
...
-  - delete
-- apiGroups:
-  - ""
-  resources:
-  - configmaps/status
-  verbs:
-  - get
-  - update
-  - patch
+  - apiGroups:
```
hiddeco, line 7: I think this should still include all the permissions required for the leader election to work, even if they are magically inherited by the cluster admin role the controller has.

stefanprodan: What is missing?

stefanprodan: found it in SC, fixed!

blaquezzz (approved):

hiddeco (approved):

PR #222, opened 2020-12-18
Release v0.5.3

stefanprodan:


hiddeco (approved):

bergemalm:
This caused healthy health checks to be emitted in every reconciliation interval. Had to revert to 0.5.2.

stefanprodan:
It should alert only of the previous health check failed, implemented in https://github.com/fluxcd/kustomize-controller/pull/219

@bergemalm please open an issue if you see a different behavior

Issue #223, opened 2020-12-23
Directory called Kustomization incorrectly implies kustomization


mflendrich:
Hi there! :wave: 

It seems that I have hit a corner case in Kustomization detection.

## Summary
Current behavior: kustomization-controller fails if it encounters a **directory** called `Kustomization` in the repo.
Expected behavior:
1. reconciliation not failing
2. flux not refusing to descend into `Kustomization/`'s parent directory

I believe that this happens because the condition of [this `if`](https://github.com/fluxcd/kustomize-controller/blob/072cd12aa911854d55a0e22610f637524e51a8f7/controllers/kustomization_generator.go#L148) should only fire for non-directories.

## Reproduction
Suppose that I have `source-controller` and `kustomization-controller` running in my cluster and the following manifests applied:

```yaml
apiVersion: source.toolkit.fluxcd.io/v1beta1
kind: GitRepository
metadata:
  name: cluster-contents
  namespace: something
spec:
  url: https://github.com/some-org/some-repo.git
  interval: 30s
  secretRef:
    name: some-https-basic-auth-secret
```

```yaml
apiVersion: kustomize.toolkit.fluxcd.io/v1beta1
kind: Kustomization
metadata:
  name: cluster-contents
  namespace: something
spec:
  interval: 30s
  path: "./"
  prune: true
  sourceRef:
    kind: GitRepository
    name: cluster-contents
```

The repo contains only plain Kubernetes manifests (i.e. no Kustomizations, just plain k8s resources). That repo follows a naming convention for resource filenames: `/<metadata.namespace>/<kind>/<metadata.name>.yaml`

This works flawlessly until I add a resource of `Kustomization` kind (the use case is that I want to manage my flux2 `Kustomization`s from this repo). Per my naming convention, it lands in a file called along the lines of `/michals-namespace/Kustomization/michals-kustomization.yaml`. Once a directory called `Kustomization` gets pushed to the repository, reconciliation of the `Kustomization` resource `cluster-contents` begins to fail:

```
Status:
  Conditions:
    Last Transition Time:  2020-12-23T21:04:46Z
    Message:               kustomize build failed: accumulating resources: 2 errors occurred:
                           * accumulateFile error: "accumulating resources from './gitops': read /tmp/cluster-contents240645327/gitops: is a directory"
[...]

PR #224, opened 2020-12-24
No longer treat dirs as kustomization files

mflendrich:
Fixes https://github.com/fluxcd/kustomize-controller/issues/223
Signed-off-by: Michał Flendrich <michal@flendrich.pro>

Before this PR, kustomize-controller (incorrectly) takes a directory named `Kustomization` as a kustomization file, with the following consequences:
- the directory which contains the directory called `Kustomization` is erroneously omitted from processing
- trying to open the `Kustomization` directory as a file results in an error, reported by kustomization-controller

After this PR, when traversing the filesystem tree looking for Kustomization files, kustomization-controller doesn't try to take directories as kustomization files.

Testing: I believe that this change deserves an e2e test case, but apparently, in the current shape of e2e tests, this requires a GitHub repository with a "malicious" directory called Kustomization. I'll happily accept maintainers' guidance as to possible ways to test this change.

### DIFF
--- a/controllers/kustomization_generator.go
+++ b/controllers/kustomization_generator.go
@@ -145,7 +145,7 @@ func (kg *KustomizeGenerator) generateKustomization(dirPath string) error {
 				// If a sub-directory contains an existing kustomization file add the
 				// directory as a resource and do not decend into it.
 				for _, kfilename := range konfig.RecognizedKustomizationFileNames() {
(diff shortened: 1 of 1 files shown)


hiddeco (approved):
LGTM, thanks 🙏

stefanprodan (approved):

PR #233, opened 2021-01-12
Add QPS and Burst kube api client parameters as CLI arguments

playertwo:
If you have a lot of flux kustomizations you may face with getting stuck on reconciliation. Throttling errors is cause. You may increase QPS and Burst values and throttling go away

### DIFF
--- a/main.go
+++ b/main.go
@@ -63,6 +63,8 @@ func main() {
 		requeueDependency    time.Duration
 		logOptions           logger.Options
 		watchAllNamespaces   bool
+		kubeapiQPS           float64
+		kubeapiBurst         int
 	)
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&eventsAddr, "events-addr", "", "The address of the events receiver.")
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -75,6 +77,8 @@ func main() {
 	flag.BoolVar(&watchAllNamespaces, "watch-all-namespaces", true,
 		"Watch for custom resources in all namespaces, if set to false it will only watch the runtime namespace.")
 	flag.Bool("log-json", false, "Set logging to JSON format.")
+	flag.Float64Var(&kubeapiQPS, "kube-api-qps", 20.0, "QPS to use while talking with kubernetes API.")
```
stefanprodan, line 80: These values are the defaults in `ctrl.GetConfigOrDie()`, so why supply them as flag defaults?

In Flagger based on a several load tests, we decided to set the defaults to 100/250: https://github.com/fluxcd/flagger/blob/main/cmd/flagger/main.go#L88-L89

playertwo: My logic was not to broke something. I don't know possibly side effects of changing default values. This values worked before PR, and they will work after. If you want to change it, you could see actual values and decide how you want to increase them. 

I can change it to 100/250 if you want

stefanprodan: Let's go with the same defaults cert-manager uses 20/50 https://github.com/jetstack/cert-manager/blob/master/cmd/controller/app/options/options.go#L103-L104

stefanprodan (approved):
LGTM

Thanks @playertwo

PR #238, opened 2021-01-15
Look for all accepted Kustomization filenames

hiddeco:
Fixes: #225 

Before this commit we only checked if a `kustomization.yaml` existed at
the root of the given directory, this caused problems when people for
example used `.yml` as the extension, as the generated
`kustomization.yaml` would conflict with the `.yml` file.

After this commit all recognized Kustomization filenames as listed by
Kustomize itself are accepted, including files _without_ an extension
(`Kustomization`).

### DIFF
--- a/controllers/kustomization_generator.go
+++ b/controllers/kustomization_generator.go
@@ -35,8 +35,7 @@ import (
 )
 
 const (
-	kustomizationFileName = "kustomization.yaml"
-	transformerFileName   = "kustomization-gc-labels.yaml"
+	transformerFileName = "kustomization-gc-labels.yaml"
 )
 
 type KustomizeGenerator struct {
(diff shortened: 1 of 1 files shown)


stefanprodan (approved):
LGTM

Thanks @hiddeco 🍰

controllers/kustomization_generator.go
```diff
@@ -129,7 +129,14 @@ func checkKustomizeImageExists(images []kustypes.Image, imageName string) (bool,
 
 func (kg *KustomizeGenerator) generateKustomization(dirPath string) error {
 	fs := filesys.MakeFsOnDisk()
-	kfile := filepath.Join(dirPath, kustomizationFileName)
+
+	// Determine if there already is a Kustomization file at the root,
+	// as this means we do not have to generate one.
+	for _, kfile := range konfig.RecognizedKustomizationFileNames() {
+		if _, err := os.Stat(filepath.Join(kfile)); err == nil {
```
stefanprodan, line 136:
```suggestion
		if _, err := os.Stat(filepath.Join(dirPath, kfile)); err == nil {
```

PR #241, opened 2021-01-17
Update kustomize/api to v0.7.2 and disable kyaml

stefanprodan:
This PR updates `kustomize/api` to v0.7.2 and disables `kyaml` when running kustomize build due to upstream bugs:
- https://github.com/kubernetes-sigs/kustomize/issues/3446
- https://github.com/kubernetes-sigs/kustomize/issues/3480

Note that when `kyaml` is enabled, `Kustomizer.Run` fails with `Error: json: unsupported type: map[interface {}]interface {}` without any hint to which object caused the error, making it impossible for Flux users to track down the incompatible manifest.

Also if a manifest contains duplicate keys, kyaml panics thus crashing the whole controller 😢 

Fix: https://github.com/fluxcd/flux2/issues/729
Fix: #243

### DIFF
--- a/controllers/kustomization_generator.go
+++ b/controllers/kustomization_generator.go
@@ -28,6 +28,7 @@ import (
 	"sigs.k8s.io/kustomize/api/k8sdeps/kunstruct"
 	"sigs.k8s.io/kustomize/api/konfig"
 	"sigs.k8s.io/kustomize/api/krusty"
+	"sigs.k8s.io/kustomize/api/resmap"
 	kustypes "sigs.k8s.io/kustomize/api/types"
 	"sigs.k8s.io/yaml"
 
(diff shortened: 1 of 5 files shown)


hiddeco (commented):
Can you provide information on what this change incorporates and how it fixes the issue referenced to in the PR? Either in-code or in the commit message.

stefanprodan:
This doesn’t fix anything, it’s just a lib update.

controllers/kustomization_generator.go
```diff
@@ -281,3 +278,26 @@ func (kg *KustomizeGenerator) generateLabelTransformer(checksum, dirPath string)
...
 }
+
+// buildKustomization wraps krusty.MakeKustomizer with the following settings:
+// - disable kyaml due to critical bugs like:
+//	 - https://github.com/kubernetes-sigs/kustomize/issues/3446
+//	 - https://github.com/kubernetes-sigs/kustomize/issues/3480
+// - reorder the resources just before output (Namespaces and Cluster roles/role bindings first, CRDs before CRs, Webhooks last)
+// - load files from outside the kustomization.yaml root
+// - disable plugins except for the builtin ones
+// - prohibit changes to resourceIds, patch name/kind don't overwrite target name/kind
```
hiddeco, line 289: :100:

hiddeco (approved):

PR #249, opened 2021-01-22
Specify kubeconfig in tests

phillebaba:
Fixes an issue in the tests where the KC running would use the shells kubeconfig to when reconciling and no the in memory test cluster. The tests would pass before because there were no checks of the resources that were being added.

### DIFF
--- a/controllers/kustomization_controller_test.go
+++ b/controllers/kustomization_controller_test.go
@@ -27,6 +27,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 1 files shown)


controllers/kustomization_controller_test.go
```diff
@@ -55,6 +58,37 @@ var _ = Describe("KustomizationReconciler", func() {
 			err = k8sClient.Create(context.Background(), namespace)
 			Expect(err).NotTo(HaveOccurred(), "failed to create test namespace")
 
+			c := clientcmdapi.NewConfig()
```
phillebaba, line 61: Converting from a rest.Config to a api.Config was actually a bit tricky and there does not seem to be an easy way to do this. We need to convert each of the parameters from one config to another.

hiddeco: You may want to take a look at helm-controller (`internal` + the bits wiring up Helm), as there are some conversion utils there.

stefanprodan (approved):
LGTM

Thanks @phillebaba

Issue #246, opened 2021-01-20
Distinct reconciliation interval for failed (previous) reconciliation state


vladimirfx:
Use case:
Cluster infrastructure is described as root Kustomization resource which contains some mixed manifest type - Helm releases and plain yamls, Kustomize configs. Plain manifests and kustomizations depends on CDRs provided by Helm releases (Prometheus Operator, Graphana Operator, Logging Operator etc). So first reconciliation is 100% fail but retry after 1-2 minutes are success. 
General reconciliation interval = 2 minutes is too frequent (and costly) for our infrastructure and normally we wish set it to 30 minutes. So we need new (optional) parameter that sets shorten internal for reconciliation retry. I know it is workaround and not the solution for that problem but now it seems acceptable for us. 

Similar workaround already implemented in controller for simple case when CRD is included in kustomization build output (retry after 5 seconds https://github.com/fluxcd/kustomize-controller/blob/3a09ac62ec608b2ecbb81932efbbbb4f9fd57bc6/controllers/kustomization_controller.go#L349).

Another workaround which we try is using several depended Kustomization resources with health checks - one for operators Helm releases and others for remaining configs. But such configurations breaks encapsulation and actually forces us to have 2 config sets for one logical thing (for example: Postgres Operator installation configuration + Postgres Operator monitoring, logging configs). Moreover such independent reconcilers makes state dispersion (operators conf vs CRD usage conf) when new configuration committed which leads to downtime.

PR #250, opened 2021-01-22
Allow failed reconciliations to be scheduled at a different interval 

stefanprodan:
This PR adds an optional field `spec.retryInterval ` to the API that allows users to requeue a failed reconciliation at a different interval than `spec.Interval`.

Changes:
- Add `RetryInterval` as optional field to API. The spec.retryInterval is the interval at which to retry a previously failed reconciliation. When not specified, it defaults to the `spec.interval` value.
- Requeue a failed reconciliation based on retry interval 
- Add reconciliation unit test to verify that objects are create in cluster and labeled accordingly
 
 Fix: #246

### DIFF
--- a/api/v1beta1/kustomization_types.go
+++ b/api/v1beta1/kustomization_types.go
@@ -48,6 +48,11 @@ type KustomizationSpec struct {
 	// +required
 	Interval metav1.Duration `json:"interval"`
 
+	// The interval at which to retry a previously failed reconciliation.
+	// When not specified, it defaults to the Interval value.
+	// +optional
+	RetryInterval *metav1.Duration `json:"retryInterval,omitempty"`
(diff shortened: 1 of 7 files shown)


api/v1beta1/kustomization_types.go
```diff
@@ -48,6 +48,11 @@ type KustomizationSpec struct {
 	// +required
 	Interval metav1.Duration `json:"interval"`
 
+	// The interval at which to retry a previously failed reconciliation.
+	// When not specified, it defaults to the Interval value.
```
hiddeco, line 52:
```suggestion
	// Defaults to the Interval value.
```

(easier to comprehend)

hiddeco (approved):

PR #254, opened 2021-01-23
Tune Dockerfile for faster builds

sylr:
- Pool go mod download for all arch
- Use go cross build capabilities instead of building from containers running on different arch
- Produce lighter binaries using `-ldflags '-s -w'`
- Upgrade Alpine to 3.13

On my macbook pro (2,6 GHz 6-Core Intel Core i7) with clean cache:

- main @ 62972bec5449bb8ac797c23f922ed2bb9947cef8: Building 731.6s (69/69)
- tune-dockerfile @ 79b7f151bd7342b00c69b792ebfd41f7f437c7e2: Building 162.4s (41/41)


hiddeco:
> Produce lighter binaries using `-ldflags '-s -w'`

As we would like to offer a `pprof` endpoint sometime in the future, I do not think it is wise to add the `-w` flag and remove DWARF debug information.

One flag you may want to add, [that has been available since Go 1.13, is the `-trimpath` flag](https://golang.org/doc/go1.13#go-command).

Makefile
```diff
@@ -87,12 +87,13 @@ generate: controller-gen
 	cd api; $(CONTROLLER_GEN) object:headerFile="../hack/boilerplate.go.txt" paths="./..."
 
 # Build the docker image
-docker-build:
-	docker build . -t ${IMG}
+docker-buildx-build:
```
hiddeco, line 90: I would keep the targets short, and keep the `docker-{build,push}` targets.

Makefile
```diff
@@ -87,12 +87,13 @@ generate: controller-gen
...
+	# docker buildx build --load currently works with one one platform
+	docker buildx build . -t ${IMG} --platform=linux/amd64 --load
```
hiddeco, line 92:
```suggestion
	# top of Makefile
	IMG_ARCHS?=linux/amd64

	# ...omitted for brevity
	docker buildx build . -t ${IMG} --platform=${IMG_ARCHS} --load
```

Makefile
```diff
@@ -87,12 +87,13 @@ generate: controller-gen
...
 
 # Push the docker image
-docker-push:
-	docker push ${IMG}
+docker-buildx-push:
+	docker buildx build . -t ${IMG} --platform=linux/amd64,linux/arm/v7,linux/arm64 --push
```
hiddeco, line 96:
```suggestion
	docker buildx build . -t ${IMG} --platform=${IMG_ARCHS} --push
```

stefanprodan:
@sylr can I run this with Docker For Mac, do I need to setup anything, like buildx profiles?

sylr:
On mine I only had to do `docker buildx create --use`

Makefile
```diff
@@ -86,13 +87,19 @@ vet:
...
+comma := ,
+
 # Build the docker image
 docker-build:
-	docker build . -t ${IMG}
+ifneq (,$(findstring $(comma),$(IMG_ARCHS)))
+	@echo "docker buildx build --load currently works with one one platform"
+	@exit 2
+endif
+	docker buildx build . -t ${IMG} --platform=${IMG_ARCHS}  --load
```
hiddeco, line 98: Given the `Makefile` is mostly aimed at making someone's life easier while developing, having a default that errors is a no-go for me.

Dockerfile
```diff
@@ -1,44 +1,54 @@
...
+COPY go.mod go.sum ./
 
 # cache modules
 RUN go mod download
 
-# copy source code
-COPY main.go main.go
-COPY controllers/ controllers/
-COPY internal/ internal/
+# ------------------------------------------------------------------------------
```
hiddeco, line 14: Instead of putting a divider here, it is better to write a short comment on what comes next. If you need it to "stand out", you can format the doc block a bit, e.g:

```Dockerfile
#
# This stands out, and is where the build part starts
#
FROM --platform=$BUILDPLATFORM golang:1.15-alpine as builder

# While this is a short comment
ARG TARGETPLATFORM

# ...omitted for brevity

#
# This is the actual container image
#
FROM --platform=$TARGETPLATFORM alpine:3.13

# ...omitted for brevity
````

File: Makefile (hunk 4 above)
Comment (stefanprodan) [reply Makefile:98]: The issue with this is that new contributors need to figure out `buildx`. The Dockerfile is no longer compatible with `docker build` due to `FROM --platform`.

File: Dockerfile [hunk 6]
```diff
@@ -1,44 +1,54 @@
...
 
 # build
-RUN CGO_ENABLED=0 go build -a -o kustomize-controller main.go
+RUN CGO_ENABLED=0 \
+GOOS=$(echo "$TARGETPLATFORM" | cut -d '/' -f1 ) \
+GOARCH=$(echo "$TARGETPLATFORM" | cut -d '/' -f2 ) \
+GOARM=$(echo "$TARGETPLATFORM" | cut -d '/' -f3 | sed "s/^v//") \
+go build -a -trimpath -o kustomize-controller main.go
+
+# ------------------------------------------------------------------------------
```
hiddeco, line 34: See other comment about dividers.

Makefile
hiddeco: On my (Arch Linux) machine which was recently installed and has a vanilla Docker config, `make docker-build IMG_ARCHS=linux/amd64` runs without any configuration tweaks.

Dockerfile
sylr: I really think dividers help having a better vision of the different steps.

I can add comments if you want to but I'd like to keep the dividers for clarity.

hiddeco: Dividers only help (some*) if you know how to navigate the file, while explanatory comments are useful to anyone reading it.

Given changes to the `Dockerfile` are scarce and we are developing OSS, people looking at it will often not be familiar with what is in the file, and we should thus aim for the latter.

---

\* I _personally_ find that they just create noise

Dockerfile
```diff
@@ -1,44 +1,56 @@
-FROM golang:1.15-alpine as builder
+# go mod download cache stage
```
mvoitko, line 1: Why leave commented out code?

stefanprodan:
We've switched to `tonistiigi/xx` and now the multi-arch build takes 6m, instead of 42m. Closing this, thanks @sylr and sorry this didn't made it.

Issue #191, opened 2020-12-04
Incorrect state in health check event


phillebaba:
When deploying a Kustomization with a health check to a deployment in an incorrect image tag, the expected event generated is expected to be an error. The image cant be pulled which means that the deployment will fail immediately and never reach a healthy state. The correct error event is received if it is the initial deployment or if the previous commit also caused the health check to fail.

If a deployment is healthy but then a new commit changes the tag to something that cant be fetched, the health check will transition from healthy in one commit to unhealthy in another commit. The first event will have the wrong severity but the correct reason. After the next reconcile loop is run for the kustomization the health check will be evaluated again and the correct severity will be sent. I have only observed this issue in the git commit notifier, and it may only affect those notifiers as they will not ignore update events sent by the kustomize controller.

Below is the status messages sent for a commit to GitHub
```json
[
  {
    "url": "",
    "avatar_url": "",
    "id": 0,
    "node_id": "",
    "state": "failure",
    "description": "health check failed",
    "target_url": null,
    "context": "kustomization/apps",
    "created_at": "2020-12-04T11:18:19Z",
    "updated_at": "2020-12-04T11:18:19Z",
    "creator": {}
  },
  {
    "url": "",
    "avatar_url": "",
    "id": 0,
    "node_id": "",
    "state": "success",
    "description": "health check failed",
    "target_url": null,
    "context": "kustomization/apps",
    "created_at": "2020-12-04T11:17:18Z",
    "updated_at": "2020-12-04T11:17:18Z",
    "creator": {}
  }
]
```

Thank you @chlunde for initially reporting this bug and helping out with digging into it!

PR #255, opened 2021-01-24
Add tests for reconciliation events

phillebaba:
This adds tests that are meant to standardize the expected behavior for events generated by KC. In the end the test should act as a guide for any future changes to how events are generated.

Steps:
1. Create Kustomization which will deploy a Pod
2. Set Pod to an unhealthy state and wait for Kustomization to receive status condition for failed health check.
3. Trigger a reconcile and check the events to make sure that only one event has been created for the failed health check.
4. Set Pod to a healthy state and wait for Kustomization to receive a healthy status condition
5. Trigger a reconcile and check the events to make sure that only one event has been created for the health check.

TBD:
1. Verify external event recorder event message matches status
2. Update revision and make sure only one event is created for health check.

### DIFF
--- a/api/v1beta1/kustomization_types.go
+++ b/api/v1beta1/kustomization_types.go
@@ -101,6 +101,7 @@ type KustomizationSpec struct {
 	TargetNamespace string `json:"targetNamespace,omitempty"`
 
 	// Timeout for validation, apply and health checking operations.
+	// Will default to 1 min if set to less than 1 min.
 	// Defaults to 'Interval' duration.
 	// +optional
 	Timeout *metav1.Duration `json:"timeout,omitempty"`
--- a/config/crd/bases/kustomize.toolkit.fluxcd.io_kustomizations.yaml
+++ b/config/crd/bases/kustomize.toolkit.fluxcd.io_kustomizations.yaml
@@ -217,7 +217,8 @@ spec:
                 type: string
               timeout:
                 description: Timeout for validation, apply and health checking operations.
-                  Defaults to 'Interval' duration.
+                  Will default to 1 min if set to less than 1 min. Defaults to 'Interval'
+                  duration.
                 type: string
               validation:
                 description: Validate the Kubernetes objects before applying them
--- a/controllers/kustomization_controller_test.go
+++ b/controllers/kustomization_controller_test.go
@@ -25,11 +25,15 @@ import (
 	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
+	eventv1beta1 "k8s.io/api/events/v1beta1"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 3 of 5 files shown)


controllers/kustomization_controller.go
```diff
@@ -237,7 +237,7 @@ func (r *KustomizationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	if reconcileErr != nil {
 		// record the reconciliation error
 		r.recordReadiness(ctx, reconciledKustomization)
-		return ctrl.Result{RequeueAfter: kustomization.Spec.Interval.Duration}, reconcileErr
+		return ctrl.Result{RequeueAfter: kustomization.Spec.Interval.Duration}, nil
```
phillebaba, line 240: If we return an error here the resource will be immediately reconciled. So what is the point of setting RequeueAfter? Reconciling immediately will mean that the Kustomization will not reach a unhealthy condition until the backoff is large enough.

stefanprodan: Hmm something is very wrong if it doesn’t schedule at the RetryInteval. Please undo this change as it breaks the retry behavior.

phillebaba: I am trying to find docs about the Reconcile behavior to be sure, but it feels like the requeue after is ignored if an error is returned. When a health check timeout error occurs the resources will reconcile again after 1 ms. As I understand it that is not what we want to happen?

stefanprodan: We want to retry the reconciliation based on what GetRetryInterval gives us. Maybe something changed in controller-runtime 0.8, is your PR based on the latest main branch?

phillebaba: Yep i rebased from main yesterday. So it might be that or the in memory cluster not behaving like a "normal" cluster. I am seeing the same behavior in a real cluster where the condition will go from progressing to progressing as it is immediately rescheduled. Also the way we are patching status updates means that we will always override conditions in that status. The only way to determine if the health check timed out is to check the events.

stefanprodan: We need to introduce a new condition called heathCheck that records the last check result. But I would not mix this into this PR. First let’s figure out why the retry interval doesn’t work.

phillebaba: I will investigate why it is reconciled immediately. So all we need to do is change the Type value in the status condition. The reason the conditions replaced is because the merge key for conditions is the Type value.

stefanprodan: The ready condition is reset by progressing and that’s ok, we need to create a dedicated condition for health checks.

phillebaba: Looking into controller-runtime, if you return an error from a reconcile call the result will be ignored. This should be the right code that does the error handling.
https://github.com/kubernetes-sigs/controller-runtime/blob/9e78e653228851684b6b9cdabe5aae8559fe3722/pkg/internal/controller/controller.go#L297

I am not sure if it has always been like this or if it is a change introduced in the latest version. Either way we need to fix KC so that it actually waits the interval time.

stefanprodan: Fixed in #256, please rebase.

phillebaba:
Its better to try to merge this PR which adds the test to avoid regression and then I can verify that #191 is not present in a later PR,

stefanprodan:
> There is still the issue that the health passed even is sent immediatly when the kustomization is reconciled, even if the health check does not pass.

I don't see how is that possible, on a health check error we return early:
https://github.com/fluxcd/kustomize-controller/blob/main/controllers/kustomization_controller.go#L695

stefanprodan:
I've added various tests for the events in the last months, closing this.

PR #256, opened 2021-01-25
Fix reconciliation retry scheduler

stefanprodan:
Reconciliation failures should be scheduled at the specified retry interval. Log the reconciliation error instead of returning it, so that controller-runtime doesn't requeue immediately. 

Followup: #250

### DIFF
--- a/controllers/kustomization_controller.go
+++ b/controllers/kustomization_controller.go
@@ -156,7 +156,7 @@ func (r *KustomizationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 			r.recordReadiness(ctx, kustomization)
 			log.Info(msg)
 			// do not requeue immediately, when the source is created the watcher should trigger a reconciliation
-			return ctrl.Result{RequeueAfter: kustomization.Spec.Interval.Duration}, nil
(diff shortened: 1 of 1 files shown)


controllers/kustomization_controller.go
```diff
@@ -215,36 +215,33 @@ func (r *KustomizationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-		return ctrl.Result{RequeueAfter: kustomization.GetRetryInterval()}, reconcileErr
+		log.Error(reconcileErr, fmt.Sprintf("Reconciliation failed after %s, next try in %s",
+			time.Now().Sub(reconcileStart).String(),
+			kustomization.GetRetryInterval().String()),
+			"revision",
+			source.GetArtifact().Revision)
+		r.event(ctx, reconciledKustomization, source.GetArtifact().Revision, events.EventSeverityError,
+			reconcileErr.Error(), nil)
+		return ctrl.Result{RequeueAfter: kustomization.GetRetryInterval()}, nil
+	} else {
```
hiddeco, line 234: Given the `if` does a `return`, this `else` isn't strictly required?

hiddeco (approved):

PR #258, opened 2021-01-25
impersonation: Clear config.BearerTokenFile when setting BearerToken.

nairb774:
When running in a cluster, the BearerTokenFile is set to point to
`/var/run/secrets/kubernetes.io/serviceaccount/token` where the service
account's token is auto-mounted. If this value is not cleared, the
setting of the BearerToken field will have no effect. Relevant
documentation:
https://pkg.go.dev/k8s.io/client-go@v0.20.2/rest#Config.BearerTokenFile


stefanprodan (approved):
LGTM 

Thanks @nairb774 good catch!

PR #260, opened 2021-01-29
Support impersonation for validation

splushii:
Having a similar setup as in https://github.com/fluxcd/kustomize-controller/issues/217. Running kustomize-controller with minimal privileges and using `serviceAccountName` to "elevate" privileges for Kustomizations.

Kustomization validation was performed with the kustomize-controller service account, which with the above setup doesn't have the required privileges. This PR adds support for service account impersonation during validation.

Same approach as for apply(): https://github.com/splushii/kustomize-controller/blob/d0f2dc6e4aab2cd6dd454adec0698d5ea2bd644e/controllers/kustomization_controller.go#L595-L604


stefanprodan (approved):
LGTM

Thanks @splushii good catch!

Issue #236, opened 2021-01-14
Health check event is emitted in every reconciliation cycle


bergemalm:
Hi,

With merge of #219 the "health check passed"-event/notification is emitted in every reconciliation cycle for all Kustomizations always.

https://github.com/fluxcd/kustomize-controller/blob/5356194f5a2791af0d14c0b2f429e653f7440065/controllers/kustomization_controller.go#L699
`ready` will always be false since `readiness.Status` will be `Unknown` at the stage of checkHealth.

```
.
.
Status:
  Conditions:
    Last Transition Time:   2021-01-13T08:38:10Z
    Message:                reconciliation in progress
    Reason:                 Progressing
    Status:                 Unknown
    Type:                   Ready
.
.
```
Status is updated later in the reconciliation cycle.
https://github.com/fluxcd/kustomize-controller/blob/5356194f5a2791af0d14c0b2f429e653f7440065/controllers/kustomization_controller.go#L385

Am I missing something?

Example of `kind: Kustomization` used:

```
---
apiVersion: kustomize.toolkit.fluxcd.io/v1beta1
kind: Kustomization
metadata:
  name: my-app
  namespace: flux-system
spec:
  interval: 15m
  path: ./mgmt/namespaces/my-namespace/my-app
  prune: true
  sourceRef:
    kind: GitRepository
    name: flux-system
  healthChecks:
    - apiVersion: helm.toolkit.fluxcd.io/v2beta1
      kind: HelmRelease
      name: my-app
      namespace: my-namespace
  timeout: 5m
  validation: client
```

PR #262, opened 2021-02-02
Add Healthy status condition

stefanprodan:
- record the last health assessment result in a dedicated status condition
- use the condition status when issuing events to prevent notifications spam (fix: #236)

### DIFF
--- a/api/v1beta1/kustomization_types.go
+++ b/api/v1beta1/kustomization_types.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1beta1
 
 import (
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


api/v1beta1/kustomization_types.go
```diff
@@ -184,6 +185,15 @@ func KustomizationProgressing(k Kustomization) Kustomization {
 	return k
 }
 
+// SetKustomizationHealthiness sets the HealthyCondition status for a Kustomization.
+func SetKustomizationHealthiness(k *Kustomization, status metav1.ConditionStatus, reason, message string) {
+	if len(k.Spec.HealthChecks) > 0 {
+		meta.SetResourceCondition(k, HealthyCondition, status, reason, trimString(message, MaxConditionMessageLength))
+	} else {
+		apimeta.RemoveStatusCondition(k.GetStatusConditions(), HealthyCondition)
+	}
```
hiddeco, line 194: Maybe cleaner to always remove and then conditionally add back?

stefanprodan: I don't think this is the intended usage of the apimeta. If a condition exists, we should update it instead of removing and creating it from scratch.

hiddeco (approved):

Issue #237, opened 2021-01-15
Feature Request: Allow disabling of prune on certain resources


bcbrockway:
We have our EKS clusters created by Terraform. Terraform then installs Flux which takes over from there but one of the repos that Flux points at has manifests for the kube-system, kube-public, default, etc. namespaces (they add a couple of labels). As soon as it applies the manifest Flux also becomes the "owner" of the namespace so we run the risk of it trying to delete kube-system if we delete the Kustomization object from flux-system and getting in a state when Kubernetes says "don't you be touching my damn namespaces".

Options at the moment:

1. We could split those specific manifests into their own folder and disable prune on a new Flux Kustomization
2. We could disable prune altogether for that repo
3. We could add the label using the Kubernetes provider in Terraform

For now we're gonna go with option 3 but ideally we'd like to be able to add an annotation to a resource (`kustomize.toolkit.fluxcd.io/prune: false` or something) to tell Flux not to touch it. What do you think?

stefanprodan:
I think you should split the manifests into directories and disable pruning  with `spec.prune: false`.

PR #267, opened 2021-02-07
Allow disabling of prune on certain resources

stefanprodan:
This PR allows disabling pruning for certain resources by either labeling or annotating them with:

```yaml
kustomize.toolkit.fluxcd.io/prune: disabled
```

Fix: #237

### DIFF
--- a/controllers/kustomization_gc.go
+++ b/controllers/kustomization_gc.go
@@ -19,7 +19,6 @@ package controllers
 import (
 	"context"
 	"fmt"
-	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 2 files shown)


controllers/kustomization_gc.go
```diff
@@ -138,6 +142,21 @@ func (kgc *KustomizeGarbageCollector) isStale(obj unstructured.Unstructured) boo
...
+
+	if obj.GetLabels()[key] == val {
+		return true
+	}
+
+	if obj.GetAnnotations()[key] == val {
+		return true
+	}
+
+	return false
```
hiddeco, line 157: Can't this simply be reduced to the following?

```suggestion
	return obj.GetLabels()[key] == val || obj.GetAnnotations()[key] == val
```

hiddeco (approved):

PR #268, opened 2021-02-08
Add source kind and name to not found error

stefanprodan:


controllers/kustomization_controller.go
```diff
@@ -147,7 +147,7 @@ func (r *KustomizationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	source, err := r.getSource(ctx, kustomization)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
-			msg := "Source not found"
+			msg := fmt.Sprintf("Source '%s/%s' not found", kustomization.Spec.SourceRef.Kind, kustomization.Spec.SourceRef.Name)
```
hiddeco, line 150: Maybe we should add a `String()` method to the `SourceRef` type? We are doing this in many places...

hiddeco (approved):

controllers/kustomization_controller.go
stefanprodan: Good point!

hiddeco (approved):

Issue #122, opened 2020-09-29
Option for using `kubectl replace --force` when fail to apply


ordovicia:
When we want change immutable fields (e.g. label selector in Deployment), currently we need to take two steps:

1. Enable garbage collection for a resource and delete the resource from a repository
2. Add the resource again to the repository with a new definition

It would be great if kustomize-controller does the above operation in one reconciliation by using `kubectl replace --force` command.

stefanprodan:
The kustomize controller doesn't apply one resource at a time, but all them bulk, I don't see how we could use replace.

stealthybox:
If you are using `kustomization.spec.prune: true`, you could do this in a single commit like so:
```diff
❯ git diff
diff --git lib/podinfo/deployment.yaml lib/podinfo/deployment.yaml
index 7357a05..9476847 100644
--- lib/podinfo/deployment.yaml
+++ lib/podinfo/deployment.yaml
@@ -1,7 +1,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: podinfo
+  name: podinfo-v2
 spec:
   minReadySeconds: 3
   revisionHistoryLimit: 5
@@ -13,6 +13,7 @@ spec:
   selector:
     matchLabels:
       app: podinfo
+      version: abcd
   template:
     metadata:
       annotations:
@@ -20,6 +21,7 @@ spec:
         prometheus.io/port: "9797"
       labels:
         app: podinfo
+        version: abcd
     spec:
       containers:
       - name: podinfod
```
This will delete the old deployment and create a new one with a slightly different name.
This explicitly gives you a new object while keeping your git history clean.

You're probably doing something similar anyway when updating ConfigMaps since old copies of objects need to exist for ReplicaSets to rollback properly.
______

Fields are immutable for a reason, so we should consider the consequences before we build a behavior that auto-replaces objects. There may be a sensible thing flux can do here though that allows the user to indicate that they want that behavior for a resource when they need it. Maybe an annotation would work.

Note: doing something special for a failed object is dependent on being able to have granular, structured status for each one as a result of the apply.
This is tracked in https://github.com/fluxcd/flux2/issues/46

ordovicia:
Thank you @stealthybox. I understand that it is difficult to build a right design for the auto-replace feature.

PR #271, opened 2021-02-13
Add support for recreating objects when immutable fields are updated 

relu:
Allow passing `--force` to kubectl apply. Useful when dealing with
immutable field changes in resources.

I have a use case in which I'm dealing with migration kustomization that is a dependency for an application overlay. The migration kustomization is using a job and it would be really useful if the controller would support force replacements.

Note that this does not work with server-side validation.

Fix: #122


controllers/kustomization_controller.go
```diff
@@ -550,8 +550,8 @@ func (r *KustomizationReconciler) validate(ctx context.Context, kustomization ku
 	applyCtx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
-	cmd := fmt.Sprintf("cd %s && kubectl apply -f %s.yaml --timeout=%s --dry-run=%s --cache-dir=/tmp",
-		dirPath, kustomization.GetUID(), kustomization.GetTimeout().String(), kustomization.Spec.Validation)
+	cmd := fmt.Sprintf("cd %s && kubectl apply -f %s.yaml --timeout=%s --dry-run=%s --cache-dir=/tmp --force=%t",
```
relu, line 553: This will fail if `server` validation is used because `--force` is not supported. I've added it intentionally so that the user can get some feedback:
> validation failed: error: --dry-run=server cannot be used with --force

config/testdata/force/source.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: source.toolkit.fluxcd.io/v1beta1
+kind: GitRepository
+metadata:
+  name: force
+spec:
+  interval: 10m
+  url: https://gist.github.com/cd83f83eebee3217d02ee5d3b97a9908.git
```
relu, line 7: Probably need a better place to keep these.

stefanprodan:
I don't think `apply --force` does what you are after, in case of a Job, it has not effect. A forced apply, recreates the object only if patching fails (e.g. immutable field error). You need `replace --force` to restart the job and I don't think kustomize-controller should perform such an operation.

To run jobs as part of the reconciliation, we could offer something similar to Helm hooks:

```yaml
kind: Kustomization
metadata:
  name: app
spec:
  interval: 5m
  prune: true
  sourceRef:
    kind: GitRepository
    name: webapp
  path: "./webapp/deploy/"
  hooks:
    - type: pre-apply
      path: "./webapp/pre-apply-jobs/" 
    - type: post-apply
      path: "./webapp/post-apply-jobs/"
```

relu:
Actually, for my use case, this works just fine because the changes made are always going to affect immutable fields (the image tag would be updated, to be more precise). I've tested it locally and it works as intended, the job is being replaced and a new pod is started.
I think there's still value in this for situations when users don't want to manually intervene when dealing with immutable field updates (for deployments and other workload resources) but rather force the update, it can be a temporary thing in the context of a migration or reconfiguration (e.g. label selectors) that can be reverted later.

I really like the idea of adding support for hooks though, would be a great addition to the customize-controller.

config/testdata/force/source.yaml
stefanprodan: Please remove these from e2e, we should add unit tests using an immutable fields.

controllers/kustomization_controller.go
stefanprodan: I propose we skip validation when force is enabled. We need to update the field description with this behaviour.

relu: That's what I thought first as well but considering that client-side validation still works I thought it would be useful to have it enabled just for that. Do you think it's not really useful in that situation?

stefanprodan: I think we should downgrade the server-side dry-run to client, when force is enabled.

relu: Yes, that makes sense.

relu:
I don't understand this spec error:
```
{"reconciler group": "kustomize.toolkit.fluxcd.io", "reconciler kind": "Kustomization", "name": "6re3a", "namespace": "kustomization-testti307", "revision": "v1", "error": "validation failed: error: error validating \"792d37ed-59f5-4e58-9c5d-6a47d3b2defe.yaml\": error validating data: ValidationError(ConfigMap): unknown field \"immutable\" in io.k8s.api.core.v1.ConfigMap; if you choose to ignore these errors, turn validation off with --validate=false\n"}
```
Immutable ConfigMaps are supported since v1.18, we're on v1.20. I first thought I had a local issue but it's reproducible here as well.

stefanprodan:
@relu you need to configure Kind like so: 

```
      - name: Setup Kubernetes
        uses: engineerd/setup-kind@v0.5.0
        with:
          version: "v0.10.0"
          image: kindest/node:v1.20.2@sha256:8f7ea6e7642c0da54f04a7ee10431549c0257315b3a634f6ef2fecaaedb19bab
```

relu:
@stefanprodan kind is not used for these tests right now, it [defaults to envtest's mock cluster](https://github.com/fluxcd/kustomize-controller/blob/main/controllers/suite_test.go#L65) since `TEST_USE_EXISTING_CLUSTER` is not set to `true` in the workflow.

I'll try to use kind locally and see if it's still reproducible, I am currently running exactly that version you pointed to in your comment.

stefanprodan:
@relu I get now what you're saying, we need to figure out how to tell envtest to use a newer Kubernetes version.

hiddeco:
You can configure custom binaries to be used by setting `TEST_ASSET_*`, see: https://book.kubebuilder.io/reference/envtest.html

stefanprodan:
I think I found the issue, https://github.com/fluxcd/pkg/blob/main/actions/kubebuilder/entrypoint.sh#L5 this pulls old binaries.

relu:
Indeed, see: https://github.com/fluxcd/kustomize-controller/pull/271/checks?check_run_id=1903336363 (Debug failure section).
```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.2", GitCommit:"f5743093fd1c663cb0cbc89748f730662345d44d", GitTreeState:"clean", BuildDate:"2020-09-16T13:41:02Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0", GitCommit:"70132b0f130acc0bed193d9ba59dd186f0e634cf", GitTreeState:"clean", BuildDate:"2020-01-14T00:09:19Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}
```

stefanprodan:
Relu I think you can should use a Deployment and change the label in the template spec. We can't upgrade kubebuilder since it's alpha.

relu:
I don't think I can use workload resources, they will not reconcile on update and cause apply to time-out, I even tried with `Service` and got the same result. These work just fine for the initial apply during when they are created, but the second one will just time-out.:disappointed:
Do you happen to know any other non-workload resources that might have immutable fields?

stefanprodan:
@relu check out the pod tests that @phillebaba wrote here https://github.com/fluxcd/kustomize-controller/pull/255/files

relu:
I tried different resources `ConfigMap`, `Deployment`, `Job`, `Service`, and for all of them I get the same issue: `kubectl apply` gets stuck, timing-out during the second reconciliation (after the manifests are updated). This works perfectly fine if I test it on a real cluster so I think it's probably an issue with envtest itself.

stefanprodan:
Ok, so we can use your tests only after kubebuilder v3 gets released as it comes with a Kubernetes version that knows about immutable ConfigMap. Remove the tests from this branch, and maybe create a new WIP PR with only the tests, and we'll get back to it after kubebuilder v3.

relu:
I don't think that's necessarily the case, as I said earlier I tried with all sorts of other resources that have immutable fields and in all situations the behavior was the same. There's something weird going on that has kubectl wait endlessly for the patch to apply. I will try with the master version of kubebuilder just to confirm this, it might be just that indeed this new version will work just fine.

stefanprodan (approved):
LGTM

Thanks @relu 🏅

controllers/kustomization_controller.go
hiddeco: This is not described on the API field.

relu: Added. Thank you, @hiddeco!

stefanprodan:
@relu you need to run `make test` and push all changes.

Issue #274, opened 2021-02-15
Add substituteFrom from ConfigMap or Secret


angelnu:
The current substitute for DRY within a cluster or Kustomization but it still requires global settings (multiple clusters, multiple Kustomizations) to be duplicated.

So we need a way to keep those common substitutions centralized.

After talking with @stefanprodan in #272 it looks like the best option is to keep global settings in a ConfigMap or Secret. This would also be consistent with the HelmReleases `valueFrom`.

angelnu:
Many thanks @stefanprodan for the quick time-around! Will give it try next (need to find how to use the dev version of flux2)

stefanprodan:
You need to clone the repo and build the container image on your own. Then apply the CRDs from the main brach and replace kustomize-controller image on your cluster.

PR #275, opened 2021-02-16
Implement var substitution from ConfigMaps and Secrets

stefanprodan:
This PR adds an array field called `substituteFrom` to the post build API.  `SubstituteFrom` holds references to `ConfigMaps` and `Secrets` containing the variables (data keys) and their values (data values) to be substituted in the YAML manifests. 

You can disable the variable substitution for certain resources by either labeling or annotating them with:

```yaml
kustomize.toolkit.fluxcd.io/substitute: disabled
``` 

Example:

```yaml
kind: Kustomization
metadata:
  name: apps
spec:
  interval: 5m
  path: "./apps/"
  postBuild:
    substitute:
      cluster_env: "prod"
      cluster_region: "eu-central-1"
    substituteFrom:
      - kind: ConfigMap
        name: cluster-vars
      - kind: Secret
        name: cluster-secret-vars
```

Fix: #274
Followup: #253


controllers/kustomization_varsub.go
```diff
@@ -0,0 +1,90 @@
...
+	for _, reference := range kustomization.Spec.PostBuild.SubstituteFrom {
+		namespacedName := types.NamespacedName{Namespace: kustomization.Namespace, Name: reference.Name}
+		switch reference.Kind {
+		case "ConfigMap":
+			resource := &corev1.ConfigMap{}
+			if err := kubeClient.Get(ctx, namespacedName, resource); err != nil {
+				return nil, fmt.Errorf("substitute from 'ConfigMap/%s' error: %w", reference.Name, err)
+			}
+			for k, v := range resource.Data {
+				vars[k] = strings.Replace(v, "\n", "", -1)
```
hiddeco, line 49: Is the idea here to trim trailing white space? If so, wouldn't it be better to use `strings.TrimRight` or `strings.TrimSpace`?

stefanprodan: Te idea is to remove all the `/n` from the value, to avoid generating an invalid yaml, as the data value can look like this:
```
data:
  some_var: |
      
      bla
      
```

hiddeco (approved):

PR #276, opened 2021-02-16
Generate OpenAPI schema and validate samples

stefanprodan:
Generate the OpenAPI JSON schema for the Kustomization CRD and use it to validate the custom resource samples with kubeval.


stefanprodan:
Validation example here: https://github.com/fluxcd/flux2-kustomize-helm-example/pull/19

Issue #277, opened 2021-02-18
Ability to control verbosity of validation errors


laszlocph:
This issue has the exact same motivation as https://github.com/fluxcd/kustomize-controller/issues/190

https://github.com/fluxcd/kustomize-controller/issues/190 was resolved by two filter function:
- https://github.com/fluxcd/kustomize-controller/blob/5f966d0ff615658dc630a89837abeb1bf459d75a/controllers/kustomization_controller.go#L645
- https://github.com/fluxcd/kustomize-controller/blob/5f966d0ff615658dc630a89837abeb1bf459d75a/controllers/kustomization_controller.go#L648

and they handle the apply case perfectly.

I would like that the `validate` function gets the same treatment too. https://github.com/fluxcd/kustomize-controller/blob/5f966d0ff615658dc630a89837abeb1bf459d75a/controllers/kustomization_controller.go#L565

There, the kustomize output is returned without any treatment: https://github.com/fluxcd/kustomize-controller/blob/5f966d0ff615658dc630a89837abeb1bf459d75a/controllers/kustomization_controller.go#L601

resulting in a long output, where only the validation error is noteworthy:

```
validation failed: namespace/flux-system configured (dry run)
customresourcedefinition.apiextensions.k8s.io/alerts.notification.toolkit.fluxcd.io configured (dry run)
customresourcedefinition.apiextensions.k8s.io/buckets.source.toolkit.fluxcd.io configured (dry run)
customresourcedefinition.apiextensions.k8s.io/gitrepositories.source.toolkit.fluxcd.io configured (dry run)
customresourcedefinition.apiextensions.k8s.io/helmcharts.source.toolkit.fluxcd.io configured (dry run)
customresourcedefinition.apiextensions.k8s.io/helmreleases.helm.toolkit.fluxcd.io configured (dry run)
customresourcedefinition.apiextensions.k8s.io/helmrepositories.source.toolkit.fluxcd.io configured (dry run)
customresourcedefinition.apiextensions.k8s.io/kustomizations.kustomize.toolkit.fluxcd.io configured (dry run)
customresourcedefinition.apiextensions.k8s.io/providers.notification.toolkit.fluxcd.io configured (dry run)
customresourcedefinition.apiextensions.k8s.io/receivers.notification.toolkit.fluxcd.io configured (dry run)
serviceaccount/helm-controller configured (dry run)
serviceaccount/kustomize-controller configured (dry run)
serviceaccount/notification-controller configured (dry run)
[...]

PR #280, opened 2021-02-18
Extracting validation error from apply dry run output

laszlocph:
An attempt to treat validation error output the same way as apply output: reducing the noise by only returning the validation error.

Fixes: #277

Note: I'm not sure about the BDD framework used in the project. Let me know if I misuse it


laszlocph:
Okay, the suite failed as even for the `Utils Suite` it started a k8s and was doing stuff. I will revisit this later unless you have some pointers how to go with tests. I'm happy to just write a plain golang test without the framework.

stefanprodan:
@laszlocph writing a test with go stdlib is ok.

laszlocph:
> @laszlocph writing a test with go stdlib is ok.

thanks, I'm done with my changes.

stefanprodan (approved):
LGTM

Thanks @laszlocph can you please squash all commits.

PR #281, opened 2021-02-20
Avoid prompts on import by adding batch flag to gpg

davidkarlsen:
Fixes https://github.com/fluxcd/source-controller/issues/296

Signed-off-by: David J. M. Karlsen <david@davidkarlsen.com>


stefanprodan (approved):
LGTM

Thanks @davidkarlsen

PR #287, opened 2021-02-25
Update sigs.k8s.io/cli-utils to v0.22.2

hiddeco:
This is the first release that includes a patch of the
`CachingClusterReader` so that it continues on all list errors.

Ref: https://github.com/fluxcd/flux2/pull/990


stefanprodan (approved):
LGTM

Thanks @hiddeco

Issue #290, opened 2021-02-26
Variable substitution failed: missing closing brace when using variable substitution in Kustomization


ilya-git:
### Describe the bug

I have set up a simple variable substitution as described in the documentation from config map:

```
  postBuild:
    substituteFrom:
      - kind: ConfigMap
        name: env-config
```

Then I use it in helm release values:

```
values:
   secrets:
      userAssignedIdentityID: "${kubelet-identity-client-id}"
```

`kubectl get configmap`

```
apiVersion: v1
data:
  ...
  kubelet-identity-client-id: <a guid in here>
kind: ConfigMap
```

When Kustomization controller runs, I get the following error:

`var substitution failed for 'my-app': variable substitution failed: missing closing brace`

If I write a random name (var does not exist in config map) instead of `kubelet-identity-client-id`, then an empty string is written.

### Expected behavior

No error or another error that describes what is the problem (the closing brace seems to be in place at least right now)

### Additional context

- Kubernetes version: 1.20.2
- Git provider: Bitbucket
- Container registry provider: Azure Container Registry

Below please provide the output of the following commands:

```cli
flux --version: 0.8.2
flux check

► checking prerequisites
✗ flux 0.8.2 <0.9.0 (new version is available, please upgrade)
✔ kubectl 1.20.0 >=1.18.0-0
✔ Kubernetes 1.20.2 >=1.16.0-0
► checking controllers
✔ helm-controller: healthy
► ghcr.io/fluxcd/helm-controller:v0.7.0
✔ kustomize-controller: healthy
► ghcr.io/fluxcd/kustomize-controller:v0.8.1
✔ notification-controller: healthy
► ghcr.io/fluxcd/notification-controller:v0.8.0
✔ source-controller: healthy
► ghcr.io/fluxcd/source-controller:v0.8.1
✔ all checks passed
```

pmialon:
Hi, 

I have a similar problem, I generate ConfigMap in my kustomization.

```
---
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
configMapGenerator:
  - files:
      - application.yml=app-conf
    name: app-conf

```

but the file app-conf contain [spring placeholders](https://docs.spring.io/spring-boot/docs/1.5.6.RELEASE/reference/html/boot-features-external-config.html#boot-features-external-config-placeholders-in-properties) like this : 

```
app.name=MyApp
app.description=${app.name} is a Spring Boot application
```

I have the same error message: 
[...]

PR #291, opened 2021-02-27
Validate the var names before substitution

stefanprodan:
The name of a variable used in post build substitutions can contain only alphanumeric and underscore characters. This PR implements validation for the variable names using `^[_[:alpha:]][_[:alpha:][:digit:]]*$` regex. 

Fix: #290


hiddeco (approved):
LGTM, thank you @stefanprodan 💯

PR #289, opened 2021-02-26
Retry with exponential backoff when fetching artifacts

stefanprodan:
This PR implements retries with exponential backoff when fetching artifacts from source-controller. By default, the controller does 10 attempts within a 3.5 minutes window, the number of max retries can be configure using the `--http-retry` cmd arg.

This mitigates the alert spam (`i/o timeout err`) when source-controller becomes unavailable for a short period of time e.g. after an upgrade. 

Example error log:

```json
{"level":"error","ts":"2021-02-26T12:03:19.710+0200","logger":"controller.kustomization","msg":"Reconciliation failed after 3m35.317406763s, next try in 5m0s","reconciler group":"kustomize.toolkit.fluxcd.io","reconciler kind":"Kustomization","name":"podinfo","namespace":"flux-system","revision":"master/e43ebfa5bf4b87c46f2e1db495eb571cd398e2f7","error":"failed to download artifact, error: GET http://localhost:8080/gitrepository/flux-system/podinfo/e43ebfa5bf4b87c46f2e1db495eb571cd398e2f7.tar.gz giving up after 10 attempt(s): Get \"http://localhost:8080/gitrepository/flux-system/podinfo/e43ebfa5bf4b87c46f2e1db495eb571cd398e2f7.tar.gz\": dial tcp [::1]:8080: connect: connection refused"}
```

Ref: https://github.com/fluxcd/flux2/discussions/661 https://github.com/fluxcd/notification-controller/issues/76

PR #293, opened 2021-03-05
Do not override the artifact fetch timeout

stefanprodan:
Use the timeout set by the http client when retrying with exponential backoff, followup #289


phillebaba (approved):
LGTM

Issue #306, opened 2021-03-26
Support Age encryption format


hiddeco:
With SOPS v3.7.0 merged, we are now able to add support for the [Age encryption format](https://github.com/FiloSottile/age).

Rough sketch of what needs to be done:

- [ ] Refactor the [`KustomizeDecryptor`](https://github.com/fluxcd/kustomize-controller/blob/453b6e6e46c28fc4de5fef04c79617272116d155/controllers/kustomization_decryptor.go) so that only `.asc` key/value pairs from the referred secret are written to disk.
- [ ] Collect `.age` key/value pairs (in bytes) to be used for decryption using Age.
- [ ] Adapt our own `sops/keyservice` server so that it is capable of encrypting and decrypting the Age encryption format with the keys we collected from the secret. Inspiration for this can be drawn from from https://github.com/mozilla/sops/blob/66043e71a81787d6513bc2e5505a29aac67dc6f1/keyservice/server.go#L294-L301. As https://github.com/mozilla/sops/blob/master/age/keysource.go can simply be loaded with bytes, we likely won't have to do much more than constructing the keysource with the right values, and then use it to decrypt.

PR #309, opened 2021-03-31
Support SOPS age encryption

hiddeco:
Fixes: #306


controllers/kustomization_decryptor.go
```diff
@@ -136,18 +138,25 @@ func (kd *KustomizeDecryptor) ImportKeys(ctx context.Context) error {
...
+				if err != nil {
+					return err
+				}
+				if err := ioutil.WriteFile(keyPath, file, os.ModePerm); err != nil {
+					return fmt.Errorf("unable to write key to storage: %w", err)
+				}
+				if err := kd.gpgImport(keyPath); err != nil {
+					return err
+				}
+			case ".txt":
```
hiddeco, line 155: Wondering if this very generic extension is going to bite us later if some other encryption format is introduced, but do not have a clue about good alternatives, given `.age` is reserved for encrypted age files (and not keys).

stefanprodan: What about `.agekey`? the Age keygen doesn't have opinions about the file extension, but SOPS does... as `txt` is hoardcoded.

hiddeco: Think `.agekey` is an OK suggestion, SOPS guards itself against collisions by looking for a `keys.txt` file in a specific (age) folder.

stefanprodan (approved):
LGTM

Awesome work @hiddeco 🏅

PR #312, opened 2021-04-01
Detect and replace empty err output on apply

hiddeco:
This should give users some guidance when `kubectl apply` itself does
not give any useful output back itself, till date only observed when
it times out waiting.

Should help some with #311.


stefanprodan (approved):
LGTM

Thanks @hiddeco

PR #313, opened 2021-04-06
Move Kustomize patch tests to Go test suite

hiddeco:


stefanprodan (approved):
LGTM

Thanks @hiddeco ☕

PR #317, opened 2021-04-08
Make log level info for 'Dependencies do not meet ready condition'

allenporter:
Reduce the log level from error to info to match the level of the event.  The `kustomize-controller` pod's log has `error` message like:

`{"level":"error","ts":"2021-03-26T02:39:43.980Z","logger":"controller.kustomization","msg":"Dependencies do not meet ready condition, retrying in 30s","reconciler group":"kustomize.toolkit.fluxcd.io","reconciler kind":"Kustomization","name":"apps","namespace":"flux-system","error":"dependency 'flux-system/infrastructure' is not ready"}`

though the events / notifications written are at level `info`.

Context is discussion at https://github.com/fluxcd/flux2/discussions/1160#discussioncomment-580148


stefanprodan (approved):
LGTM

Thanks @allenporter

PR #322, opened 2021-04-14
Refactor indexers to use high-order functions

chanwit:
This PR refactored indexers by using high-order functions to replace redundant codes by half without altering the semantics.


stefanprodan (approved):
LGTM

Thanks @chanwit

PR #325, opened 2021-04-19
Make status compatible with kstatus

SomtochiAma:
This PR projects the `Ready` status condition to `Reconciling` and `Stalled` for better [kstatus](https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus) compliance.

### DIFF
--- a/api/v1beta1/kustomization_types.go
+++ b/api/v1beta1/kustomization_types.go
@@ -288,6 +288,8 @@ func RemoveKustomizationReconciling(k *Kustomization) {
 // KustomizationNotReady registers a failed apply attempt of the given Kustomization,
 // including a Snapshot.
 func KustomizationNotReadySnapshot(k Kustomization, snapshot *Snapshot, revision, reason, message string) Kustomization {
+	RemoveKustomizationReconciling(&k)
+	SetKustomizationStalled(&k, metav1.ConditionTrue, reason, message)
 	SetKustomizationReadiness(&k, metav1.ConditionFalse, reason, trimString(message, MaxConditionMessageLength), revision)
 	SetKustomizationHealthiness(&k, metav1.ConditionFalse, reason, reason)
 	k.Status.Snapshot = snapshot
--- a/api/v1beta1/kustomization_types.go
+++ b/api/v1beta1/kustomization_types.go
@@ -35,6 +35,8 @@ const (
 	KustomizationFinalizer    = "finalizers.fluxcd.io"
 	MaxConditionMessageLength = 20000
 	DisabledValue             = "disabled"
+	StalledCondition          = "Stalled"
+	ReconcilingCondition      = "Reconciling"
 )
 
 // KustomizationSpec defines the desired state of a kustomization.
@@ -230,7 +232,9 @@ type KustomizationStatus struct {
 // KustomizationProgressing resets the conditions of the given Kustomization to a single
 // ReadyCondition with status ConditionUnknown.
 func KustomizationProgressing(k Kustomization) Kustomization {
+	RemoveKustomizationStalled(&k)
 	meta.SetResourceCondition(&k, meta.ReadyCondition, metav1.ConditionUnknown, meta.ProgressingReason, "reconciliation in progress")
+	meta.SetResourceCondition(&k, ReconcilingCondition, metav1.ConditionTrue, meta.ProgressingReason, "reconciliation in progress")
 	return k
 }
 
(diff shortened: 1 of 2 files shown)


hiddeco (commented):
Based on [the documentation](https://github.com/kubernetes-sigs/cli-utils/blob/e351b2bc43cec2107ba1d874c3dec54fd0956c59/pkg/kstatus/README.md#conditions), the `Stalled` state must be complimented with a `Reconciling` condition (== what we call "Progressing" at the moment).

It looks like this technically also makes it possible to drop the `Ready` condition as it seems to be that `Ready == !Stalled && !Reconciling`. For backward compatibility reasons, and because of the limited adoption of `kstatus`, I would however still keep the condition.

controllers/kustomization_controller_status_test.go
```diff
@@ -0,0 +1,197 @@
...
+			defer k8sClient.Delete(context.TODO(), kustomization)
+
+			Eventually(func() bool {
+				var obj kustomizev1.Kustomization
+				_ = k8sClient.Get(context.Background(), ObjectKey(kustomization), &obj)
+				content, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(&obj)
+				unstructured := unstructured.Unstructured{}
+				unstructured.SetUnstructuredContent(content)
+				res, _ := status.Compute(&unstructured)
+				return apimeta.IsStatusConditionTrue(obj.Status.Conditions, StalledCondition) && (res.Status == status.FailedStatus)
```
stefanprodan, line 125: Please add here a check with kstatus

controllers/kustomization_controller_status_test.go
```diff
@@ -0,0 +1,197 @@
...
+				_ = k8sClient.Get(context.Background(), ObjectKey(kustomization), &obj)
+
+				content, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(&obj)
+				unstructured := unstructured.Unstructured{}
+				unstructured.SetUnstructuredContent(content)
+				res, _ := status.Compute(&unstructured)
+
+				isReady := apimeta.IsStatusConditionTrue(obj.Status.Conditions, meta.ReadyCondition)
+				stalledCondition := apimeta.FindStatusCondition(obj.Status.Conditions, StalledCondition)
+				return isReady && (res.Status == status.CurrentStatus) && (stalledCondition == nil)
```
stefanprodan, line 147: Please add here a check with kstatus

controllers/kustomization_controller_status_test.go
```diff
@@ -0,0 +1,197 @@
...
```
stefanprodan, line 193: Please add here a check with kstatus

hiddeco:
To add to my review comment:

* Add `Reconciling` condition to `fluxcd/pkg`, this seems to be the right thing to do as it follows the complete spec and the conditions are [the new standard](https://github.com/fluxcd/kustomize-controller/pull/325).
* Implement `Reconciling` condition here, it can be set when the resource is currently "progressing", and removed when it moves to `Ready` state.
* I would see if it is possible to scope the API around `Reconciling` and `Stalled`, and remove `NotReady`. This should save you a setter.

controllers/kustomization_controller_status_test.go
```diff
@@ -0,0 +1,216 @@
...
+	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/fluxcd/pkg/apis/meta"
+	"github.com/fluxcd/pkg/testserver"
+
+	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1beta1"
+)
+
+const StalledCondition = "Stalled"
```
stefanprodan, line 27: Replace this with fluxcd/apis/meta

api/v1beta1/kustomization_types.go
```diff
@@ -254,13 +256,35 @@ func SetKustomizationReadiness(k *Kustomization, status metav1.ConditionStatus,
...
+// RemoveKustomizationReconciling removes the ReconcilingCondition on the Kustomization.
+func RemoveKustomizationReconciling(k *Kustomization) {
+	if len(k.Status.Conditions) != 0 {
+		apimeta.RemoveStatusCondition(&k.Status.Conditions, meta.ReconcilingCondition)
+	}
+}
+
 // KustomizationNotReady registers a failed apply attempt of the given Kustomization,
 // including a Snapshot.
 func KustomizationNotReadySnapshot(k Kustomization, snapshot *Snapshot, revision, reason, message string) Kustomization {
```
hiddeco, line 290: Shouldn't this also include the logic to remove and/or add the `Reconciling` and `Stalled` conditions?

api/v1beta1/kustomization_types.go
```diff
@@ -254,13 +256,35 @@ func SetKustomizationReadiness(k *Kustomization, status metav1.ConditionStatus,
...
 	if revision != "" {
 		k.Status.LastAttemptedRevision = revision
 	}
 	return k
 }
 
+// SetKustomizeStalled sets the StalledCondition, ObservedGeneration on the Kustomization.
+func SetKustomizationStalled(k *Kustomization, status metav1.ConditionStatus, reason, message string) {
+	meta.SetResourceCondition(k, meta.StalledCondition, status, reason, trimString(message, MaxConditionMessageLength))
+	k.Status.ObservedGeneration = k.Generation
```
hiddeco, line 271: Should this be set on `Stalled`, or on reaching `Ready`? Or do we want to record it in the `ObservedGeneration` field of the `metav1.Condition` Status, but only move the one in the `Status` on a successful run?

api/v1beta1/kustomization_types.go
stefanprodan: Yes it should cc @SomtochiAma

api/v1beta1/kustomization_types.go
SomtochiAma: `kstatus` seems to return an `InProgress` status once `status.ObserveredGeneration` != `metadata.generation` regardless of the conditions present. So I thought we would want to set just incase,

hiddeco: Good catch! Which brings us to a new question: does every error mean that the reported condition status is `Stalled`, and `kstatus` should thus stop waiting because `ObservedGeneration == Generation` or `Status == Stalled`?

If we for example run into a transient error (e.g. transient resource lookup failure that returned an error from the K8s API server). This would mean that `kstatus` would stop waiting for the resource to reach the expected state if we would set both `ObservedGeneration` and `Stalled`, while if we would set `Ready==False` or left the `ObservedGeneration` untouched, it would continue to wait for the next reconciliation to (hopefully) succeed: https://github.com/kubernetes-sigs/cli-utils/blob/e351b2bc43cec2107ba1d874c3dec54fd0956c59/pkg/kstatus/status/generic.go#L39-L54

SomtochiAma: > If we for example run into a transient error (e.g. transient resource lookup failure that returned an error from the K8s API server).

Hmm. We probably have to identify transient errors and real errors, then decide how best to handlethe transient. I think we should set it `Reconciling` status for transient error and `Ready=false`.
 kstatus takes `ObservationGeneration != generated` to mean that the controller might not be running or unable to update the resource. 
 
 Plus a generic message is when `ObservationGeneration != generated`. If we use `Reconciling` condition, kstatus sets the reason and message from the message.

stefanprodan:
@hiddeco should we close this? I guess the kstatus conditions will be implemented in #347?

hiddeco:
This can be closed in favor of #347 indeed.

Issue #328, opened 2021-04-21
Decrypt base64 encoded SOPS encrypted secrets


bob-rohan:
Kustomize-controller can currently decrypt SOPS encrypted files - but whole files only. It does this by, among other things, string matching "sops" and "ENC".

Kubernetes Secrets are base64 encoded, therefore when a SOPS encrypted file is added as base64 encoded data to a Kubernetes Secret, it is not decrypted. 

I tried to explain this here  https://github.com/fluxcd/flux2/discussions/745#discussioncomment-601014

You may well ask yourself, why someone would want to base64 encode a SOPS encrypted file. We, and I believe others, would and have (using fluxv1 workaround not available in v2) been using Kustomize secretGenerator. It is a useful feature in that it appends a hash of the content of the secret, and propagates that suffixed name to pods that need that secret. This enables that if the underlying secret should change, appropriate deployment/pod rollouts are scheduled.

I will submit a very rough PR, showing a *functional* solution. This is my first foray into Go development, with such a complicated project I'm surprised it worked - credit to the CONTRIBUTING instructions on getting a local version running so easily. If there is appetite to accept the idea, then I and hopefully others can work to improve the quality of the code and tests etc..

bob-rohan:
https://github.com/fluxcd/kustomize-controller/pull/329

kingdonb:
Can you say more about this solution and what it looks like in practice?

Eg. from my current understanding, the secret presents as a regular secret when opened as YAML in an editor, but if you decode the base64 data, it is a SOPS-encrypted (original plaintext) file?

I appreciate that you've done some work and are presenting an implementation approach! This makes it much easier to talk about what it does and how it might need to change, in order to be accepted for merge.

Can you present some minimal usage examples too? (Ideally in the form of documentation, but it would help my understanding of what you've tried to do, just to see an example secret and explanation of how you're using it.)

stefanprodan:
@kingdonb I was also confused about this, but now that @bob-rohan has added tests to the PR, I can understand what it does. Thanks @bob-rohan for adding those tests 🏅

hodge-devops:
No worries, I appreciate the guidance. Do you have any recommended next steps with regards to getting the PR in a state to be merged?

PR #329, opened 2021-04-21
SOPS: Decrypt Kubernetes secrets generated by kustomize

bob-rohan:
Kustomize-controller can currently decrypt SOPS encrypted files - but
whole files only.

Kubernetes Secrets are base64 encoded, therefore when a SOPS encrypted
file, is added as base64 encoded data to a Kubernetes Secret, it is not
decrypted.

Fix: #328


stefanprodan (commented):
@bob-rohan can you please add a sub-section here https://github.com/fluxcd/kustomize-controller/blob/main/docs/spec/v1beta1/kustomization.md#secrets-decryption and document the new decryption option. Please mention kustomize secret generator and how to use it.

stefanprodan:
@bob-rohan can you please squash all commits into a single one and rename it to "Decrypt base64 encoded SOPS encrypted secrets". Thanks!

stefanprodan (approved):
LGTM

Thanks @bob-rohan 🏅

apeschel:
Thank you @bob-rohan, this is fantastic work!

PR #337, opened 2021-04-30
Update go.mod

rverma-nsl:
Bumping kustomize version


stefanprodan:
It's not that easy, the new API comes with many breaking changes, you need to refactor the whole controller.

PR #346, opened 2021-05-18
Fix validation and application timeout handling

ordovicia:
`err` returned from `command.CombinedOutput()` is an error of the command execution, not an error of the context.
To determine if the context was canceled, we should check out `context.Err()`.

Signed-off-by: Hidehito Yabuuchi <hdht.ybuc@gmail.com>


hiddeco (approved):
Good catch! :fishing_pole_and_fish:

PR #347, opened 2021-05-20
Implement helpers from pkg/runtime

SomtochiAma:


go.mod
```diff
@@ -39,3 +39,7 @@ require (
 	sigs.k8s.io/kustomize/api v0.7.4
 	sigs.k8s.io/yaml v1.2.0
 )
+
+replace (
```
stefanprodan, line 43: Please remove this, it will break the build.

controllers/kustomization_controller.go
```diff
@@ -123,278 +126,272 @@ func (r *KustomizationReconciler) SetupWithManager(mgr ctrl.Manager, opts Kustom
...
-		meta.ReconciliationSucceededReason,
-		"Applied revision: "+source.GetArtifact().Revision,
-	), nil
+		conditions.MarkFalse(kustomization, meta.ReadyCondition, kustomizev1.HealthCheckFailedReason, err.Error())
+		conditions.MarkFalse(kustomization, kustomizev1.HealthyCondition, kustomizev1.HealthCheckFailedReason, err.Error())
+		return ctrl.Result{RequeueAfter: kustomization.GetRetryInterval()}, err
+	}
+
+	kustomization.Status.LastAppliedRevision = artifact.Revision
+	kustomization.Status.Snapshot = snapshot
```
hiddeco, line 335: I would set this as soon as it can be set, and I think that might be a bit earlier.

@stefanprodan: at what time becomes the previous snapshot irrelevant?

stefanprodan: We have to set the snapshot right after prune.

controllers/kustomization_controller.go
```diff
@@ -123,278 +126,272 @@ func (r *KustomizationReconciler) SetupWithManager(mgr ctrl.Manager, opts Kustom
...
-		kustomization,
-		snapshot,
-		source.GetArtifact().Revision,
-		meta.ReconciliationSucceededReason,
-		"Applied revision: "+source.GetArtifact().Revision,
-	), nil
+		conditions.MarkFalse(kustomization, meta.ReadyCondition, kustomizev1.HealthCheckFailedReason, err.Error())
+		conditions.MarkFalse(kustomization, kustomizev1.HealthyCondition, kustomizev1.HealthCheckFailedReason, err.Error())
+		return ctrl.Result{RequeueAfter: kustomization.GetRetryInterval()}, err
+	}
```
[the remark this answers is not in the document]
stefanprodan: To preserve the current behaviour, the `ReadyCondition` should stay `Unknown` until the check finishes. The `HealthyCondition` could be set to `Unknown` with the message `Health check in progress` before the check starts.

controllers/kustomization_controller.go
```diff
@@ -127,279 +126,291 @@ func (r *KustomizationReconciler) SetupWithManager(mgr ctrl.Manager, opts Kustom
...
+		// this controller
+		patchOpts := []patch.Option{
+			patch.WithOwnedConditions{
+				Conditions: []string{
+					sourcev1.ArtifactAvailableCondition,
+					sourcev1.SourceAvailableCondition,
+					meta.ReadyCondition,
+					meta.ReconcilingCondition,
+					meta.StalledCondition,
+					meta.ProgressingReason,
```
stefanprodan, line 178: Why is here `ProgressingReason`? Do we need `HealtyCondition` also?

SomtochiAma: I think we need to set `HealthyCondition` too

stefanprodan:
Superseded by #575

Issue #305, opened 2021-03-25
Checksum label is missing when prune is false


stealthybox:
When prune is true, child objects of Kustomizations have these labels:
```
labels:
  kustomize.toolkit.fluxcd.io/checksum: 21ff4959f5c2052e702115c585ea8cf694c67613
  kustomize.toolkit.fluxcd.io/name: flux-system
  kustomize.toolkit.fluxcd.io/namespace: flux-system
```
When you set prune to false, the checksum label is dropped:
```
labels:
  kustomize.toolkit.fluxcd.io/name: flux-system
  kustomize.toolkit.fluxcd.io/namespace: flux-system
```

Even if the user does not have garbage collection enabled, it is still very useful to have these checksum labels because you can use them to identify stale resources.
This would for example be useful in a `flux tree` view or a web interface like the Flux UI that shows which objects are managed and which are not.

I don't see any reason not to include these checksums.
It should be a relatively safe change for existing Kustomizations when kustomize-controller is updated, but it will cause every object to be patched with new labels.

stefanprodan:
We used to have the checksum labels no matter if GC was enabled and that caused many issues for our users, controllers like Kafka, MySQL and others would delete the stateful sets every time the label changes, we also spam with alerts since all objects are changed.

stealthybox:
Fair enough, although that sounds like the behavior with prune is suboptimal in those cases.
Would a matching annotation do any harm?
Could we modify the GC to filter annotations and take a small performance hit?

stealthybox:
An issue with the above suggestion is that the GC gets slower linearly as garbage grows, but maybe it's worth the tradeoff.

Is the issue with the Kafka/MySQL controller that the labels are being copied down to the child STS objects from a parent CR?  (and the labels aren't stable)
I assume using annotations would fix that since they are designed to be a RW extension point for API's.

stefanprodan:
> Could we modify the GC to filter annotations and take a small performance hit?

I don’t think you can query based on annotations and fetching all objects of a kind could be expensive and due to rate limiting, on some distributions it will result in timeouts.

stealthybox:
If the label selector is only the kustomization Name + Namespace, we can still query more efficiently, but it would include all of the garbage.

simongottschlag:
Is the biggest problem here that some controllers doesn't handle updates of labels well?

[...]

PR #348, opened 2021-05-23
Add Checksum Annotation as Transformer to Kustomization

jonathan-innis:
To avoid always adding this to labels (and to not break current users), the proposed solution is to add an AnnotationTransformer to add the checksum as an annotation to track stale resources deployed by Kustomization


jonathan-innis:
@stealthybox This addresses the checksum issue

hiddeco:
Can you add a little information about the issue that is being addressed? As I can see _what_ this does, but can not directly see _why_ we want to do this.

jonathan-innis:
Sure, addresses #305

jonathan-innis:
Closing this because it will not be taken in its current state 😄

PR #353, opened 2021-05-31
Support decrypting any file format in secret generator

gmaiztegi:
This is an extension of #329, which added the support for decrypting files passed to the secret generator. The previous implementation only supported encrypted binary types, but failed when the content was a JSON or a YAML, with the message `EmitPlainFile: No binary data found in tree`.

This small change detects the format using the key/filename in the secret.


stefanprodan (approved):
LGTM

Thanks @gmaiztegi 🏅

Issue #315, opened 2021-04-07
Garbage collection conflicts with Stash controller


Legion2:
I'm using [Stash](https://stash.run/) to backup my kubernetes applications. Stash is a k8s operator. I manage the configuration of stash with a kustomization, which applies `BackupConfiguration`s to the cluster. Stash will create a service account for those `BackupConfiguration`s, therefore it copies the labels of the `BackupConfiguration`, which includes the `kustomize.toolkit.fluxcd.io/checksum` label.

Because the service account is not managed by flux and has the `kustomize.toolkit.fluxcd.io/checksum` label and pruning is activated, flux garbage collects the service account. As a result, the backups with stash do not work. This is probably a problem which must be fixed in stash, so I created an issue there https://github.com/stashed/stash/issues/1334.

However, the garbage collection behavior of flux is inconsistent. I have four `BackupConfiguration`s each with its own service account, but flux only deletes two of them. The two remaining service accounts have the `kustomize.toolkit.fluxcd.io/checksum` label, but are not garbage collected by flux.

kingdonb:
Maybe they are not in a namespace that was targeted by Kustomization with prune enabled?

It is hard to say what might be inconsistent about Flux's behavior, or if it is more likely a configuration surprise that we will understand only by reviewing details of your implementation...

https://toolkit.fluxcd.io/components/kustomize/kustomization/#garbage-collection

There's a workaround for you, in case you can't get Stash to stop doing that and you are still struggling with this until a fix is available, just disable garbage collection of your BackupConfigurations with an annotation or label. Let Stash go on copying the labels, it will copy this one to the service accounts too. (I guess also be sure to re-enable pruning first in the event you ever need to disable and delete an old BackupConfiguration.)

> You can disable pruning for certain resources by either labeling or annotating them with:

```yaml
  kustomize.toolkit.fluxcd.io/prune: disabled
```

stefanprodan:
>  I have four BackupConfigurations each with its own service account, but flux only deletes two of them. The two remaining service accounts have the kustomize.toolkit.fluxcd.io/checksum label, but are not garbage collected by flux.

It’s a race condition, by the time Flux gets to those SAs, Stash has copied the label value with the new checksum. I don’t see how we can fix any of this in kustomize-controller.

Legion2:
> Maybe they are not in a namespace that was targeted by Kustomization with prune enabled?

[...]

PR #362, opened 2021-06-09
Move the GC checksum from labels to annotations

JaneLiuL:
This PR moves the `kustomize.toolkit.fluxcd.io/checksum` from labels to annotations. This should fix any conflicts with 3rd party controllers like Stash (fix: #315) that are copying the labels from their custom resources to generated objects.

### DIFF
--- a/controllers/kustomization_gc.go
+++ b/controllers/kustomization_gc.go
@@ -137,9 +137,21 @@ func (kgc *KustomizeGarbageCollector) Prune(timeout time.Duration, name string, 
 	return changeSet, true
 }
 
+// Check both labels and annotations for the checksum to preserve backwards compatibility
 func (kgc *KustomizeGarbageCollector) isStale(obj unstructured.Unstructured) bool {
-	itemChecksum := obj.GetLabels()[fmt.Sprintf("%s/checksum", kustomizev1.GroupVersion.Group)]
-	return kgc.newChecksum == "" || itemChecksum != kgc.newChecksum
+	itemLabelChecksum := obj.GetLabels()[fmt.Sprintf("%s/checksum", kustomizev1.GroupVersion.Group)]
+	itemAnnotationChecksum := obj.GetAnnotations()[fmt.Sprintf("%s/checksum", kustomizev1.GroupVersion.Group)]
+
+	switch kgc.newChecksum {
+	case "":
+		return true
+	case itemAnnotationChecksum:
+		return false
+	case itemLabelChecksum:
+		return false
+	default:
+		return true
+	}
 }
 
 func (kgc *KustomizeGarbageCollector) shouldSkip(obj unstructured.Unstructured) bool {
@@ -156,7 +168,11 @@ func gcLabels(name, namespace, checksum string) map[string]string {
 	return map[string]string{
 		fmt.Sprintf("%s/name", kustomizev1.GroupVersion.Group):      name,
 		fmt.Sprintf("%s/namespace", kustomizev1.GroupVersion.Group): namespace,
-		fmt.Sprintf("%s/checksum", kustomizev1.GroupVersion.Group):  checksum,
+	}
+}
+func gcAnnotation(checksum string) map[string]string {
+	return map[string]string{
+		fmt.Sprintf("%s/checksum", kustomizev1.GroupVersion.Group): checksum,
 	}
 }
 
@@ -165,4 +181,5 @@ func selectorLabels(name, namespace string) map[string]string {
 		fmt.Sprintf("%s/name", kustomizev1.GroupVersion.Group):      name,
 		fmt.Sprintf("%s/namespace", kustomizev1.GroupVersion.Group): namespace,
 	}
+
 }
--- a/controllers/kustomization_generator.go
+++ b/controllers/kustomization_generator.go
@@ -36,13 +36,13 @@ import (
 	kustypes "sigs.k8s.io/kustomize/api/types"
 	"sigs.k8s.io/yaml"
 
-	"github.com/fluxcd/pkg/apis/kustomize"
-
 	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1beta1"
+	"github.com/fluxcd/pkg/apis/kustomize"
 )
 
 const (
-	transformerFileName = "kustomization-gc-labels.yaml"
+	transformerFileName           = "kustomization-gc-labels.yaml"
+	transformerAnnotationFileName = "kustomization-gc-annotations.yaml"
 )
 
 type KustomizeGenerator struct {
@@ -68,6 +68,9 @@ func (kg *KustomizeGenerator) WriteFile(ctx context.Context, dirPath string) (st
 	if err := kg.generateLabelTransformer(checksum, dirPath); err != nil {
 		return "", err
 	}
+	if err = kg.generateAnnotationTransformer(checksum, dirPath); err != nil {
+		return "", err
+	}
 
 	data, err := ioutil.ReadFile(kfile)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


controllers/utils.go
```diff
@@ -48,6 +48,53 @@ func parseApplyOutput(in []byte) map[string]string {
...
+// diff -u -N /tmp/LIVE-738467670/xx.v1alpha1.Keycloak.namespace.name /tmp/MERGED-747032253/xx.v1alpha1.Keycloak.namespace.name
+// --- /tmp/LIVE-738467670/xx.v1alpha1.Keycloak.namespace.name	2021-06-07 11:38:56.287331739 +0200
+// +++ /tmp/MERGED-747032253/xx.v1alpha1.Keycloak.namespace.name.test-mapping
+// @@ -8,7 +8,7 @@
+// ...
+// diff -u -N /tmp/LIVE-656588208/kustomize.toolkit.fluxcd.io.v1beta1.Kustomization.namespace.name /tmp/MERGED-532750671/kustomize.toolkit.fluxcd.io.v1beta1.Kustomization.namespace.name
+// --- /tmp/LIVE-656588208/kustomize.toolkit.fluxcd.io.v1beta1.Kustomization.namespace.name	2021-06-07 12:58:20.738794982 +0200
+// +++ /tmp/MERGED-532750671/kustomize.toolkit.fluxcd.io.v1beta1.Kustomization.namespace.name	2021-06-07 12:58:20.798795908 +0200
+// @@ -0,0 +1,36 @@
+func parasDiffOutput(in []byte) map[string]string {
```
stefanprodan, line 62:
```suggestion
func parseDiffOutput(in []byte) map[string]string {
```

controllers/kustomization_gc.go
```diff
@@ -138,7 +138,8 @@ func (kgc *KustomizeGarbageCollector) Prune(timeout time.Duration, name string,
 }
 
 func (kgc *KustomizeGarbageCollector) isStale(obj unstructured.Unstructured) bool {
-	itemChecksum := obj.GetLabels()[fmt.Sprintf("%s/checksum", kustomizev1.GroupVersion.Group)]
```
stefanprodan, line 141: We can't drop the label check in this version. If the label exists but the annotation doesn't, then we have to mark the object as stale otherwise after upgrading to this version the controller will fail to prune the objects reconciled with the older version. If both the label and annotation exists, then we should use the annotation.

stefanprodan:
@JaneLiuL I found some serious issues with `diff`, for example, if a user adds namespaces and other objects in those namespaces, then the diff fails while the apply works. I see no way around this limitation but to detect diff failures, ignore them and switch back to the old way of detecting changes.

controllers/utils.go
```diff
@@ -48,6 +48,50 @@ func parseApplyOutput(in []byte) map[string]string {
...
+// diff -u -N /tmp/LIVE-738467670/xx.v1alpha1.Keycloak.namespace.name /tmp/MERGED-747032253/xx.v1alpha1.Keycloak.namespace.name
+// --- /tmp/LIVE-738467670/xx.v1alpha1.Keycloak.namespace.name	2021-06-07 11:38:56.287331739 +0200
+// +++ /tmp/MERGED-747032253/xx.v1alpha1.Keycloak.namespace.name.test-mapping
+// @@ -8,7 +8,7 @@
+// ...
+// diff -u -N /tmp/LIVE-656588208/kustomize.toolkit.fluxcd.io.v1beta1.Kustomization.namespace.name /tmp/MERGED-532750671/kustomize.toolkit.fluxcd.io.v1beta1.Kustomization.namespace.name
+// --- /tmp/LIVE-656588208/kustomize.toolkit.fluxcd.io.v1beta1.Kustomization.namespace.name	2021-06-07 12:58:20.738794982 +0200
+// +++ /tmp/MERGED-532750671/kustomize.toolkit.fluxcd.io.v1beta1.Kustomization.namespace.name	2021-06-07 12:58:20.798795908 +0200
+// @@ -0,0 +1,36 @@
+func parseDiffOutput(in []byte) map[string]string {
```
stefanprodan, line 62: Please remove this as it's unrelated to this PR.

controllers/kustomization_generator.go
```diff
@@ -21,12 +21,10 @@ import (
 	"crypto/sha1"
 	"encoding/json"
 	"fmt"
+	"github.com/fluxcd/pkg/apis/kustomize"
```
stefanprodan, line 24: This needs to be added further down, the first import section is for Go std lib.

controllers/kustomization_generator.go
```diff
@@ -86,18 +88,19 @@ func (kg *KustomizeGenerator) WriteFile(ctx context.Context, dirPath string) (st
...
 	} else {
-		var exists bool
-		for _, transformer := range kus.Transformers {
-			if transformer == transformerFileName {
-				exists = true
-				break
-			}
-		}
-		if !exists {
+		//kus.Transformers = append(kus.Transformers, transformerFileName)
```
stefanprodan, line 93: Please remove these commented lines.

controllers/kustomization_controller.go
```diff
@@ -648,6 +649,9 @@ func (r *KustomizationReconciler) apply(ctx context.Context, kustomization kusto
 		}
 	}
 
+	changeSet := ""
```
stefanprodan, line 652: This change is unrelated to this PR please undo it.

controllers/kustomization_gc.go
```diff
@@ -137,9 +137,25 @@ func (kgc *KustomizeGarbageCollector) Prune(timeout time.Duration, name string,
 	return changeSet, true
 }
 
+// We can't drop the label check in this version.
```
stefanprodan, line 140: Please replace this with:

> Check both labels and annotations for the checksum to preserve backwards compatibility

stefanprodan (commented):
Please update the GC docs here https://github.com/fluxcd/kustomize-controller/blob/main/docs/spec/v1beta1/kustomization.md#garbage-collection

controllers/kustomization_generator.go
```diff
@@ -277,13 +278,49 @@ func (kg *KustomizeGenerator) checksum(ctx context.Context, dirPath string) (str
...
+		return err
+	}
+
+	return nil
+}
+
 func (kg *KustomizeGenerator) generateLabelTransformer(checksum, dirPath string) error {
 	labels := selectorLabels(kg.kustomization.GetName(), kg.kustomization.GetNamespace())
 
-	// add checksum label only if GC is enabled
```
stefanprodan, line 283: This has been lost, please restore it for the annotions

controllers/kustomization_generator.go
```diff
@@ -277,13 +278,49 @@ func (kg *KustomizeGenerator) checksum(ctx context.Context, dirPath string) (str
...
+			{Path: "metadata/annotations", CreateIfNotPresent: true},
+		},
+	}
+
+	data, err := yaml.Marshal(lt)
+	if err != nil {
+		return err
+	}
+
+	labelsFile := filepath.Join(dirPath, transformerAnnotationFileName)
```
stefanprodan, line 312:
```suggestion
	annotationsFile := filepath.Join(dirPath, transformerAnnotationFileName)
```

controllers/kustomization_generator.go
```diff
@@ -36,13 +36,13 @@ import (
...
-	"github.com/fluxcd/pkg/apis/kustomize"
-
 	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1beta1"
+	"github.com/fluxcd/pkg/apis/kustomize"
 )
 
 const (
-	transformerFileName = "kustomization-gc-labels.yaml"
+	transformerFileName           = "kustomization-gc-labels.yaml"
+	transformerAnnotationFileName = "kustomization-annotation-labels.yaml"
```
stefanprodan, line 45:
```suggestion
	transformerAnnotationFileName = "kustomization-gc-annotations.yaml"
```

JaneLiuL: Fix done~

controllers/kustomization_generator.go
JaneLiuL: Fix done~

controllers/kustomization_gc.go
JaneLiuL: Fix done~

stefanprodan (approved):
LGTM

Thanks @JaneLiuL 🌷 

PS. I'll merge this and do a followup PR to update the GC documentation.

PR #364, opened 2021-06-10
Add support for in-line generic patches to Flux Kustomization API

Kissy:
The `patchesJson6902` or `patchesStrategicMerge` fields in a Kustomize resource only allow patching single named resources. In order to patch multiple resources with a single patch, the generic `patches` field need to be used.

This PR implement the generic `patches` at the kustomize.toolkit.fluxcd.io/v1beta1/Kustomization level to unlock that feature.

One use case, would be for instance to override all path of all Kustomization at once without having to patch then one by one:
```yaml
kind: Kustomization
metadata:
  name: podinfo
  namespace: flux-system
spec:
  # ...omitted for brevity
  patches:
    - patch: |-
        - op: replace
          path: /spec/path
          value: ./deploy/lab/
      target:
        group: kustomize.toolkit.fluxcd.io
        version: v1beta1
        kind: Kustomization
```

The field `options` from the original Kustomization is **not implemented**
```
  options:
    allowNameChange: true
    allowKindChange: true
```

This PR can only be merge if [pkg PR #109](https://github.com/fluxcd/pkg/pull/109) is accepted.

### DIFF
--- a/docs/spec/v1beta1/kustomization.md
+++ b/docs/spec/v1beta1/kustomization.md
@@ -598,14 +598,15 @@ The Kustomization has a set of fields to extend and/or override the Kustomize
 patches and namespace on all the Kubernetes objects reconciled by the resource,
 offering support for the following Kustomize directives:
 
-- [namespace](https://kubectl.docs.kubernetes.io/references/kustomize/namespace/)
+- [namespace](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/namespace/)
+- [patches](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/patches/)
 - [patchesStrategicMerge](https://kubectl.docs.kubernetes.io/references/kustomize/patchesstrategicmerge/)
-- [patchesJson6902](https://kubectl.docs.kubernetes.io/references/kustomize/patchesjson6902/)
-- [images](https://kubectl.docs.kubernetes.io/references/kustomize/images/)
+- [patchesJson6902](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/patchesjson6902/)
+- [images](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/images/)
 
 ### Target namespace
 
-To configure the [Kustomize `namespace`](https://kubectl.docs.kubernetes.io/references/kustomize/namespace/)
+To configure the [Kustomize `namespace`](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/namespace/)
 and overwrite the namespace of all the Kubernetes objects reconciled by the `Kustomization`,
 `spec.targetNamespace` can be defined:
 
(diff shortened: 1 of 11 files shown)


Kissy:
Converted the PR to draft while waiting for [pkg PR #109](https://github.com/fluxcd/pkg/pull/109)

docs/spec/v1beta1/kustomization.md
```diff
@@ -622,9 +623,35 @@ spec:
 
 The `targetNamespace` is expected to exist.
 
+### Patches
+
+To add [Kustomize `patches` entries](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/patches/)
+to the configuration, and patch resources using either a [strategic merge](https://kubectl.docs.kubernetes.io/references/kustomize/glossary#patchstrategicmerge) 
+patch or a [JSON](https://kubectl.docs.kubernetes.io/references/kustomize/glossary#patchjson6902) patch,
+`spec.patches` items must contain a `target` selector and a `patch` document.
```
stefanprodan, line 631: `items must contain a target` but in the blow example there is no target specified

Kissy: Thank you for the catch :)

In the original Kustomize resource, it's possible to apply a strategic merge patch (not JSON6902) without specifying the target.
I decided to force the target to be present in the Flux implementation for simplicity. Maybe that is something could be added in the future.

stefanprodan:
> One use case, would be for instance to override all path of all Kustomization at once without having to patch then one by one.

Having more than one Kustomization that points at the same path is not something you should ever do, as the controller will delete and recreate everything. Why are you doing this?

stefanprodan:
@Kissy please update apis/kustomize to `v0.2.0` in both go.mod files and then run `make tidy`.

Kissy:
The use case I have in mind is:
each micro-services have their own git repository following that layout:
```
|- src & build files
|- deploy
    |- base
        |- Kustomization.yaml
        |- xxx.yaml
    |- test
        |- Kustomization.yaml
        |- patches.yaml
    |- prod 
        |- Kustmoization.yaml
        |- patches.yaml
```

Each git repository can be handled by a PR & permissions per teams with fine grain settings.
This layout follow the podinfo repository layout globally.

On the other side, the Infra team is managing a single repository with:
```
|- applications
    |- Kustomization.yaml
    |- app-1.yaml (1)
    |- app-2.yaml (2)
    |- ...
|- clusters
    |- test
        |- applications.yaml (3)
    |- prod
        |- applications.yaml (4)
```

(1) Would be a `kustomize.toolkit.fluxcd.io/v1beta1/Kustomization` source-ref to git repo of app-1
(2) Would be a `kustomize.toolkit.fluxcd.io/v1beta1/Kustomization` source-ref to git repo of app-2
(3) Would be creating a `kustomize.toolkit.fluxcd.io/v1beta1/Kustomization` with a patch targeting all `kustomize.toolkit.fluxcd.io/v1beta1/Kustomization` and overriding `path` to `./deploy/test`
(4) Would be creating a kustomize.toolkit.fluxcd.io/v1beta1/Kustomization with a patch targeting all `kustomize.toolkit.fluxcd.io/v1beta1/Kustomization` and overriding `path` to `./deploy/prod`

The Infra repo is clean and adding a new app / removing an app is done by adding or removing file in the `applications` folder.

hiddeco:
I think this patch could be almost identically applied to the helm-controller which also has support for JSON6902 and strategic merge patches at the moment.

stefanprodan:
@hiddeco I'm not sure it would work in all cases, as the patch can be on disk. While KC has the full source locally, HC doesn't. I don't expect people to place patches inside the chart, but for Git could work...

stefanprodan:
> each micro-services have their own git repository

Thanks @Kissy now I understand your use-case, the path is the same but the source is different so no conflicts 👌

stefanprodan:
@Kissy apis/kustomize `v0.2.0` contains the `Patch` field you've added to fluxcd/pkg.

Kissy:
Successfully ran a test with an external cluster with:
```
    - patch: |-
        - op: add
          path: /metadata/labels/app.kubernetes.io~1instance
          value: podinfo-abcxzy
      target:
        labelSelector: app=podinfo
```

stefanprodan (approved):
LGTM

Thanks @Kissy 🏅

PR #366, opened 2021-06-14
Release v0.13.0

stefanprodan:


CHANGELOG.md
```diff
@@ -2,6 +2,36 @@
...
+## 0.13.0
+
+**Release date:** 2021-06-14
+
+This prerelease brings the controller on a par with **Kustomize v4**.
+The Kubernetes and controller-runtime dependencies have been updated to match the Kubernetes 1.21 release.
+
+**Breaking change**: Due to the removal of `hashicorp/go-getter` from Kustomize v4,
+the set of URLs accepted by Kustomize in the `resources` filed is reduced to only
+file system paths or values compatible with git clone.
```
hiddeco, line 14:
```suggestion
file system paths or values compatible with `git clone`.
```

or

```suggestion
file system paths or values compatible with Git clone.
```

up to you :-)

---

If it also understands the scp-like Git URL syntax, I would opt for the former.

stefanprodan: It's actually `git`, kustomize shells out to the git CLI inside the kustomize-controller container.

hiddeco (approved):
LGTM, it took us awhile, but looks like we are now almost there :tada: :beers:

PR #376, opened 2021-06-25
Add test for prune toggling

ViBiOh:
Signed-off-by: Vincent Boutour <bob@vibioh.fr>

### DIFF
--- a/controllers/kustomization_controller_gc_test.go
+++ b/controllers/kustomization_controller_gc_test.go
@@ -334,5 +334,118 @@ spec:
 
 			Expect(k8sClient.Get(context.Background(), deploymentName, deployment)).To(Succeed())
 		})
+
+		It("reset the checksum annotation when prune is turned on", func() {
+			deploymentManifest := func(namespace string) string {
+				return fmt.Sprintf(`---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: test-deployment
+  namespace: %s
+spec:
+  selector:
+    matchLabels:
+      app: test-deployment
+  template:
+    metadata:
+      labels:
+        app: test-deployment
+    spec:
+      containers:
+      - name: test
+        image: podinfo
+`,
+					namespace)
+			}
+
+			manifests := []testserver.File{
+				{
+					Name: "deployment.yaml",
+					Body: deploymentManifest(namespace.Name),
+				},
+			}
+			artifact, err := artifactServer.ArtifactFromFiles(manifests)
+			Expect(err).NotTo(HaveOccurred())
+
+			url := fmt.Sprintf("%s/%s", artifactServer.URL(), artifact)
+
+			repositoryName := types.NamespacedName{
+				Name:      fmt.Sprintf("%s", randStringRunes(5)),
+				Namespace: namespace.Name,
+			}
+			repository := readyGitRepository(repositoryName, url, "v1", "")
+			Expect(k8sClient.Create(context.Background(), repository)).To(Succeed())
+			Expect(k8sClient.Status().Update(context.Background(), repository)).To(Succeed())
+			defer k8sClient.Delete(context.Background(), repository)
+
+			kName := types.NamespacedName{
+				Name:      fmt.Sprintf("%s", randStringRunes(5)),
+				Namespace: namespace.Name,
+			}
+			k := &kustomizev1.Kustomization{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      kName.Name,
+					Namespace: kName.Namespace,
+				},
+				Spec: kustomizev1.KustomizationSpec{
+					KubeConfig: kubeconfig,
+					Interval:   metav1.Duration{Duration: time.Hour},
+					Path:       "./",
+					Prune:      false,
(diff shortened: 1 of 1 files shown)


stefanprodan:
Please rebase with the upstream `main` branch to get the CI fixed. Thanks!

controllers/kustomization_controller_gc_test.go
```diff
@@ -334,5 +334,110 @@ spec:
...
+			}, timeout, time.Second).Should(BeTrue())
+			Expect(got.Status.LastAppliedRevision).To(Equal("v1"))
+
+			deployment := &appsv1.Deployment{}
+			deploymentName := types.NamespacedName{Name: "test-deployment", Namespace: namespace.Name}
+			Expect(k8sClient.Get(context.Background(), deploymentName, deployment)).To(Succeed())
+			Expect(deployment.Annotations[fmt.Sprintf("%s/checksum", kustomizev1.GroupVersion.Group)]).To(BeEmpty())
+
+			// Turn Kustomization pruning on
+			k.Spec.Prune = true
```
stefanprodan, line 422: This results in a an update conflict `Operation cannot be fulfilled on kustomizations.kustomize.toolkit.fluxcd.io \"wsoat\": the object has been modified`. We need to do a `k8sClient.Get` before updating the prune field.

ViBiOh: I just added it (as a fixup, so before merging, I'll need to rebase). Sorry for force-pushing, my fixup was not signed-off 😬

controllers/kustomization_controller_gc_test.go
```diff
@@ -334,5 +334,111 @@ spec:
...
+			Expect(k8sClient.Update(context.Background(), k)).To(Succeed())
+
+			Eventually(func() bool {
+				got := &kustomizev1.Kustomization{}
+				_ = k8sClient.Get(context.Background(), kName, got)
+				c := apimeta.FindStatusCondition(got.Status.Conditions, meta.ReadyCondition)
+				return c != nil && c.Reason == meta.ReconciliationSucceededReason
+			}, timeout, time.Second).Should(BeTrue())
+			Expect(deployment.Annotations[fmt.Sprintf("%s/checksum", kustomizev1.GroupVersion.Group)]).NotTo(BeEmpty())
```
stefanprodan, line 432: Before this, we need to re-read the Deployment with a `k8sClient.Get`

ViBiOh: Indeed, I checked the following instructions, we must be good now 😌

controllers/kustomization_controller_gc_test.go
```diff
@@ -334,5 +334,114 @@ spec:
...
+					Suspend:    false,
+					Timeout:    &metav1.Duration{Duration: 60 * time.Second},
+					Validation: "client",
+				},
+			}
+			Expect(k8sClient.Create(context.Background(), k)).To(Succeed())
+			defer k8sClient.Delete(context.Background(), k)
+
+			got := &kustomizev1.Kustomization{}
+			Eventually(func() bool {
```
stefanprodan, line 409: This will return instantly with the previous reconciliation result. To fix this, we need to bump the source revision and wait for it to reconcile like so:

```go
repository.Status.Artifact.Revision = "v2"
Expect(k8sClient.Status().Update(context.Background(), repository)).To(Succeed())

// Turn Kustomization pruning on
_ = k8sClient.Get(context.Background(), kName, k)
k.Spec.Prune = true
Expect(k8sClient.Update(context.Background(), k)).To(Succeed())

updated := &kustomizev1.Kustomization{}
Eventually(func() bool {
	_ = k8sClient.Get(context.Background(), kName, updated)
	return updated.Status.LastAppliedRevision == "v2"
}, timeout, time.Second).Should(BeTrue())

Expect(updated.Spec.Prune).To(BeTrue())
Expect(updated.Status.LastAppliedRevision).To(Equal("v2"))
```

ViBiOh: Whoop, it's all green \o/

ViBiOh:
I've rebased upstream and cleaned up my fixup commits 👍

stefanprodan:
So @ViBiOh this test proves that the controller does add the checksum annotations to all objects after you set `prune: true` when the source has a new revision.

ViBiOh:
> So @ViBiOh this test proves that the controller does add the checksum annotations to all objects after you set `prune: true` when the source has a new revision.

Oh I thought you've made change when asked me to rebase.

I've made a deep look inside the resources that was not deleted, the label on it was pointing the wrong Kustomization. I think file has been deleted while I was switching kustomization:
* turning `prune: false` on my global Kustomization -> remove of checksum annotation
* deleting the kustomization and creating one kustomization per namespace with `prune: true` (in the same commit, but it takes some time to reconcile). In the interval, file describing my ressources has been deleted -> the annotations are unchanged and still pointing old kustomization and checksum is not added -> the resources remains because nothing owns it.

I close the related issue, do you want to keep the unit test or I close it too?

Thank you for you time and patience 😌

stefanprodan (approved):
LGTM

Thanks @ViBiOh for contributing this test 🏅

PR #377, opened 2021-06-28
Test kustomization depend on for improve ut

JaneLiuL:
Add UT test for kustomization controller, Steps as:
1.  Create kustomization with depend on
2.  Kustomization Obj status should equal with DependencyNotReadyReason
3.  Get resource fail for kustomization not ready

### DIFF
--- a/controllers/kustomization_controller_test.go
+++ b/controllers/kustomization_controller_test.go
@@ -396,7 +396,7 @@ spec:
 						Interval:   metav1.Duration{Duration: reconciliationInterval},
 						Path:       "./",
 						Prune:      true,
-						DependsOn:  []dependency.CrossNamespaceDependencyReference{
+						DependsOn: []dependency.CrossNamespaceDependencyReference{
 							{
 								Name: "test-kustomization",
 							},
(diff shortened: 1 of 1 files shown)


squaremo:
I suspect the failure has to do with this line in the test log: https://github.com/fluxcd/kustomize-controller/pull/377/checks?check_run_id=2929345023#step:10:55, specifically this part of the error message:

> Error: invalid argument \"client\" for \"--dry-run\"

I don't know the tests inside out, but to me that suggests that it's shelling out to `kubectl` and expecting a version which supports --dry-run=client, but getting an older version which only expects --dry-run as a boolean.

stefanprodan:
Please rebase with the upstream `main` branch to get the CI fixed. Thanks!

controllers/kustomization_controller_test.go
```diff
@@ -359,6 +360,74 @@ spec:
...
+					for _, c := range got.Status.Conditions {
+						if c.Reason == meta.DependencyNotReadyReason {
+							return true
+						}
+					}
+					return false
+				}, timeout, interval).Should(BeTrue())
+
+				deployment := &appsv1.Deployment{}
+				Expect(k8sClient.Get(context.Background(), types.NamespacedName{Name: "test", Namespace: namespace.Name}, deployment)).NotTo(Succeed())
```
stefanprodan, line 429: I don't think we need this, can you remove it please.

stefanprodan (approved):
LGTM

Thanks @JaneLiuL 🌹

PR #378, opened 2021-06-28
Fix CI: Override kubebuilder kubectl

stefanprodan:


squaremo (approved):
Looks like it'll work :-) It may prove brittle if workflow steps change order, or if the location kubebuilder puts the binary changes; but we may not care.

 - could there be a version check to guard against regressions?
 - would changing the PATH work (and be more future proof)?

stefanprodan:
@squaremo this is just a temporary fix, we'll get rid of kubebuilder from CI and replace it the new envtest composite action that I made [here](https://github.com/fluxcd/pkg/blob/main/actions/envtest/action.yml) and it will no longer conflict.

Issue #352, opened 2021-05-26
Detect changes/drift through a filtered diff of the apply/prune


stealthybox:
Instead of relying on kubectl apply output to send Object Update Events, we should:
- run kubectl diff first
- parse the output
- remove all objects where only the checksum annotation changed
- use this list for Object Update Events

> Once we do that, we can drop the checksum label, and use the annotation for GC as well. This would be a major improvement for those that have GC enabled, as they will receive events only for the objects where there is an actual drift.
> 
> https://github.com/fluxcd/kustomize-controller/issues/305#issuecomment-848480607

Some behavior that should clarified:
- **Intentional Update**:
  Do we also emit a "Checksum Changed" Event?
  If so, how are Object Prunes included in a change?
- **Drift Correction**:
  Should we generate an Event when the checksum does not change but there is a diff?

There are potentially some performance characteristics in the suggested implementation we should consider:
> Hopefully cache-hits in the apply for all of the objects from the diff will make adding the diff unnoticeable.
> 
> It would be really ideal if the kubectl output had structured diff information already.
I wonder if using kubectl as a library would make this more possible.
> 
> https://github.com/fluxcd/kustomize-controller/issues/305#issuecomment-848516852

JaneLiuL:
This is very good idea actually~
Only one think I am very confuse is :
If the first commit, we generate 4 resource, and apply to cluster by the kustomize-controller.
Then we commit, and remove 3 of them, only 1 resource on the git repository.

Now when we use `kustomize build > output.yaml`, and run `# kubectl  diff -f output.yaml`, it seems there will no any tips for the 3 resource which have been delete from git repository.

So for the  "Object Prunes "，if we don't use the `checksum` label,  could it possible to use the `diff first-output.yaml` second-output.yaml ? If yes, then we need the source-controller to artifact the last output and latest output.yaml.

Please correct me if I am wrong :) .

stefanprodan:
@JaneLiuL the controller knows which objects with get deleted, we need to implement our own GC dry run and append the diff to the apply one.

PR #379, opened 2021-06-29
Detect changes/drift through a filtered diff of the apply

JaneLiuL:
This PR including below contents:
1. run kubectl diff first and parse the output. .
2. update the object events.
3. test for the output

Detail could be check on: https://github.com/fluxcd/kustomize-controller/issues/352

### DIFF
--- a/controllers/kustomization_controller.go
+++ b/controllers/kustomization_controller.go
@@ -624,11 +624,15 @@ func (r *KustomizationReconciler) apply(ctx context.Context, kustomization kusto
 	start := time.Now()
 	timeout := kustomization.GetTimeout() + (time.Second * 1)
 	applyCtx, cancel := context.WithTimeout(ctx, timeout)
+	var resources map[string]string
 	defer cancel()
 	fieldManager := "kustomize-controller"
 
 	cmd := fmt.Sprintf("cd %s && kubectl apply --field-manager=%s -f %s.yaml --timeout=%s --cache-dir=/tmp --force=%t",
 		dirPath, fieldManager, kustomization.GetUID(), kustomization.Spec.Interval.Duration.String(), kustomization.Spec.Force)
+	// kubectl diff command to get the changeset
+	diffcmd := fmt.Sprintf("cd %s && kubectl diff --field-manager=%s -f %s.yaml  --cache-dir=/tmp",
+		dirPath, fieldManager, kustomization.GetUID())
 
 	if kustomization.Spec.KubeConfig != nil {
 		kubeConfig, err := imp.WriteKubeConfig(ctx)
@@ -636,6 +640,7 @@ func (r *KustomizationReconciler) apply(ctx context.Context, kustomization kusto
 			return "", err
 		}
 		cmd = fmt.Sprintf("%s --kubeconfig=%s", cmd, kubeConfig)
+		diffcmd = fmt.Sprintf("%s --kubeconfig %s", diffcmd, kubeConfig)
 	} else {
 		// impersonate SA
 		if kustomization.Spec.ServiceAccountName != "" {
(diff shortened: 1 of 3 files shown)


stefanprodan:
We need a test that proves that we ignore objects that are changed only due to the checksum annotation. We also need to test diff failures, for example a namespace and configmap will error out on create, but will succeed on update.

JaneLiuL:
Get ~ I will add it now~

hiddeco:
I think that as a future optimization it would also be an idea to create some type of wrapper around the working directory and/or commands we run so that we can omit the `cd ... &&` (or make them simple methods that run specific commands, or something something).

controllers/utils.go
```diff
@@ -87,3 +154,15 @@ func ObjectKey(object metav1.Object) client.ObjectKey {
 		Name:      object.GetName(),
 	}
 }
+
+func containsChangeInSlice(tmpslice []string) bool {
+	for _, s := range tmpslice {
+		if strings.HasPrefix(s, "+    kustomize.toolkit.fluxcd.io/checksum:") || strings.HasPrefix(s, "-    kustomize.toolkit.fluxcd.io/checksum:") || strings.HasPrefix(s, "--- /tmp") || strings.HasPrefix(s, "+++ /tmp") {
```
stefanprodan, line 160: Please declare a var above like `checksumAnnotation := fmt.Sprintf("%s/checksum", kustomizev1.GroupVersion.Group)`

stefanprodan:
Some things to note about this PR:
* if the diff returns no changeset, then the controller skips the apply
* the diff excludes changes to the checksum annotation trying to reduce the event spam but it fails to do so for Kubernetes Deployments, Statefulsets and Daemonsets due to the fact that an annotation change bumps the object generation 
* the diff errors out with exit code 2 if newly created objects are part of a new namespace
* we ignore the diff crash and fallback to the apply output to compose the changeset
* the apply output is very different to the diff one, users will receive inconsistent events, when diff fails the changeset format is `<object-name> <action>`, if diff works then the format is `<api-version>.<kind>.<namespace>.<name> <action>`
* we have no idea about the performance impact of this change, in theory this would double the number of API calls made to the API server

controllers/kustomization_controller.go
```diff
@@ -645,28 +650,41 @@ func (r *KustomizationReconciler) apply(ctx context.Context, kustomization kusto
...
-
-		applyErr := parseApplyError(output)
-		if applyErr == "" {
-			applyErr = "no error output found, this may happen because of a timeout"
-		}
-		return "", fmt.Errorf("apply failed: %s", applyErr)
-	}
-
-	resources := parseApplyOutput(output)
 	log.Info(
```
stefanprodan, line 688: We shouldn't log that we applied something on the cluster if we didn't, please move the log below inside `else if differr.ExitCode() == 1 `

controllers/kustomization_controller.go
```diff
@@ -645,28 +650,41 @@ func (r *KustomizationReconciler) apply(ctx context.Context, kustomization kusto
...
+	// For the "diff" command Exit status:
+	//		 0
+	//		No differences were found.
+	//		 1
+	//		Differences were found.
+	//		 >1
+	//		Kubectl or diff failed with an error.
+	if differr, ok := err.(*exec.ExitError); ok {
+		if differr.ExitCode() > 1 {
+			log.Info("Kubectl or diff failed with an error, will execute apply soon")
```
stefanprodan, line 671: Make this a debug log please.

controllers/kustomization_controller.go
```diff
@@ -645,28 +650,41 @@ func (r *KustomizationReconciler) apply(ctx context.Context, kustomization kusto
...
+			output, applyErr := execApply(ctx, cmd)
+			if applyErr != nil {
+				return "", applyErr
+			}
+			resources = parseApplyOutput(output)
+		} else if differr.ExitCode() == 1 {
+			resources = parseDiffOutput(diffoutput)
+			//	Since we found difference in "diff" command, so we need to execute apply
+			log.Info("Differences found, will execute apply soon")
```
stefanprodan, line 680: Make this a debug log please.

stefanprodan:
One unfortunate outcome of skipping the apply when the diff returns no changes is that enabling pruning will not be possible, as the the annotations will never be added.

controllers/kustomization_controller.go
JaneLiuL: Fix done

controllers/kustomization_controller.go
JaneLiuL: Fix done~

controllers/kustomization_controller.go
JaneLiuL: Fix done

controllers/utils.go
JaneLiuL: Fix done

JaneLiuL:
Any update for this PR? Should I still keep this PR or not? @stefanprodan

hiddeco:
:wave: 

We have not forgotten about this PR, but are currently in the process of a bigger refactor which will drastically change the structure of all controller's reconcilers (see https://github.com/fluxcd/flux2/issues/1601 for details), and improve the test coverage of the individual operations a controller performs.

Given one of the goals of this PRs is to hypothetically improve performance, we think it is better for the refactor to land first, so that we are properly able to benchmark your proposal against the current way the controller does things.

Well aware this may be a bummer to you, and I am really sorry about that. :sunflower:

JaneLiuL:
Close this PR since already have https://github.com/stefanprodan/kustomizer/pull/17 to cove

PR #382, opened 2021-06-30
Release v0.13.1

stefanprodan:


Makefile
```diff
@@ -2,7 +2,7 @@
 IMG ?= fluxcd/kustomize-controller:latest
 # Produce CRDs that work back to Kubernetes 1.16
 CRD_OPTIONS ?= crd:crdVersions=v1
-SOURCE_VER ?= v0.14.0
+SOURCE_VER ?= v0.15.3
```
hiddeco, line 5: Not related to this release, but wondering why this is a requirement while I was able to work without it in the helm-controller. If we are able to get rid of it, we could automate the updates of the source-controller dependency in GitHub actions more easily so that we no longer forget to update it, as that seems to have become an incidental pattern.

hiddeco (approved):
:flying_saucer:

Makefile
stefanprodan: It's a requirement to run the unit tests against the right Source API version. We could automate the version bump here and the one in `default/kustomization.yaml` for e2e testing.

PR #389, opened 2021-07-07
Var substitution opt-in docs

### LINKED ISSUE #388: Run ENV Substitution if empty substituted
Hi

I was playing around with variable substitutions. I made heavy use of defaulting mechanisms. Till I noticed, that defaults are not evaluated, when the owning Kustomization does not have any variables substituted. In fact no expression is evaluated. Looking at the code, this condition causes this: https://github.com/fluxcd/kustomize-controller/blob/80ebb792241accf107488603dd747123559d328b/controllers/kustomization_varsub.go#L75

I am not quiet sure if this is on purpose but the expected behavior for me would be if i have a default expression, it should default  if the variable isn't set.  But before trying to change the code I wanted to check with you guys, what you think about it.

oliverbaehler:
Signed-off-by: Oliver Bähler <oliverbaehler@hotmail.com>

Documents that var substitution must have set at least one var the be executed. Fixes #388

### DIFF
--- a/docs/spec/v1beta1/kustomization.md
+++ b/docs/spec/v1beta1/kustomization.md
@@ -803,7 +803,10 @@ labeling or annotating them with:
 kustomize.toolkit.fluxcd.io/substitute: disabled
 ``` 
 
-Substitution needs to be defined opt-in. If not at least a single variable or resource to substitute from is defined, no var substitution is performed. So even expressions which should evaluate to a default value won't be evaluated. To enable var substitution, just set at least one variable which will enable var substitution for the Kustomization, if you don't have any other variables on the Kustomization. For example:
+Substitution of variables only happens if at least a single variable or resource to substitute
+from is defined. This may cause issues if you rely on expressions which should evaluate to a
+default, even if no other variables are configured. To work around this, one can set an
+arbitrary key/value pair to enable the substitution of variables. For example: 
 
 ```
 apiVersion: kustomize.toolkit.fluxcd.io/v1beta1
@@ -814,7 +817,7 @@ spec:
   ...
   postBuild:
     substitute:
-      enabled: "true"
+      var_substitution_enabled: "true"
 ```
 
 You can replicate the controller post-build substitutions locally using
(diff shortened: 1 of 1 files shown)


docs/spec/v1beta1/kustomization.md
```diff
@@ -803,6 +803,20 @@ labeling or annotating them with:
...
+```
+apiVersion: kustomize.toolkit.fluxcd.io/v1beta1
+kind: Kustomization
+metadata:
+  name: apps
+spec:
+  ...
+  postBuild:
+    substitute:
+      enabled: "true"
```
hiddeco, line 817: Given this can be _any_ value, I would use something more explicitly unique, to not accidentally set something we do not want to set and overwrite people their value.

oliverbaehler: Something like that?

docs/spec/v1beta1/kustomization.md
```diff
@@ -803,6 +803,20 @@ labeling or annotating them with:
 kustomize.toolkit.fluxcd.io/substitute: disabled
 ``` 
 
+Substitution needs to be defined opt-in. If not at least a single variable or resource to substitute from is defined, no var substitution is performed. So even expressions which should evaluate to a default value won't be evaluated. To enable var substitution, just set at least one variable which will enable var substitution for the Kustomization, if you don't have any other variables on the Kustomization. For example:
```
hiddeco, line 806:
```suggestion
Substitution of variables only happens if at least a single variable or resource to substitute
from is defined. This may cause issues if you rely on expressions which should evaluate to a
default, even if no other variables are configured. To work around this, one can set an
arbitrary key/value pair to enable the substitution of variables. For example: 
```

Rephrased it a bit, as this suggestion is easier to read for myself, but I am open to counter suggestions :-)

docs/spec/v1beta1/kustomization.md
hiddeco: Perfect!

docs/spec/v1beta1/kustomization.md
oliverbaehler: Seems good to me

hiddeco (approved):
LGTM, thank you @oliverbaehler :1st_place_medal:

PR #374, opened 2021-06-24
feat: make it easier to reason about health check failures

makkes:
Whenever a health check times out now, the most recently collected error for each resource will be printed as part of the error message. This excludes errors for those resources for which no error was reported in the last update. This is because whenever a timeout occurs, an error is reported on ALL resources, even those that have been seen as healthy before.

Also, this commit causes all successfully checked resources to be omitted in the error event.

This is how it will look in the API:

Before:
```
$ k -n flux-system get events --field-selector=involvedObject.kind=Kustomization -w
LAST SEEN   TYPE     REASON   OBJECT                MESSAGE
34m         Normal   info     kustomization/test1   configmap/test1 created
32m         Normal   info     kustomization/test1   Update completed
0s          Normal   error    kustomization/test1   Health check timed out for [Deployment 'default/fails', Deployment 'default/fails2', Job 'default/wrong', Deployment 'default/git']

```

After:
```
$ k -n flux-system get events --field-selector=involvedObject.kind=Kustomization -w
LAST SEEN   TYPE     REASON   OBJECT                MESSAGE
34m         Normal   info     kustomization/test1   configmap/test1 created
32m         Normal   info     kustomization/test1   Update completed
0s          Normal   error    kustomization/test1   Health check failed for [Deployment 'default/fails' (status 'InProgress'): context deadline exceeded, Deployment 'default/fails2' (status 'NotFound'): context deadline exceeded, Job 'default/wrong' (status 'Unknown'): no matches for kind "Job" in group ""]
```

Signed-off-by: Max Jonas Werner <mail@makk.es>

PR #394, opened 2021-07-26
fix: prevent nil pointer dereference in health checks

makkes:
The `lastStatus` map now stores `event.ResourceStatus` values instead
of pointers to them, preventing nil pointers being dereferenced by
accident.

refs https://github.com/fluxcd/kustomize-controller/pull/374

Signed-off-by: Max Jonas Werner <mail@makk.es>

### DIFF
--- a/controllers/kustomization_healthcheck.go
+++ b/controllers/kustomization_healthcheck.go
@@ -64,6 +64,9 @@ func (hc *KustomizeHealthCheck) Assess(pollInterval time.Duration) error {
 		func(statusCollector *collector.ResourceStatusCollector, e event.Event) {
 			var rss []*event.ResourceStatus
 			for _, rs := range statusCollector.ResourceStatuses {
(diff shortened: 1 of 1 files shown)


makkes:
@stefanprodan as promised.

hiddeco:
Wouldn't this result in L92-100 tripping because it now deals with an initialized but empty object, and wouldn't it thus be better to keep the pointer list, but deal with `nil` values there?

(In general I think, pointer object lists often use pointers because `nil` has a special meaning, and you want to maintain this information for as long as you may need it).

hiddeco (approved):
This addresses all my (personal) previous concerns, thank you @makkes :100:

hiddeco:
DCO signoff (and while you are at it, a smash of) commits is still required.

stefanprodan (approved):
LGTM 

Thanks @makkes

Issue #399, opened 2021-08-03
SOPS encrypted secretGenerator fails to decrypt yaml if key name does not include a file extension


chadswen:
https://github.com/fluxcd/kustomize-controller/blob/81a40903287df4b2f77e0f1b2d04c0d527ac29e2/controllers/kustomization_decryptor.go#L160

The kustomization decryptor sets the storage format based on the file extension of a path, but it uses the key name of the generated Secret as the path, which may not always match the file name and storage format of the encrypted file.

The change to determine file format with the key name was introduced in #353 to resolve similar `EmitPlainFile: No binary data found in tree` errors, but did not address what happens if key names can't include a file extension.

I ran into this when trying to use a secretGenerator for `kubeconfig` files stored encrypted in git. My goal is to create a consistent UX for remote clusters with kubeconfig Secrets that are not managed by CAPI.

Example:
```yaml
# This fails to decrypt because the key name `value` will always be mapped to the binary format when it's actually yaml
secretGenerator:
- name: kubeconfig
  files:
  - value=kubeconfig.enc.yaml
```

```yaml
# This decrypts, but can't be used to configure remote clusters because the Secret's key must be named `value`
secretGenerator:
- name: kubeconfig
  files:
  - value.yaml=kubeconfig.enc.yaml
```

To workaround this, I either have to:
1. encrypt the yaml file as binary, because the call to `common.DefaultStoreForPath(key)` always falls through to binary when the key name is `value`.
1. opt out of using `secretGenerator` at all and precreate/encrypt a secret containing each kubeconfig manually.

Would it be possible to use the filename specified in the kustomization's secretGenerator (e.g. `kubeconfig.enc.yaml` in the example) when fetching the Store format instead of the generated Secret's key name? That would allow the kustomization decryptor to at least match the expected behavior of `sops` decrypt on the same file (e.g. `sops --decrypt kubeconfig.enc.yaml`).

Alternatively there is also a [DefaultStoreForPathOrFormat](https://pkg.go.dev/go.mozilla.org/sops/v3/cmd/sops/common#DefaultStoreForPathOrFormat) function which takes a format string as input and falls back to the filename, but that requires some way of storing and retrieving the intended format string.

PR #400, opened 2021-08-04
Make the kubeconfig secrets compatible with SOPS

stefanprodan:
Add `values.yaml` to the supported kubeconfig secret key names in order for SOPS to correctly detect the storage format based on the file extension.

Fix: #399 

~~TODO: This must be ported to helm-controller before the next flux release.~~

### DIFF
--- a/controllers/kustomization_impersonation.go
+++ b/controllers/kustomization_impersonation.go
@@ -195,8 +195,15 @@ func (ki *KustomizeImpersonation) getKubeConfig(ctx context.Context) ([]byte, er
 		return nil, fmt.Errorf("unable to read KubeConfig secret '%s' error: %w", secretName.String(), err)
 	}
 
-	kubeConfig, ok := secret.Data["value"]
-	if !ok {
+	var kubeConfig []byte
+	for k, _ := range secret.Data {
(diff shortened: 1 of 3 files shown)


controllers/kustomization_impersonation.go
```diff
@@ -195,8 +195,15 @@ func (ki *KustomizeImpersonation) getKubeConfig(ctx context.Context) ([]byte, er
 		return nil, fmt.Errorf("unable to read KubeConfig secret '%s' error: %w", secretName.String(), err)
 	}
 
-	kubeConfig, ok := secret.Data["value"]
-	if !ok {
+	var kubeConfig []byte
+	for k, _ := range secret.Data {
+		if k == "value" || k == "value.yaml" {
```
hiddeco, line 200: Given these are parts of a contract (they are defined explicitly in the spec), I would make them more "official" part of the programmatic API by e.g. making them constants.

stefanprodan: This must move to `fluxcd/pkg/runtime/client` so I would leave it here as it is since this function will no longer be in each controller.

hiddeco (approved):

PR #404, opened 2021-08-08
feat: set default observedGeneration to -1 on Kustomizations

makkes:
This is a follow-up to
https://github.com/fluxcd/helm-controller/pull/294, porting the same
code to the kustomize-controller so that all Flux 2 controllers work
the same way in this regard.

Signed-off-by: Max Jonas Werner <mail@makk.es>


config/testdata/status-defaults/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kustomize.toolkit.fluxcd.io/v1beta1
```
stefanprodan, line 1: A Flux Kustomization file name should never be kustomization.yaml, as this file name is reserved for Kustomize configs. How come this manifests can be applied on the cluster? It has no spec and it should fail validation: https://github.com/fluxcd/kustomize-controller/blob/main/config/crd/bases/kustomize.toolkit.fluxcd.io_kustomizations.yaml#L327

stefanprodan (changes requested):

PR #408, opened 2021-08-13
Add tests for various kustomize transformers.

SomtochiAma:
This pull request adds test for most kustomize transformers when added inline and in a separate file.
This ensures Kustomize compatibility.


stefanprodan (approved):
LGTM

Thanks @SomtochiAma 🌹

PR #411, opened 2021-08-18
Skip garbage collection of objects with owner references

stefanprodan:
Looks like #362 didn't do much, various Kubernetes controllers copy the metadata (labels and annotations) from their custom resources to generated objects. To fix this once and for all, kustomize-controller will skip any object that has [ownerReference.BlockOwnerDeletion=true](https://v1-18.docs.kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#controlling-how-the-garbage-collector-deletes-dependents).

Fix: #410


phillebaba:
I think that it would be good to add a small test for this if possible just because the impact seems to be very large on end users if this is ever removed in the future by accident. Other than that this looks good.

stefanprodan:
@phillebaba I've added tests for when BlockOwnerDeletion is enabled or not.

phillebaba (approved):
LGTM

PR #425, opened 2021-09-10
Refactor tests

stefanprodan:
Changes:
- Install envtest binaries in testbin to avoid conflicting with setup-envtest on dev machines
- Drop kubebuilder from CI
- Drop Ginko and refactor tests
- Upgrade controller-runtime to v0.10.0


controllers/suite_test.go
```diff
@@ -140,3 +146,163 @@ func randStringRunes(n int) string {
 	}
 	return string(b)
 }
+
+func createNamespace(name string) error {
```
hiddeco, line 150: Would maybe leave a comment here to remind ourselves that once we switch over to the wrapper from `fluxcd/pkg`, this can be switched to the method the wrapper has to offer. As we tend to forget about these little details.

stefanprodan: Comment added!

hiddeco (approved):

Issue #213, opened 2020-12-16
Dry-run support


raffis:
A dry-run feature would be quite useful to have on the ks pipeline. If you deal with big sources a dry-run is something useful (Also if one migrates a source with existing manifests).
Basically if dry-run is enabled just stop before apply or after the validation.

stefanprodan:
Why not do this in CI with Kubernetes Kind and `kustomize build . | kubectl apply -f --dry-run=client`?

william-salt-cko:
I think we are all trying to avoid giving CI access to our k8s clusters, this is one of the benefits of using Flux and GitOps. Even a read only SA will still have to read secrets when doing a kustomize diff in most use cases.

Currently, we are doing kubectl dry-runs and helm diffs on our machines with bash scripts whilst we migrate to flux2 from flux1 and plain helm. 

After that, we need to build something into our CI for staging, which I don't think we will be able to do for production as it will give it access to secrets.

I think my other discussion is related to this too https://github.com/fluxcd/flux2/discussions/820

JaneLiuL:
@raffis  I just upload the code can know what will be change for flux
 flux-precheck: https://github.com/JaneLiuL/flux-precheck/releases/tag/0.0.1
Welcome that you could use to know what will be change by flux, the result will be like
You can just download the release bin file and run as `flux-precheck --kustomizationName=kustomizationname --kustomizationNamespace=namespace`

JaneLiuL:
> Why not do this in CI with Kubernetes Kind and `kustomize build . | kubectl apply -f- --dry-run=client`?
@stefanprodan  I think we can use kubectl apply --dry-run, but that only know what will be create or change, 
if we want to know what will be delete, it does not support.
So I just write a script to do that part.

mbrancato:
I think it makes sense here to have a webhook handle this and report status back. Currently, the webhook can get push updates (from GitHub, others) and trigger an internal `git pull` on the configured branch. But for CI, it makes sense to have Flux checkout a PR branch, do an in-cluster dry run. Ideally, this can report on create, update, and delete along with the details of what will be changing. I think this might require changes to the notifications to allow a process like: PR -> webhook to Flux -> dry run -> notification webhook to GitHub with result. Today, I don't think there is any way to pass data from the initial webhook into the notifications.

Issue #197, opened 2020-12-11
Health assessment for all resources of a given kind.


chhsia0:
The current health assessment mechanism requires users to provide a white list of resource references in a Kustomization resource, which introduces a tight coupling between the Kustomization resource itself and the contents retrieved from its source git repository. This is not ideal in certain use cases, for example, if you add a new deployment or remove a deployment in the git repository, you have to modify the Kustomization resource in the cluster for health assessment to work.

Ideally, it would be nice to have a mechanism for the kustomize controller to automatically detect what resources should be health-checked, but this is hard to do because not all resources are kstatus-compliant, especially for custom resources.

That said, a possible improvement is to allow health assessment to be applied on a white list of *resource kinds* provided by users. Although there is still a coupling between the Kustomization resource and the git repository contents, the coupling is not as high because adding/removing resource kinds happens less often than changing resource names.

Would it be a good idea if empty names are allowed in a health check entry to indicate that the health check applies to all resources of the same group/version/kind that is applied by this Kustomization.

Additionally, how about listing all resources to whom the health checks apply and the health results in Kustomization statuses for better observability?

stefanprodan:
> That said, a possible improvement is to allow health assessment to be applied on a white list of resource kinds provided by users.

I think this is the best way forward without breaking the current API, it can look similar to the [Alert API](https://toolkit.fluxcd.io/components/notification/alert/#example):

```yaml
  healthChecks:
    - apiVersion: apps/v1
      kind: Deployment
      name: '*'
    - apiVersion: helm.toolkit.fluxcd.io/v2beta1
      kind: HelmRelease
      name: '*'
```

Relevant discussions on this topic: https://github.com/fluxcd/flux2/discussions/324 https://github.com/fluxcd/flux2/discussions/325

chhsia0:
Yes I also thought about using asterisks.

Pros and cons of empty strings:
+ Looks more consistent with PolicyRule (which only use '*' to match apiGroup and kind, not resourceName).
- Violates the requirements of CrossNamespaceObjectReference (name is required), so we might need to define an identical Go struct with a different name, e.g., HealthCheckRule. API-wise this won't be a breaking change since Go types are implementation internals.

Pros and cons of asterisks:
+ Meets the CrossNamespaceObjectReference requirements so fully backward compatible.
- Magic string constant.

But since the Alert API already uses asterisk, it seems for consistency, using asterisks would be the way to go.

Nils98Ar:
[...]

PR #426, opened 2021-09-11
Server-side reconciliation for the v1beta2 API

### LINKED ISSUE #352: Detect changes/drift through a filtered diff of the apply/prune

### LINKED ISSUE #413: Kustomize applies and notifies when no changes occur
In a multi-cluster setup (e.g. https://github.com/fluxcd/flux2-kustomize-helm-example), the kustomize controller will apply manifests and send notifications on any change to the git repo.

Looking at this example of a `Kustomization` resource:

https://github.com/fluxcd/flux2-kustomize-helm-example/blob/f0c2aaef48461d8099a8fff05893e9ebb96f1561/clusters/production/apps.yaml#L6-L13

The `spec.path` tells Flux where the config is for this cluster. If a change is made in a different path for a different cluster, it causes the kustomize controller to trigger. If would be great if the controller could detect if there are changes (`git diff --name-only <currently applied commit> <new commit>`) in the `spec.path` and only apply when there are changes.

stefanprodan:
This PR introduces a new reconciler based on Kubernetes [server-side apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) and graduates the API to v1beta2.

## API v1beta2

The `kustomize.toolkit.fluxcd.io/v1beta2` API is backwards compatible with v1beta1, with the following deprecations and removals:
- `spec.validation` removed (server-side validation is implicit)
- `spec.patchesStrategicMerge` deprecated in favour of `spec.patches`
- `spec.patchesJson6902` deprecated in favour of `spec.patches`
- `spec.status.snapshot` replaced by `spec.status.inventory`

## SSA Reconciler

Behaviour changes:
- Creates an inventory of objects to be applied (persisted in-cluster under `.status.inventory`).
- Applies first custom resource definitions (CRDs) and namespaces, waits for them to register and only then applies the custom resources.
- Validates all resources with server-side dry-run apply (namespaced objects must contain `metadata.namespace`, defaulting to the `default` namespace is no longer supported).
- Reconciles only the resources that are in drift (viewing the diff could be made available in flux CLI or via trace events).
  Fixes #352
  Ref #213
- Prunes the objects that were previously applied but are missing from the current inventory (the garbage collector no longer annotates the objects after each reconciliation).
  Ref #415
- Emits change events based on the dry-run diff, and only for the resources that where created, configured or deleted.
    Fixes #413
- Emits health check events only for when the previous status changed.
  Fixes: #191


api/v1beta2/inventory_types.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+// Entry contains the information necessary to locate a resource within a cluster.
+type Entry struct {
+	// ObjectID is the string representation of a Kubernetes object metadata,
+	// in the format '<namespace>_<name>_<group>_<kind>'.
+	ObjectID string `json:"id"`
+
+	// ObjectVersion is the API version of this  Kubernetes object kind.
+	ObjectVersion string `json:"ver"`
```
hiddeco, line 32: Any specific reason to not simply use `version`? I personally find the `ver` shortening often confusing as it isn't an acronym.

api/v1beta2/kustomization_types.go
```diff
@@ -0,0 +1,361 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/fluxcd/pkg/apis/kustomize"
+	"github.com/fluxcd/pkg/apis/meta"
+	"github.com/fluxcd/pkg/runtime/dependency"
+)
+
+const (
+	KustomizationController   = "kustomize-controller"
```
hiddeco, line 34: What is the reason this is included in the API package? Has it any special meaning to a consumer? Otherwise I think it is a controller implementation detail and does not belong in the API.

api/v1beta2/kustomization_types.go
```diff
@@ -0,0 +1,361 @@
...
+	KustomizationKind         = "Kustomization"
+	KustomizationFinalizer    = "finalizers.fluxcd.io"
```
hiddeco, line 36: We have had a discussion about this in the past, but I think it would be wise if we would make the finalizers domain specific. Reason for this is that the pattern of using various finalizers to control the garbage collection process of an object if multiple reconcilers are dealing with an object has become more common.

I think your objection in the past was that removing "all Flux reconcilers" in cases where they get stuck would become more difficult, but if we would nest all under a sub(domain)name, you could still loop over all that match that suffix. E.g. `applier.finalizers.fluxcd.io`, `source.finalizers.fluxcd.io`.

api/v1beta2/kustomization_types.go
stefanprodan: It's used for the managed fields owner name. I think this must be part of the API as if this value changes, the controller will conflict with itself.

api/v1beta2/kustomization_types.go
stefanprodan: If we change this now, wouldn't result in orphan resources after a Flux upgrade? I would introduce this change into a followup PR.

api/v1beta2/inventory_types.go
stefanprodan: The reason is the size of the status object, `ver` repeats itself for each entry in the inventory, using `version` would take up more space.

hiddeco: In that case, why not just short it to `v` which is also used in e.g. SemVer and thus more familiar to people.

api/v1beta2/kustomization_types.go
hiddeco: Follow up PR (e.g. bundled with the other refactoring efforts) is fine with me, just wanted to have it noted given it showed up again :-)

api/v1beta2/kustomization_types.go
hiddeco: Yes, but the managed fields owner name is set by the controller. I would thus expect this value to be part of the application (contract), and not of the API that can be consumed (and theoretically, also driven) by others.

stefanprodan: Isn't the finalizer part of the application contract? I'm for moving both out the API.

hiddeco: Will need to confirm this based on some sources that I don't have at hand right now, will get back to you about this on Monday (with references).

hiddeco: Finalizers do seem to be part of the API: https://pkg.go.dev/sigs.k8s.io/cluster-api@v0.3.10-rc.1/exp/addons/api/v1alpha3#pkg-constants

SomtochiAma:
Tested this PR on kubernetes version  `1.18.8`, `1.19.1`, `1.20.2` and it works💥.
It fails on `1.18.2` with the following error:
```
{"level":"error","ts":"2021-09-13T15:02:33.963Z","logger":"controller.kustomization","msg":"unable to update status after reconciliation","reconciler group":"kustomize.toolkit.fluxcd.io","reconciler kind":"Kustomization","name":"podinfo","namespace":"flux-system","error":"failed to update object (json PATCH for kustomize.toolkit.fluxcd.io/v1beta2, Kind=Kustomization) managed fields: failed to update ManagedFields: failed to convert new object: .status.inventory: field not declared in schema"}
{"level":"error","ts":"2021-09-13T15:02:33.964Z","logger":"controller.kustomization","msg":"Reconciler error","reconciler group":"kustomize.toolkit.fluxcd.io","reconciler kind":"Kustomization","name":"podinfo","namespace":"flux-system","error":"failed to update object (json PATCH for kustomize.toolkit.fluxcd.io/v1beta2, Kind=Kustomization) managed fields: failed to update ManagedFields: failed to convert new object: .status.inventory: field not declared in schema"}
{"level":"info","ts":"2021-09-13T15:02:34.495Z","logger":"controller.kustomization","msg":"server-side apply completed","reconciler group":"kustomize.toolkit.fluxcd.io","reconciler kind":"Kustomization","name":"podinfo","namespace":"flux-system","output":{"Deployment/flux-system/podinfo":"unchanged","HorizontalPodAutoscaler/flux-system/podinfo":"unchanged","Service/flux-system/podinfo":"unchanged"}}
```
This is likely due to [k/k#91748](https://github.com/kubernetes/kubernetes/pull/91748)

api/v1beta2/inventory_types.go
```diff
@@ -0,0 +1,33 @@
...
+
+// Entry contains the information necessary to locate a resource within a cluster.
+type Entry struct {
+	// ObjectID is the string representation of a Kubernetes object metadata,
+	// in the format '<namespace>_<name>_<group>_<kind>'.
+	ObjectID string `json:"id"`
+
+	// ObjectVersion is the API version of this  Kubernetes object kind.
+	ObjectVersion string `json:"v"`
+}
```
hiddeco, line 33: Couple of questions and/or suggestions:

1. While `Inventory` as a name makes sense if you know the context, I am wondering if giving it a prefix which provides a more detailed explanation if you are just browsing through e.g. IDE suggestions would ease consumer's their life.
2. Same applies to `Entry`, which provides even less context.
3. The top doc-block comment uses "resource", while the in-struct comments refers to "Kubernetes object". When I added a bunch of helpers to the `runtime` package, I tried to consistently use "Kubernetes resource object", which is also used throughout the Kubernetes core.
4. There is a double space in the `ObjectVersion` doc block.

Suggestion based on the above:

```go
// Inventory contains a list of Kubernetes resource object references that have been applied by a Kustomization.
// Suggestion(hidde): ResourceInventory
type Inventory struct {
	// Entries of Kubernetes resource object references.
	Entries []ResourceRef `json:"entries"`
}

// ResourceRef contains the information necessary to locate a Kubernetes resource object within a cluster.
type ResourceRef struct {
	// ID is the string representation of the Kubernetes resource object's metadata,
	// in the format '<namespace>_<name>_<group>_<kind>'.
	ID string `json:"id"`

	// Version is the API version of the Kubernetes resource object's kind.
	Version string `json:"v"`
}

File: api/v1beta2/kustomization_types.go [hunk 5]
```diff
@@ -0,0 +1,364 @@
...
+)
+
+const (
+	KustomizationKind         = "Kustomization"
+	KustomizationFinalizer    = "finalizers.fluxcd.io"
+	MaxConditionMessageLength = 20000
+	DisabledValue             = "disabled"
+)
+
+// KustomizationSpec defines the desired state of a kustomization.
```
hiddeco, line 39: On the fence if this can be considered an accurate description, and/or adds any value as is.

I personally feel that while it does declare some desire, there is a lot in the object that I wouldn't call "desired state", but rather "build instructions". Making something like this maybe a bit more fitting / descriptive:

```suggestion
// KustomizationSpec defines the configuration to calculate the desired state from a Source using Kustomize.
```

api/v1beta2/kustomization_types.go
```diff
@@ -0,0 +1,364 @@
...
+func (in *Kustomization) GetStatusConditions() *[]metav1.Condition {
+	return &in.Status.Conditions
+}
+
+const (
+	// GitRepositoryIndexKey is the key used for indexing kustomizations based on their Git sources.
+	GitRepositoryIndexKey string = ".metadata.gitRepository"
+	// BucketIndexKey is the key used for indexing kustomizations based on their S3 sources.
+	BucketIndexKey string = ".metadata.bucket"
+)
```
hiddeco, line 323: Given index keys are client-specific, I think this is reconciler / application logic.

api/v1beta2/reference_types.go
```diff
@@ -0,0 +1,47 @@
...
+limitations under the License.
+*/
+
+package v1beta2
+
+import "fmt"
+
+// CrossNamespaceSourceReference contains enough information to let you locate the
+// typed referenced object at cluster level
+type CrossNamespaceSourceReference struct {
```
hiddeco, line 23:
```suggestion
// CrossNamespaceSourceReference contains enough information to let you locate the
// typed referenced Kubernetes resource object at cluster level.
type CrossNamespaceSourceReference struct {
```

api/v1beta2/reference_types.go
```diff
@@ -0,0 +1,47 @@
...
+	// API version of the referent
```
hiddeco, line 24:
```suggestion
	// API version of the referent.
```

api/v1beta2/reference_types.go
```diff
@@ -0,0 +1,47 @@
...
+	// +optional
+	APIVersion string `json:"apiVersion,omitempty"`
+
+	// Kind of the referent
```
hiddeco, line 28:
```suggestion
	// Kind of the referent.
```

api/v1beta2/reference_types.go
```diff
@@ -0,0 +1,47 @@
...
+	// +kubebuilder:validation:Enum=GitRepository;Bucket
+	// +required
+	Kind string `json:"kind"`
+
+	// Name of the referent
```
hiddeco, line 33:
```suggestion
	// Name of the referent.
```

api/v1beta2/reference_types.go
```diff
@@ -0,0 +1,47 @@
...
+	// +required
+	Name string `json:"name"`
+
+	// Namespace of the referent, defaults to the Kustomization namespace
```
hiddeco, line 37:
```suggestion
	// Namespace of the referent, defaults to the namespace of the Kubernetes resource object that contains the reference.
```

internal/ssa/manager_apply_test.go
```diff
@@ -0,0 +1,183 @@
+package ssa
```
hiddeco, line 1: Missing copyright header.

api/v1beta2/inventory_types.go
stefanprodan: I've updated the inventory according to your suggestions. Thanks!

squaremo:
> The `kustomize.toolkit.fluxcd.io/v1beta2` API is backwards compatible with `v1beta1`.
> [...]
> * `.spec.validation` removed (server-side validation is implicit)

Is that field ignored or removed, and if the latter, does this mean a v1beta1 resource won't validate as v1beta2?

stefanprodan:
> Is that field ignored or removed, and if the latter, does this mean a v1beta1 resource won't validate as v1beta2?

I haven't bumped the API version in the e2e tests to prove that the v1beta1 CRs work as expected. What happens is that the Kubernetes API drops all unknown fields before validating the CR, so validation works since `.spec.validation` is removed.

squaremo:
> * When `.spec.wait` is `true`, the controller runs the health assessment for all resources ignoring the ones in `spec.healthChecks`.
>   Fixes [Health assessment for all resources of a given kind. #197](https://github.com/fluxcd/kustomize-controller/issues/197)

In the linked issue, it's explained that waiting on everything wouldn't work because some resource types are incompatible with kstatus (and this is the motivation for suggesting an allow-list or patterns). The change quoted above would make the controller wait on everything -- so is it the case that

 1. the issue report is mistaken, and it's fine to wait on everything
 2. the mechanism here works around incompatibility with kstatus
 3. this will not work when you're using such resources

stefanprodan:
@squaremo all Kubernetes builtin kinds are compatible with kstatus. As for custom resources, if the those have no status, or no ready condition is status, or no observed generation, then kstatus instead of crashing will report that resource as ready.

squaremo:
> Prunes the objects that were previously applied but are missing from the current inventory (the garbage collector no longer annotates the objects after each reconciliation).

There's lots of upsides to this approach, but also a downside: you are storing state that you rely on in `.status`, rather than observing the state of the cluster (this was partially true of the prior implementation too, but in principle it did not _rely_ on it). I think it is worth thinking through the things that could go wrong as a consequence, and if there are remedies. For instance,

 - It's possible to have a partial failure when reconciling, creating objects but failing to record them in the inventory. What happens then? Will it come right the next time it succeeds? Can the user detect this situation, and can they fix it?
 - Can two Kustomizations fight over a resource (e.g., if you move a resource from one kustomization to another, what happens)?
 - Are there any problems with going from one implementation of garbage collection to another? (i.e., when you have an empty inventory, but not an empty set of objects that were created by Flux)

stefanprodan:
> It's possible to have a partial failure when reconciling, creating objects but failing to record them in the inventory. What happens then? Will it come right the next time it succeeds? Can the user detect this situation, and can they fix it?

The apply error is recorded in the ready status, events, alerts logs etc. When the user fixes the issue, the controller will update the inventory accordingly and GC will cleanup the stale objects (same behaviour as before).

> Can two Kustomizations fight over a resource (e.g., if you move a resource from one kustomization to another, what happens)?

Yes but we can improve the situation and allow users to migrate resources from one Kustomization to another without having to disable pruning via annotations. 

> Are there any problems with going from one implementation of garbage collection to another? (i.e., when you have an empty inventory, but not an empty set of objects that were created by Flux)

If the upgrade happens at the same time with a config change, then yes, the GC will skip the objects removed as it has no prior recored of them. But I think this is an edge case with slim chances, you need to push to upstream and upgrade Flux at the same time.

dsogari:
Why can't we have the option to disable the dry-run behavior? The implicit behavior fails for me when deploying a `kube-prometheus` stack:

> Pod/monitoring/kube-prometheus-grafana-test dry-run failed, reason: Forbidden, error: pods "kube-prometheus-grafana-test" is forbidden: error looking up service account monitoring/kube-prometheus-grafana-test: serviceaccount "kube-prometheus-grafana-test" not found

It fails because the service account mentioned is only created during the actual reconciliation.

PR #433, opened 2021-09-28
Update kustomize to v4.4.0

hiddeco:
This release should include a fix for YAML anchors.

Ref: https://github.com/fluxcd/flux2/issues/1522#issuecomment-925731858


stefanprodan:
Included in #426

PR #439, opened 2021-10-06
envtest: Add cancellable context to stop controllers

darkowlzz:
In suite test, the context created by `SetupSignalHandler()` watches for
shutdown signal to cancel the context. This makes it possible to stop
the controllers by sending a kill signal that cancels the context.

This change allows controller context cancellation by creating another
context from `SetupSignalHandler()` context with a `CancelFunc` that's
called at the end of the TestMain, instead of sending a kill signal.

__NOTE__: I wasn't able to reproduce the failure due to hang issue when
I removed the kill call. But I did notice extra time required to end the tests
without the kill call. By replacing the kill call with CancelFunc, I observed
the same short test cleanup time as with the kill call.

### DIFF
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -27,7 +27,6 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"syscall"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


stefanprodan (approved):
LGTM

Thanks @darkowlzz I tested this locally and it works great!

PR #458, opened 2021-10-12
Use tonistiigi/xx for cross-platform builds

darkowlzz:
Update the Dockerfile to use tonistiigi/xx to cross-compile the
manager binaries natively and speed up the multi-arch image build
time.

Refer fluxcd/flux2#1910

__NOTE__: Multi-arch build used to take about 7 minutes on my
machine. Cross-compiling reduces the multi-arch build to under 3
minutes.
Manually tested the built image on amd64 and arm/v7 (raspberry pi).
Haven't tested on arm64.

__Observations of the build time in github actions CI__: Building all the
images for multi-arch builds takes about 6 minutes, compared to 42
minutes before.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -12,6 +12,15 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+      - name: Setup QEMU
+        uses: docker/setup-qemu-action@v1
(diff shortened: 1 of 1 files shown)


hiddeco:
This PR didn't take into account the changes to the GitHub workflows as I explicitly documented in the issue.

PR #463, opened 2021-10-14
SOPS: Decrypt dotenv files used in kustomize secret generator

somtochiama:
This PR adds support for decrypting dotenv files (encrypted with SOPS) that are used with kustomize secret generator.

## Example

Create an env file:
```
ROUTER_PASSWORD=admin
DB_PASSWORD=admin
```

Encrypt it with SOPS:
```
sops --encrypt --in-place podinfo.env
```

Create a kustomization.yaml that generates an env secret:

```
kind: Kustomization
namespace: default
secretGenerator:
  - name: env-secret
    envs:
      - podinfo.env
```

Configure Flux to decrypt the secret before apply:

```yaml
kind: Kustomization
metadata:
  name: secrets
  namespace: flux-system
spec:
  decryption:
    provider: sops
    secretRef:
      name: sops-age
  interval: 1m0s
  path: ./secrets
  prune: true
  sourceRef:
    kind: GitRepository
    name: flux-system
```

Ref: https://github.com/fluxcd/flux2/issues/779


danufr:
Guys, is this functionality works only with secretGenerator or with configMapGenerator also?

kingdonb:
@danufr I have not tried it, but from reading:

https://github.com/fluxcd/kustomize-controller/pull/463/files#diff-efcb6cc0bdb62e54699cd11dfb9fbbab9f2ae76ada019846e853667ab29c86b0R216-R218

I would say it looks to me like this feature only works with `secretGenerator`. The `decryptDotEnvFiles` function does not go and decrypt every .env.enc with a wildcard everywhere in the Kustomization path, it only decrypts those that are in the path of an actual `secretGenerator` (and then, only if the file contains a string that matches the format of an encrypted data file.)

PR #478, opened 2021-10-27
Update Go to v1.17 and controller-runtime to v0.11

relu:
Part of: https://github.com/fluxcd/flux2/issues/2016


Dockerfile
```diff
@@ -1,8 +1,9 @@
-ARG XX_VERSION=1.0.0-rc.2
+ARG GO_VERSION=1.17.5
```
stefanprodan, line 1: Can we please not hardcode the Go patch version

relu: Yes, I'm going to make this change everywhere I also just noticed it and am preparing a PR for NC and SC.

stefanprodan (approved):
LGTM

Thanks @relu 

PS. I've updates source-controller to latest version that contains controller-runtime 0.11

PR #480, opened 2021-11-04
Remove gopass dependency 

### LINKED ISSUE #477: Find a replacement for gopass dependency
From Ada Logics:

Looking at https://github.com/fluxcd/kustomize-controller/blob/76eb4833d2aa8e169d0cc0d9993db32ad62f24bd/go.mod the following mod should potentially be updated:

- github.com/hashicorp/go-retryablehttp v0.6.8
- google.golang.org/grpc v1.38.0
- github.com/howeyc/gopass v0.0.0-20170109162249-bf9dde6d0d2c (this was **archived** upstream)

relu:
Fixes #477

### DIFF
--- a/internal/sops/pgp/keysource.go
+++ b/internal/sops/pgp/keysource.go
@@ -17,8 +17,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/howeyc/gopass"
-	gpgagent "go.mozilla.org/gopgagent"
 	"golang.org/x/crypto/openpgp"
 	"golang.org/x/crypto/openpgp/armor"
 )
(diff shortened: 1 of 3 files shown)


internal/sops/pgp/keysource.go
```diff
@@ -332,7 +332,7 @@ func (key *MasterKey) passphrasePrompt() func(keys []openpgp.Key, symmetric bool
 		conn, err := gpgagent.NewConn()
 		if err == gpgagent.ErrNoAgent {
 			fmt.Print("Enter PGP key passphrase: ")
-			pass, err := gopass.GetPasswd()
+			pass, err := term.ReadPassword(int(os.Stdin.Fd()))
```
stefanprodan, line 335: I think we should error out here, there is no stdin available in the controller. This way we can get rid of x/term.

hiddeco: This was originally copied over to maintain full compatibility with SOPS' way of working, while breaking free from the user (environment) specific assumptions in upstream.

But I do agree dropping the dependency is better.

relu: Thanks, guys. I was actually going to ask you about it since I was confused about how to test this and I noticed in the docs it is explicitly mentioned that only passphrase-less keys are supported. Removing the dependency makes perfect sense in this case. :ok_hand:

stefanprodan (approved):
LGTM

Thanks @relu 🏅

PR #481, opened 2021-11-04
Lowercase first letters for fmt.Errorf messages

relu:
Some of these error messages were likely imported from somewhere else.


hiddeco (approved):
Awesome, thanks :broom::dash:

PR #483, opened 2021-11-08
Warn when secrets are not decrypted before apply

stefanprodan:
If decryption is not enabled, SOPS encrypted secrets will fail to apply with a validation error that doesn't give any hints. It's better to exit early and throw an error that tells users to enable decryption.

### DIFF
--- a/controllers/kustomization_decryptor.go
+++ b/controllers/kustomization_decryptor.go
@@ -33,6 +33,7 @@ import (
 	"go.mozilla.org/sops/v3/cmd/sops/formats"
 	"go.mozilla.org/sops/v3/keyservice"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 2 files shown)


controllers/kustomization_decryptor.go
```diff
@@ -21,6 +21,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
```
hiddeco, line 24: Should be ordered.

controllers/kustomization_decryptor.go
```diff
@@ -288,3 +289,13 @@ func isDir(path string) (bool, error) {
 
 	return fileInfo.IsDir(), nil
 }
+
+// IsEncryptedSecret checks if the given object is a Kubernetes Secret encrypted with Mozilla SOPS.
+func IsEncryptedSecret(object *unstructured.Unstructured) bool {
```
hiddeco, line 294: :+1:

hiddeco (approved):

PR #484, opened 2021-11-09
Allow disabling the reconciliation of in-cluster resources

stefanprodan:
Introduce `kustomize.toolkit.fluxcd.io/reconcile` annotation. When set to `disabled`, the controller will no longer apply changes from source, nor it will prune the annotated resource.

Ref: https://github.com/fluxcd/flux2/issues/614


hiddeco (commented):
I would expect the label to either be `%s/reconciliation: disabled` or `%s/reconcile: false`.

docs/spec/v1beta2/kustomization.md
```diff
@@ -334,6 +334,16 @@ kubectl get all --all-namespaces \
...
 ```
 
+You can configure the controller to ignore in-cluster resources by labeling or annotating them:
+
+```sh
+kubectl annotate service/podinfo kustomize.toolkit.fluxcd.io/reconcile=disabled
+```
+
+Note that when the `kustomize.toolkit.fluxcd.io/reconcile` annotation is set to `disabled`,
+the controller will no longer apply changes from source, nor it will prune the resource.
```
hiddeco, line 344:
```suggestion
the controller will no longer apply changes from source, nor will it prune the resource.
```

hiddeco (approved):

PR #489, opened 2021-11-12
Verify artifacts integrity

stefanprodan:
After downloading an artifact, compute its checksum and verify that it matches the original checksum as advertised by source-controller.

For backwards compatibility with source-controller v0.17.2, the checksum is computed with SHA1, on clusters with a newer source-controller we use SHA256.

Addresses https://github.com/fluxcd/source-controller/issues/468

### DIFF
--- a/controllers/kustomization_controller.go
+++ b/controllers/kustomization_controller.go
@@ -19,7 +19,10 @@ package controllers
 import (
 	"bytes"
 	"context"
+	"crypto/sha1"
+	"crypto/sha256"
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
 	"os"
@@ -296,7 +299,7 @@ func (r *KustomizationReconciler) reconcile(
 	defer os.RemoveAll(tmpDir)
 
 	// download artifact and extract files
-	err = r.download(source.GetArtifact().URL, tmpDir)
+	err = r.download(source.GetArtifact(), tmpDir)
 	if err != nil {
 		return kustomizev1.KustomizationNotReady(
 			kustomization,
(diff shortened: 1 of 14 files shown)


controllers/kustomization_controller.go
```diff
@@ -521,14 +525,44 @@ func (r *KustomizationReconciler) download(artifactURL string, tmpDir string) er
...
+	if len(artifact.Checksum) == 40 {
+		hasher = sha1.New()
+	}
+
+	// compute checksum
+	_, _ = io.Copy(hasher, reader)
+	checksum := fmt.Sprintf("%x", hasher.Sum(nil))
+
+	if artifact.Checksum != checksum {
+		return fmt.Errorf("failed to verify artifact, the computed checksum '%s' doesn't match the original '%s'",
```
hiddeco, line 559:
```suggestion
		return fmt.Errorf("failed to verify artifact: computed checksum '%s' doesn't match advertised '%s'",
```

controllers/kustomization_controller.go
```diff
@@ -521,14 +525,45 @@ func (r *KustomizationReconciler) download(artifactURL string, tmpDir string) er
...
+	if _, err := io.Copy(mw, reader); err != nil {
+		return err
+	}
+
+	checksum := fmt.Sprintf("%x", hasher.Sum(nil))
+
+	if artifact.Checksum != checksum {
+		return fmt.Errorf("failed to verify artifact: computed checksum '%s' doesn't match advertised '%s'",
+			checksum, artifact.Checksum)
+	}
```
hiddeco, line 562:
```suggestion
	if checksum := fmt.Sprintf("%x", hasher.Sum(nil)); checksum != artifact.Checksum {
		return fmt.Errorf("failed to verify artifact: computed checksum '%s' doesn't match advertised '%s'",
			checksum, artifact.Checksum)
	}
```

hiddeco (approved):
LGTM, thanks for following through on the changes :bow: :100:

PR #430, opened 2021-09-17
Draft: Prebuild Stage Feature

oliverbaehler:
I tried implementing a solution for issue #414 
Since this is my first contribution in go i am not quiet sure how good it is. But i tested it and it works. But hard to tell.

I created this PR as Draft, so we can discuss over the feature.
If the feature is seen as plausible, I will commit further changes (tests, etc) which comply with the contributions guidelines. 

I didn't make them yet, because I don't   want to waste time for something that could be rejected.

Signed-off-by: Oliver Bähler <oliverbaehler@hotmail.com>

PR #493, opened 2021-11-18
Add support for Azure Keyvault

dquagebeur:
This PR is to support Azure Keyvault for SOPS. 
Azure credentials have to be provided in the SOPS secret under the key 'azure_kv.json'.
This PR is for discussion as it is based on another PR (#430) that we already use internally.


stefanprodan:
@dquagebeur please keep this PR scoped to Azure KeyVault.

internal/sops/keyservice/server.go
```diff
@@ -109,6 +256,18 @@ func (ks Server) Encrypt(ctx context.Context,
 		response = &keyservice.EncryptResponse{
 			Ciphertext: ciphertext,
 		}
+	case *keyservice.Key_AzureKeyvaultKey:
+		if _, err := os.Stat(filepath.Join(ks.HomeDir, "azure_kv.json")); os.IsNotExist(err) {
```
stefanprodan, line 260: The RootFS is readonly, there is no way `azure_kv.json` could be in the home dir. Please remove this code.

dquagebeur: ks.HomeDir is not the root fs but a temporary directory already used by the decryptor for gpg.

Issue #324, opened 2021-04-15
SOPS multi-tenancy


moolen:
Context:
* running on EKS, using IRSA and pod-identity-webhook
* multiple tenants should be onboarded onto the cluster
* we're enforcing `spec.serviceAccountName` 
* one tenant should not be able to decrypt secrets of other tenants

Problem:
Right now, `kustomize-controller` uses a local set of credentials that are defined by environment variables. These credentials allow the controller to make API calls against AWS. 
I want to onboard multiple tenants and they should not be able to decrypt the secrets of other tenants. However, since they share the same set of credentials they can decrypt secrets of other tenants.

With GPG we have a better tenant isolation due to the fact that `Kind=Kustomization` is namespaced and it stores a local reference to a gpg private key. (i.e. one gpg-key per namespace).

Proposed solutions:

(1) use multiple flux instances
This is meh for obvious reasons. Question: can i have multiple `kustomize-controllers` ? Would you support having sth. similar to `ingressClassName` to shard responsibilities between controllers?

(2) use serviceAccount/TokenRequest API 
Since the Kustomization refers to an `spec.serviceAccountName` we could implement sth. similar to the service account token volume projection API in Kubelet ([docs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection)).
We can feed those tokens into the aws sdk to do a `sts:assumeRoleWithWebIdentity` this needs to be implemented in the sops repo, of course.
With that we basicially mimic the behavior of the eks-pod-identity-webhook but have the ability to assume multiple roles within one pod. I'm not sure yet what implications that has to elevate the privileges to use the token request api.

travisgroth:
Before seeing this issue, I opened https://github.com/fluxcd/flux2/issues/1478.  Not sure if that should move to this repo but it seems related to the ask here.  Would something like that support your use case @moolen?

FWIW, I think currently you can run a `kustomize-controller` per namespace by setting `--watch-all-namespaces=false` for multi-tenant use cases.  I'd prefer something with less operational overhead when you're primarily worried about segmenting secrets access.

dquagebeur:
We would like to use Azure Keyvault instead of pgp keys. What do you think about this proposal:
Integrate Azure Keyvault SDK in the keyservice to encrypt/decrypt using the azure crendentials stored in a secret from the current namespace (behaviour similar to current pgp implementation).
[...]

PR #495, opened 2021-11-19
SOPS: add support for Azure Key Vault credentials using SecretRef

dquagebeur:
This PR is to support Azure Keyvault for SOPS.
Azure credentials have to be provided in the SOPS secret under the key 'azure_kv.json'.


dquagebeur:
I have rebased my commit on the main branch to remove the preBuild dependency.
As we already use it, could you give your feedback on this PR too (#430). If you agree with it, after this PR, I will rebase the other one.

.gitignore
```diff
@@ -18,3 +18,5 @@ bin/
 config/release/
 config/crd/bases/gitrepositories.yaml
 config/crd/bases/buckets.yaml
+
+kustomize-controller
```
hiddeco, line 22: This shouldn't be required.

hiddeco (commented):
Will try to get this reviewed on Monday, thanks in advance @dquagebeur :bow:

stefanprodan:
> This PR is to support Azure Keyvault for SOPS.

This is misleading, Flux does supports Azure Key Vault, here is the e2e tests that proves this https://github.com/fluxcd/flux2/blob/main/tests/azure/azure_test.go#L525

dquagebeur:
Yes, it's true.
To be exact, this PR adds a per namespace support for Azure Key Vault . 
With it, because credentials are provided as a namespace's secret, each namespace can have its own key vault and be sure that other namespaces cannot decrypt its secrets.

stefanprodan:
> To be exact, this PR adds a per namespace support for Azure Key Vault .

Right! This makes Flux usable on multi-tenant AKS clusters, it does solve the issue raised in #324 but only for Azure Key Vault.

stefanprodan:
@dquagebeur please signoff your commits, see https://github.com/fluxcd/kustomize-controller/pull/495/checks?check_run_id=4266364175

internal/sops/keyservice/server.go
```diff
@@ -109,6 +256,18 @@ func (ks Server) Encrypt(ctx context.Context,
 		response = &keyservice.EncryptResponse{
 			Ciphertext: ciphertext,
 		}
+	case *keyservice.Key_AzureKeyvaultKey:
+		if _, err := os.Stat(filepath.Join(ks.HomeDir, "azure_kv.json")); os.IsNotExist(err) {
```
hiddeco, line 260: `azure_kv.json` should be a `const` to ensure consistent usage.

internal/sops/keyservice/server.go
```diff
@@ -169,6 +328,18 @@ func (ks Server) Decrypt(ctx context.Context,
 		response = &keyservice.DecryptResponse{
 			Plaintext: plaintext,
 		}
+	case *keyservice.Key_AzureKeyvaultKey:
+		if _, err := os.Stat(filepath.Join(ks.HomeDir, "azure_kv.json")); os.IsNotExist(err) {
+			return ks.DefaultServer.Decrypt(ctx, req)
+		} else {
+			plaintext, err := ks.decryptWithAzureKeyvault(k.AzureKeyvaultKey, req.Ciphertext)
```
hiddeco, line 335: `azure_kv.json` should be a `const` to ensure consistent usage.

internal/sops/keyservice/server.go
```diff
@@ -86,6 +101,138 @@ func (ks *Server) decryptWithAge(key *keyservice.AgeKey, ciphertext []byte) ([]b
...
+
+	kv := keyvault.New()
+	if kv.Authorizer, err = ks.newKeyvaultAuthorizerFromFile(filepath.Join(ks.HomeDir, "azure_kv.json")); err != nil {
+		return nil, err
+	}
+
+	plainstring := string(plaintext)
+	result, err := kv.Encrypt(context.Background(), key.VaultUrl, key.Name, "", keyvault.KeyOperationsParameters{Algorithm: keyvault.RSAOAEP256, Value: &plainstring})
+	if err != nil {
+		return nil, fmt.Errorf("Failed to encrypt data: %w", err)
```
hiddeco, line 203:
```suggestion
		return nil, fmt.Errorf("failed to encrypt data: %w", err)
```

(applies to all errors)

controllers/kustomization_decryptor.go
```diff
@@ -160,6 +160,11 @@ func (kd *KustomizeDecryptor) ImportKeys(ctx context.Context) error {
 				if err := kd.gpgImport(keyPath); err != nil {
 					return err
 				}
+			case ".json":
+				keyPath := filepath.Join(kd.homeDir, name)
+				if err := os.WriteFile(keyPath, file, os.ModePerm); err != nil {
```
hiddeco, line 165: We are writing all `.json` files but only seem to be interested in `azure_kv.json`, this has to change so it becomes consistent.

main.go
```diff
@@ -24,7 +24,7 @@ import (
 	flag "github.com/spf13/pflag"
 	"k8s.io/apimachinery/pkg/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
-	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
```
hiddeco, line 27: We should only explicitly include the ones we use, as loading other ones may have side-effects.

internal/sops/keyservice/server.go
dquagebeur: yes, you're right

controllers/kustomization_decryptor.go
dquagebeur: that's right, but i have done it that way for two reasons :
- other file (.asc .agekey) works the same way, based on extension
- if someone add gcp support later, it probably also needs a json credential file

hiddeco: > other file (.asc .agekey) works the same way, based on extension

But in that case the files are also taken into account.

> if someone add gcp support later, it probably also needs a json credential file

This would be another single JSON file, likely with a specific filename. Given this, I think something like the following would be better:

```
case keyservice.SOPSCredentialsFileAzureKeyvault:
	keyPath := filepath.Join(kd.homeDir, name)
	if err := os.WriteFile(keyPath, file, os.ModePerm); err != nil {
```

internal/sops/keyservice/server.go
```diff
@@ -14,6 +23,17 @@ import (
...
+	"github.com/Azure/azure-sdk-for-go/profiles/latest/keyvault/keyvault"
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/azure"
+	"github.com/Azure/go-autorest/autorest/azure/auth"
+	"github.com/dimchansky/utfbom"
+)
+
+const (
+	// SOPS credentials files names
+	SOPSCredentialsFileAzureKeyvault string = "azure_kv.json"
```
hiddeco, line 36:
```suggestion
	// SOPSCredentialsFileAzureKeyvault is the expected filename for a JSON file containing an Azure Key Vault authentication file.
	SOPSCredentialsFileAzureKeyvault = "azure_kv.json"
```

internal/sops/keyservice/server.go
```diff
@@ -86,6 +106,138 @@ func (ks *Server) decryptWithAge(key *keyservice.AgeKey, ciphertext []byte) ([]b
 	return plaintext, err
 }
 
+func (ks *Server) newKeyvaultAuthorizerFromFile(fileLocation string) (autorest.Authorizer, error) {
```
hiddeco, line 109: Instead of embedding this logic into the Server, this should move to a separate `azure` package within `sops`, as done for the other key services.

dquagebeur: I have named the package azure_kv for consistancy, but ok.

hiddeco: Can you name it `azkv` in that case, that aligns with SOPS itself: https://github.com/mozilla/sops/tree/master/azkv

controllers/kustomization_decryptor.go
```diff
@@ -148,20 +148,27 @@ func (kd *KustomizeDecryptor) ImportKeys(ctx context.Context) error {
...
+						return err
+					}
+					if err := os.WriteFile(keyPath, file, os.ModePerm); err != nil {
+						return fmt.Errorf("unable to write key to storage: %w", err)
+					}
+					if err := kd.gpgImport(keyPath); err != nil {
+						return err
+					}
+				case ".agekey":
+					ageIdentities = append(ageIdentities, string(file))
```
hiddeco, line 170: Think this can be simplified to:

```go
if name == intkeyservice.SOPSCredentialsFileAzureKeyvault {
	keyPath := filepath.Join(kd.homeDir, name)
	if err := os.WriteFile(keyPath, file, os.ModePerm); err != nil {
		return fmt.Errorf("unable to write '%s' to storage: %w", name, err)
	}
	continue
}
switch filepath.Ext(name) {
	case ".asc":
...
````

**Comment by dquagebeur** [comment]
hi @hiddeco, do you have any more changes before validating the PR ?

File: internal/sops/azkv/keysource.go [hunk 10]
```diff
@@ -0,0 +1,195 @@
...
+	return err
+}
+
+// AADSettings contains the selection of fields from an Azure authentication file
+// required for Active Directory authentication.
+//
+// It is based on the unpublished contract in
+// https://github.com/Azure/go-autorest/blob/c7f947c0610de1bc279f76e6d453353f95cd1bfa/autorest/azure/auth/auth.go#L331-L342,
+// which seems to be due to an assumption of configuration through environment
+// variables over file based configuration.
```
hiddeco, line 56: @phillebaba is there another file format more people might be familiar with that e.g. is provided by the UI (or tutorials around AKS)?

phillebaba: I think we resolved this during our Slack discussion :smiley:

internal/sops/azkv/keysource.go
```diff
@@ -0,0 +1,195 @@
...
+func (key *MasterKey) SetEncryptedDataKey(enc []byte) {
+	key.EncryptedKey = string(enc)
+}
+
+// Encrypt takes a SOPS data key, encrypts it with Key Vault and stores the result in the EncryptedKey field.
+func (key *MasterKey) Encrypt(plaintext []byte) error {
+	c := newKeyvaultClient(key.authorizer())
+
+	data := base64.RawURLEncoding.EncodeToString(plaintext)
+	p := keyvault.KeyOperationsParameters{Value: &data, Algorithm: keyvault.RSAOAEP256}
```
phillebaba, line 104: Might not be relevant today but hard coding the key algo may hurt us in the future.

internal/sops/azkv/keysource.go
```diff
@@ -0,0 +1,195 @@
...
+func (key *MasterKey) authorizer() autorest.Authorizer {
+	if key.token == nil {
+		return &autorest.NullAuthorizer{}
+	}
+	return autorest.NewBearerAuthorizer(key.token)
+}
+
+func newKeyvaultClient(authorizer autorest.Authorizer) keyvault.BaseClient {
+	c := keyvault.New()
+	c.Authorizer = authorizer
```
phillebaba, line 172: The base client should be configured with proper rate limit back off. It sounds weird but it is totally possible to hit the rate limits for Key Vault, especially in poorly designed systems that share vaults. Adding back off settings alleviates a lot of the pressure. In some situations it becomes an infinite loop of throttling if you have multiple clients that are all just retrying rapidly.

More details can be found in the docs.
https://docs.microsoft.com/en-us/azure/key-vault/general/overview-throttling

hiddeco:
I need to have another round on Microsoft internals on how they present things to their (groups) of users, as I am getting the sense that there are multiple instructions out there that people may use as a starting point. Meaning that we either need to guide them to not do this, or support multiple JSON structures.

In addition, I need to have another look at the (various) SDK(s) we built upon, to ensure that besides rate limiting, I am not missing another setting to facilitate smooth operations. Thanks for making me aware of this @phillebaba :bow: 

In any case, and given this touches the area of security, I do not want to rush this. Which means that with an eye on the holiday period that's arriving soon, many of us enjoying some well deserved time off then, and my current pile of TODOs. I think this is more likely to land at the beginning of 2022.

internal/sops/azkv/keysource.go
hiddeco: The keysource (like the other keysources), are forks of the SOPS native ones but with modifications to ensure isolation. This setting is thus aligned with SOPS (non-configurable) default behavior.

internal/sops/azkv/keysource.go
hiddeco: https://docs.microsoft.com/en-us/azure/aks/kubernetes-service-principal?tabs=azure-cli#manually-create-a-service-principal

stefanprodan (approved):
LGTM

Thanks @hiddeco and @dquagebeur 🏅

marshallford:
Out of curiosity, what was the reasoning behind changing the pinned version of kustomize from `v4.5.2` to `v4.4.1`? I often look to that comment to determine which version of `kustomize` to use elsewhere in my CI/testing process to best replicate how the controller will render manifests.

stefanprodan:
@marshallford the version got bork during merge, we’ll do a patch release today with the right one, thanks for reporting this.

PR #501, opened 2021-11-23
Document behaviour when changes are made to fields not stored in git 

SomtochiAma:
Closes https://github.com/fluxcd/flux2/issues/2099

### DIFF
--- a/docs/spec/v1beta2/kustomization.md
+++ b/docs/spec/v1beta2/kustomization.md
@@ -344,6 +344,11 @@ Note that when the `kustomize.toolkit.fluxcd.io/reconcile` annotation is set to 
 the controller will no longer apply changes from source, nor will it prune the resource.
 To resume reconciliation, set the annotation to `enabled` or remove it.
 
+Note that due to the way [Kubernetes server-side apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/)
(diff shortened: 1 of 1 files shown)


docs/spec/v1beta2/kustomization.md
```diff
@@ -344,6 +344,11 @@ Note that when the `kustomize.toolkit.fluxcd.io/reconcile` annotation is set to
 the controller will no longer apply changes from source, nor will it prune the resource.
 To resume reconciliation, set the annotation to `enabled` or remove it.
 
+Note:
```
stefanprodan, line 347: This can be "Note that due to...."

docs/spec/v1beta2/kustomization.md
```diff
@@ -344,6 +344,11 @@ Note that when the `kustomize.toolkit.fluxcd.io/reconcile` annotation is set to
...
+Due to the way kubernetes server-side apply works, the kustomize-controller can only revert
```
stefanprodan, line 348:
```suggestion
Due to the way [Kubernetes server-side apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) works, the kustomize-controller can only revert
```

stefanprodan (approved):

PR #513, opened 2021-12-09
sops: ensure proper wiring to default server

souleb:
make sure we use the right service to encrypt.

### DIFF
--- a/internal/sops/keyservice/server.go
+++ b/internal/sops/keyservice/server.go
@@ -110,7 +110,7 @@ func (ks Server) Encrypt(ctx context.Context,
 			Ciphertext: ciphertext,
 		}
 	default:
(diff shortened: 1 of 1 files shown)


stefanprodan:
> make sure we use the right service to encrypt

But we don't encrypt anything

hiddeco:
That does not really matter, as the intent of the full implementation is that it should work anyway for e.g. potential further integration in the Flux CLI (or something).

This is technically a bug, and is right to be addressed :-).

hiddeco (approved):

hiddeco:
@souleb the commit takes care of encryption but mentions "decryption".

souleb:
@hiddeco sorry 🙏 , I have rewritten the commit message.

PR #521, opened 2021-12-17
Add condition to checkDependencies when SourceRef is the same

HardBrainer:
As discussed in https://cloud-native.slack.com/archives/CLAJ40HV3/p1639665724208200 we identified a race condition in checkDependencies when using Receiver and both Kustomizations have the same SourceRef. 

If triggered by the Receiver the checkDependencies is executed before the dependent Kustomization applied the latest Revision. To prevent this we check if the LastAppliedRevision is already the new Revision.

In this Case deployment-demoapp has an dependsOn to init-demoapp. This time deployment-demoapp waits untils init-demoapp applied the latest Revision:

image

### DIFF
--- a/controllers/kustomization_controller.go
+++ b/controllers/kustomization_controller.go
@@ -205,7 +205,7 @@ func (r *KustomizationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 	// check dependencies
 	if len(kustomization.Spec.DependsOn) > 0 {
-		if err := r.checkDependencies(kustomization); err != nil {
+		if err := r.checkDependencies(source, kustomization); err != nil {
 			kustomization = kustomizev1.KustomizationNotReady(
 				kustomization, source.GetArtifact().Revision, meta.DependencyNotReadyReason, err.Error())
 			if err := r.patchStatus(ctx, req, kustomization.Status); err != nil {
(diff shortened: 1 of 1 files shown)


controllers/kustomization_controller.go
```diff
@@ -490,6 +490,10 @@ func (r *KustomizationReconciler) checkDependencies(kustomization kustomizev1.Ku
 		if !apimeta.IsStatusConditionTrue(k.Status.Conditions, meta.ReadyCondition) {
 			return fmt.Errorf("dependency '%s' is not ready", dName)
 		}
+		
+		if len(k.Status.Conditions) == 0 || k.Spec.SourceRef.Name == kustomization.Spec.SourceRef.Name && k.Spec.SourceRef.Namespace == kustomization.Spec.SourceRef.Namespace && source.GetArtifact().Revision != k.Status.LastAppliedRevision {
```
stefanprodan, line 494: There is no need for `len(k.Status.Conditions) == 0` this is evaluated above. Another issue is that we only check the source name and namespace, but we really need to check the kind as well, as a GitRepository can have the same name as a Bucket.

HardBrainer: Fair enough i already thought about other specs to check but forgot the kind.

HardBrainer: I added the kind Check and remove the duplicated Conditions Check

stefanprodan (changes requested):
Please run `make fmt` commit the changes then squash all commits into one. Thanks!

HardBrainer:
Hey,

i performed a rebase on my branch and squashed all commits into one. `make fmt` did not create any new changes.

stefanprodan (approved):
LGTM

Thanks @HardBrainer 🏅

PR #522, opened 2021-12-21
Fix inconsistent code-style raised at security audit

pjbgf:
Add check for nil to fix inconsistent code-style raised at the security audit.

Relates to https://github.com/fluxcd/pkg/issues/173


stefanprodan (approved):
LGTM

Thanks @pjbgf

PR #524, opened 2022-01-04
Capture errors from file close operation

tomhuang12:
This is to address issue outlined in https://github.com/fluxcd/pkg/issues/174. I am following the second recommendation from the [blog](https://www.joeshaw.org/dont-defer-close-on-writable-files/) to capture potential error from file close operation.

### DIFF
--- a/internal/sops/pgp/keysource.go
+++ b/internal/sops/pgp/keysource.go
@@ -295,7 +295,12 @@ func (key *MasterKey) loadRing(path string) (openpgp.EntityList, error) {
 	if err != nil {
 		return openpgp.EntityList{}, err
 	}
-	defer f.Close()
+	defer func() {
+		cerr := f.Close()
+		if err == nil {
+			err = cerr
+		}
+	}()
 	keyring, err := openpgp.ReadKeyRing(f)
 	if err != nil {
 		return keyring, err


internal/sops/pgp/keysource.go
```diff
@@ -295,7 +295,12 @@ func (key *MasterKey) loadRing(path string) (openpgp.EntityList, error) {
 	if err != nil {
 		return openpgp.EntityList{}, err
 	}
-	defer f.Close()
+	defer func() {
+		cerr := f.Close()
+		if err == nil {
+			err = cerr
```
pjbgf, line 301: `err` will never be returned on this specific implementation. Note that on the example in the blog you referred to, the `func` using this has a [named return parameter](https://www.geeksforgeeks.org/named-return-parameters-in-golang/) which they leverage to return the error value back to its callers.

Also note that the problem is only when `f.Close()` is called in write operations, which is not the case here as `ReadKeyRing` expects a `io.Reader`.

Nalum:
The only issue I see here is that you're not using the named return variables so the error would not be handled.

In this particular case would it be better to open the file in read only mode? It seems like the `Close` issue is around writing the file and here there doesn't appear to be any writes happening with the file.

pjbgf (commented):
@tomhuang12 thank you for taking a look at this. 

I have just gone through this specific file and the only case I could find in which files could be used for write (via `io.WriteCloser`) is properly handled as it is not deferring its Close operation: https://github.com/fluxcd/kustomize-controller/blob/da0b8e80123f066d364eaf3e08d85b0960e41a47/internal/sops/pgp/keysource.go#L155

I believe the reported issue was either a false positive or it has been fixed since reported. Do you mind closing this PR and updating the issue about this specific file?

internal/sops/pgp/keysource.go
tomhuang12: Thanks for the explanation. I thought it would get caught and return by this as `err` is not scoped by `func()` right?
```
if err != nil {
		return keyring, err
```

pjbgf: No worries. It would not be caught there as the defer would be the last thing to be executed before returning, so no code executes after it - with the exception of other defers.

tomhuang12: That makes sense. Thanks for bearing with me as I am new to Golang 😅

PR #537, opened 2022-01-14
Docs - Add var substitution operator escape syntax

gvillafanetapia:
Describe explicitly in docs how to escape variable substitution operator when you need to keep the dollar sign and curly braces.

Closes: https://github.com/fluxcd/website/issues/705


kingdonb (approved):
This is almost like in `Makefile`

Thanks for adding this @gvillafanetapia 👍

stefanprodan (approved):
LGTM 

Thanks @gvillafanetapia

PR #538, opened 2022-01-17
SOPS: Add support for HashiCorp Vault token-based authentication

### LINKED ISSUE #516: Support loading a secret with a HashiCorp Vault token to decrypt sops secrets.
We want to be able to pass a secret containing a `HashiCorp Vault` token to a `kustomization`. The Kustomization-controller would pass the token to the `Sops Decryptor` for decrypting secrets.

This should work like it does with age and pgp, e.g.
```yaml
kind: Kustomization
metadata:
  name: my-secrets
  namespace: default
spec:
  interval: 5m
  path: "./"
  sourceRef:
    kind: GitRepository
    name: my-secrets
  decryption:
    provider: sops
    secretRef:
      name: sops-vault-token
```

souleb:
fixes #516 

If implemented, the kustomize controller will be able to retrieve a secret containing a VAULT TOKEN and use it to decrypt the sops encrypted master key. Which will then be used to decrypt the data key that will decrypt the final data.

The kustomize-controller retrieve the secret specified in the kustomization in the same namespace. We keep the behaviour set by age and pgp decryption. 

In case a token does not exist for a HashiCorp vault encrypted token, we fall back to the default server to try decrypting/encrypting the data. This will preserve the behaviour for customer relying on VAULT_TOKEN env var.

Signed-off-by: Soule BA <bah.soule@gmail.com>


controllers/kustomization_decryptor.go
```diff
@@ -147,24 +148,31 @@ func (kd *KustomizeDecryptor) ImportKeys(ctx context.Context) error {
...
+				if err := os.WriteFile(keyPath, value, os.ModePerm); err != nil {
 					return fmt.Errorf("unable to write key to storage: %w", err)
 				}
 				if err := kd.gpgImport(keyPath); err != nil {
 					return err
 				}
 			case ".agekey":
-				ageIdentities = append(ageIdentities, string(file))
+				ageIdentities = append(ageIdentities, string(value))
+			case ".vault-token":
```
hiddeco, line 167: As `secret.Data` is a map, and the [iteration order of maps is undefined in Go](https://go.dev/blog/maps#iteration-order), this has a chance on picking a random key every time if multiple are defined.

I think this should be changed to either support multiple tokens (and collect them like the age identities above), use a fixed name (e.g. `token.vault`), or at least be stable sorted.

docs/spec/v1beta2/kustomization.md
```diff
@@ -1001,6 +1001,51 @@ spec:
       name: sops-age
 ```
 
+### HashiCorp Vault
+
+Export the `VAULT_ADDR`  and `VAULT_TOKEN` enviromnet variables to your shell,
+then use `sops` to encrypt a kubernetes secret (see [HashiCorp Vault](https://www.vaultproject.io/) for more details on enabling the transit backend and [sops](https://github.com/mozilla/sops#encrypting-using-hashicorp-vault) for using the `sops` command):
```
hiddeco, line 1007:
```suggestion
then use `sops` to encrypt a kubernetes secret (see [HashiCorp Vault](https://www.vaultproject.io/) for more details on enabling the transit backend and [sops](https://github.com/mozilla/sops#encrypting-using-hashicorp-vault)
for using the `sops` command):
```

controllers/kustomization_decryptor.go
souleb: We are expecting one key with the `vault-token` extension to be in the secret, but you are right we never know what might happen. I'll go for the fixed key name as I don't think we should accept several tokens.

controllers/kustomization_decryptor.go
```diff
@@ -45,13 +45,17 @@ import (
 	intkeyservice "github.com/fluxcd/kustomize-controller/internal/sops/keyservice"
 )
 
-const DecryptionProviderSOPS = "sops"
+const (
+	DecryptionProviderSOPS = "sops"
+	vaultTokenFileName     = "sops.vault-token"
```
hiddeco, line 50:
```suggestion
	DecryptionVaultTokenFileName = "sops.vault-token"
```

controllers/kustomization_decryptor.go
```diff
@@ -147,24 +151,34 @@ func (kd *KustomizeDecryptor) ImportKeys(ctx context.Context) error {
...
 					return fmt.Errorf("unable to write key to storage: %w", err)
 				}
 				if err := kd.gpgImport(keyPath); err != nil {
 					return err
 				}
 			case ".agekey":
-				ageIdentities = append(ageIdentities, string(file))
+				ageIdentities = append(ageIdentities, string(value))
+			case ".vault-token":
+				// make sure we have the expected value
```
hiddeco, line 171: `// Make sure we have the absolute file name` maybe?

internal/sops/hcvault/keysource.go
```diff
@@ -0,0 +1,245 @@
...
+	raw, err := cli.Logical().Write(fullPath, payload)
+	if err != nil {
+		return fmt.Errorf("the encryption to %s has failed: %w", fullPath, err)
+	}
+	if raw == nil || raw.Data == nil {
+		return fmt.Errorf("the transit backend %s is empty", fullPath)
+	}
+	encrypted, ok := raw.Data["ciphertext"]
+	if !ok {
+		return fmt.Errorf("there's not encrypted data")
```
hiddeco, line 145:
```suggestion
		return fmt.Errorf("there's no encrypted data")
```

internal/sops/hcvault/keysource.go
```diff
@@ -0,0 +1,245 @@
...
+	raw, err := cli.Logical().Write(fullPath, payload)
+	if err != nil {
+		return nil, fmt.Errorf("the decryption from %s has failed: %w", fullPath, err)
+	}
+	if raw == nil || raw.Data == nil {
+		return nil, fmt.Errorf("the transit backend %s is empty", fullPath)
+	}
+	decrypted, ok := raw.Data["plaintext"]
+	if !ok {
+		return nil, fmt.Errorf("there's no decrypted data")
```
hiddeco, line 181:
```suggestion
		return nil, fmt.Errorf("there's no decrypted data")
```

internal/sops/hcvault/keysource.go
```diff
@@ -0,0 +1,245 @@
...
+	if raw == nil || raw.Data == nil {
+		return fmt.Errorf("the transit backend %s is empty", fullPath)
+	}
+	encrypted, ok := raw.Data["ciphertext"]
+	if !ok {
+		return fmt.Errorf("there's not encrypted data")
+	}
+	encryptedKey, ok := encrypted.(string)
+	if !ok {
+		return fmt.Errorf("the ciphertext cannot be casted to string")
```
hiddeco, line 149:
```suggestion
		return fmt.Errorf("the ciphertext cannot be cast to string")
```

internal/sops/hcvault/keysource.go
```diff
@@ -0,0 +1,245 @@
...
+	if raw == nil || raw.Data == nil {
+		return nil, fmt.Errorf("the transit backend %s is empty", fullPath)
+	}
+	decrypted, ok := raw.Data["plaintext"]
+	if !ok {
+		return nil, fmt.Errorf("there's no decrypted data")
+	}
+	dataKey, ok := decrypted.(string)
+	if !ok {
+		return nil, fmt.Errorf("the plaintest cannot be casted to string")
```
hiddeco, line 185:
```suggestion
		return nil, fmt.Errorf("the plaintext cannot be cast to string")
```

controllers/kustomization_decryptor.go
souleb: Why do you want to export it?

souleb:
@hiddeco I have checked the Azure Vault PR. We should probably make the `func NewServer(prompt bool, homeDir, vaultToken string, agePrivateKeys []string) keyservice.KeyServiceServer` accept options, if we are to keep adding parameters.

docs/spec/v1beta2/kustomization.md
```diff
@@ -1001,6 +1001,53 @@ spec:
       name: sops-age
 ```
 
+### HashiCorp Vault
+
+Export the `VAULT_ADDR`  and `VAULT_TOKEN` enviromnet variables to your shell,
+then use `sops` to encrypt a kubernetes secret (see [HashiCorp Vault](https://www.vaultproject.io/docs/secrets/transit) for more details on enabling the transit backend and [sops](https://github.com/mozilla/sops#encrypting-using-hashicorp-vault)).
```
hiddeco, line 1007:
```suggestion
then use `sops` to encrypt a kubernetes secret (see [HashiCorp Vault](https://www.vaultproject.io/docs/secrets/transit)
for more details on enabling the transit backend and [sops](https://github.com/mozilla/sops#encrypting-using-hashicorp-vault)).
```

controllers/kustomization_decryptor.go
hiddeco: Because it's static value being checked on and thus arguably part of a contract.

hiddeco (approved):
No further comments from me, nice work @souleb 🙇

.github/workflows/e2e.yaml
```diff
@@ -51,6 +51,12 @@ jobs:
         uses: fluxcd/pkg/actions/kubectl@main
         with:
           version: 1.21.2
+      - name: Setup SOPS
```
stefanprodan, line 54: Please create an actions in pkg for sops based on this https://github.com/fluxcd/pkg/blob/main/actions/kustomize/action.yml

stefanprodan (changes requested):

stefanprodan (approved):
LGTM

Thanks @souleb 🏅

PR #540, opened 2022-01-19
Update development documentation

### LINKED ISSUE #468: Review/extend DEVELOPMENT documentation
> Each of the controllers should have clear guidelines on how to build and run the
> projects. This includes the sections:
> ○ How to install dependencies
> ○ How to install the controller
> ○ How to run the test suite locally
> ○ How to run the controller locally

Some of the information might already be there - let's double-check it's up to date and accurate.

yiannistri:
Fixes #468

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -4,33 +4,43 @@
 > to find out about how to contribute to Flux and how to interact with the
 > Flux Development team.
 
+## Installing required dependencies
+
+There are a number of dependencies required to be able to run the controller and its test suite locally:
+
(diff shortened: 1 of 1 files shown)


yiannistri:
I noticed that `make deploy` deploys the `source-controller` as well (same in `helm-controller`). I'm aware of the dependency but people may find this surprising. Is this something that we should call out in the docs?

stefanprodan (approved):

stefanprodan:
source-controller and its CRDs are a requirement for running KC/HC on a cluster, we could call it out in docs.

yiannistri:
Ok let me add a note for this.

Issue #542, opened 2022-01-20
kustomization resources are silently dropped if there is an typo an yaml is no valid kubernetes yaml anymore


haarchri:
kustomization resources are silently dropped if there is an typo an yaml is no valid kubernetes yaml anymore

let's say we have a repo with lot of yaml files and `kustomization.yaml` inside - and the commit id is`master/db9bb76889ce22441f2e63e2d79702b97f4e35f1`

```yaml
---
apiVersion: pkg.crossplane.io/v1
kind: Provider
metadata:
  name: crossplane-provider-aws
spec:
  package: crossplane/provider-aws:v0.23.0
  controllerConfigRef:
    name: provider-aws
```

we see the following in the logs - everything okay and expected:
```yaml
{"level":"info","ts":"2022-01-20T11:03:48.030Z","logger":"controller.kustomization","msg":"server-side apply completed","reconciler group":"kustomize.toolkit.fluxcd.io","reconciler kind":"Kustomization","name":"central-crossplane-prod","namespace":"flux-system","output":{"Cluster/default/central-crossplane-prod":"unchanged","Configuration/crossplane-sop":"unchanged","ControllerConfig/provider-aws":"unchanged","Kustomization/flux-system/sources":"unchanged","Network/default/central-crossplane-prod-network":"unchanged","Provider/crossplane-provider-aws":"unchanged","ProviderConfig/aws-provider-1234567891012":"unchanged"}}
```

with the next commit the yaml file content changed because of a typo to `apiVersion` <-> `piVersion` - and the commit id is now `master/f218296faac11060a98b6b3a9babfcaabf934553`

```yaml
---
piVersion: pkg.crossplane.io/v1
kind: Provider
metadata:
  name: crossplane-provider-aws
spec:
  package: crossplane/provider-aws:v0.23.0
  controllerConfigRef:
    name: provider-aws
```

we see the following in the logs:

```yaml
[...]

PR #544, opened 2022-01-21
Fix preflight validation

stefanprodan:
Validate that the resources built with kustomize conform to the Kubernetes API conventions before passing them to the server-side apply engine.

Fix: #542


darkowlzz (approved):
LGTM!

PR #549, opened 2022-01-26
Allow disabling cross-namespace references

stefanprodan:
Introduce the flag `--no-cross-namespace-refs` (defaults to false) for allowing cluster admins to disable cross-namespace references to sources.

When the controller is run with `--no-cross-namespace-refs=true` and a `Kustomization.spec.sourceRef.namespace` refers to a `GitRepository` or `Bucket` in a different namespace than the `Kustomization` object, the reconciliation will fail with the `AccessDenied` reason.

On access denied errors the controller logs and sends an event e.g.:

```
can't access 'GitRepository/my-other-namespace/my-source', cross-namespace references have been blocked
```

And the `Kustomization` status `Ready` condition it set to:

```yaml
status:
  conditions:
  - lastTransitionTime: "2022-01-26T07:26:48Z"
    message: "can't access 'GitRepository/my-other-namespace/my-source', cross-namespace references have been blocked"
    reason: AccessDenied
    status: "False"
    type: Ready
```

Part of: https://github.com/fluxcd/flux2/issues/2337

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -5,7 +5,7 @@ go 1.17
 require (
 	github.com/fluxcd/pkg/apis/kustomize v0.3.1
 	github.com/fluxcd/pkg/apis/meta v0.10.2
-	github.com/fluxcd/pkg/runtime v0.12.3
+	github.com/fluxcd/pkg/runtime v0.12.4
 	k8s.io/apiextensions-apiserver v0.23.1
 	k8s.io/apimachinery v0.23.1
 	sigs.k8s.io/controller-runtime v0.11.0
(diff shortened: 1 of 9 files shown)


SomtochiAma:
I have tested that the flag disables cross namespace references.
Logs from the controller:
```
{"level":"error","ts":"2022-01-26T14:29:08.655Z","logger":"controller.kustomization","msg":"AccessDenied","reconciler group":"kustomize.toolkit.fluxcd.io",
"reconciler kind":"Kustomization","name":"podinfo","namespace":"flux-system","error":"can't access 'default/podinfo', cross-namespace references have been blocked"}
```

SomtochiAma (approved):
I have tested that the flag disables cross namespace references.
Logs from the controller:
```
{"level":"error","ts":"2022-01-26T14:29:08.655Z","logger":"controller.kustomization","msg":"AccessDenied","reconciler group":"kustomize.toolkit.fluxcd.io",
"reconciler kind":"Kustomization","name":"podinfo","namespace":"flux-system","error":"can't access 'default/podinfo', cross-namespace references have been blocked"}
```

hiddeco (approved):

makkes (approved):
That's brilliant. Should we add a link to RFC 0002 somewhere for posterity?

stefanprodan:
@makkes after we implement this flag in all controllers and we release flux 0.26.0, I'll update RFC-0002 and RFC-0003.

Issue #422, opened 2021-09-06
Multi-tenancy enforcement settings


kopiczko:
I'd like to request a feature to create a setting(s) to enforce multi-tenancy.

**Motivation:**

We would like to allow our customers to use flux as GitOps engine in our Management Clusters (in CAPI terms).

Right now flux components can impersonate Service Account which is great. But without requiring:
- `spec.serviceAccountName` to be set
- referencing sources from the same namespace

This is not really effective way to enforce multi-tenancy in isolation.

**Current state:**

As described [here](https://github.com/fluxcd/flux2-multi-tenancy/#enforce-tenant-isolation) this can be achieved with validation webhooks (or intermediate engine like kyverno or OPA).

We don't feel comfortable with setting our kyverno failurePolicy to fail and we don't use it for anything security-related.

**Proposal:**

Add a setting to `kustomize-controller` and `helm-controller` to enforce setting `spec.serviceAccountName` and `metadata.namespace == spec.sourceRef.namespace`.

Correct me if I'm wrong but it doesn't seem like a big change but can enhance multi-tenancy setup quite a bit allowing secure setup without webhooks involved.

Please let me know if this is a feature you'd be willing to accept.

hiddeco:
While I can understand your desire, and would like to help find a fitting solution. I am personally opposed to adding flags that define and/or overwrite behaviors for our Custom Resources. Reason being that it results in loose definitions, and manifests that will behave differently based on the controller that takes care of them.

As we are working on expanding our cross-namespace model in a way that it is both more extensible and securely configurable, I am wondering if the planned https://github.com/fluxcd/flux2/issues/1704 doesn't cover your needs.

kopiczko:
Hi @hiddeco, thanks and I understand your concerns.

> I am wondering if the planned [fluxcd/flux2#1704](https://github.com/fluxcd/flux2/issues/1704) doesn't cover your needs.

Unfortunately I don't think it does. We don't want to predefine sources for our customers. We'd like to give them first-class flux experience with as few tweaks as possible, allowing them creating sources/kustomizations/helm charts they want while being protected with K8s RBAC rules that we have already defined. This is more flexible and lowers operational effort on our end. That's why SA impersonation feels like a great fit. We thought [multi tenancy](https://github.com/fluxcd/flux2-multi-tenancy/) setup would be ideal, but we'd like to avoid employing webhooks. Especially for enforcing security.

[...]

PR #550, opened 2022-01-27
Allow setting a default service account for impersonation

stefanprodan:
Introduce the flag `--default-service-account` for allowing cluster admins to enforce impersonation across the cluster without having to use an admission controller. 

When the flag is set to a value that's not empty string, all Kustomizations which don't have `spec.serviceAccountName` specified, they will use the service account name provided by `--default-service-account=<SA Name>` in the namespace of the object.

**Breaking changes**: 
- When `--default-service-account=<SA name>` and/or `spec.ServiceAccountName` are specified, the controller no longer queries the API to extract and use the SA token. Instead, the controller relies on the Kubernetes impersonation feature to run the reconciliation under the `system:serviceaccount:<SA Namespace>:<SA Name>` identity. This impacts only clusters with custom made tokens and `automountServiceAccountToken` disabled. If a service account token points to a different identity (which should never be the case), then the controller will fail to reconcile with `Forbidden : User "system:serviceaccount:my-namespace:my-sa-name" cannot patch resource`.
- When both `spec.kubeConfig` and `spec.ServiceAccountName` are specified, the controller will impersonate the service account on the target cluster, previously the controller ignored the service account. 

Fix: #422
Part of: https://github.com/fluxcd/flux2/issues/2340

### DIFF
--- a/controllers/kustomization_controller.go
+++ b/controllers/kustomization_controller.go
@@ -86,8 +86,10 @@ type KustomizationReconciler struct {
 	StatusPoller          *polling.StatusPoller
 	ControllerName        string
 	NoCrossNamespaceRefs  bool
+	DefaultServiceAccount string
 }
 
+// KustomizationReconcilerOptions contains options for the KustomizationReconciler.
 type KustomizationReconcilerOptions struct {
 	MaxConcurrentReconciles   int
 	HTTPRetry                 int
@@ -339,7 +341,7 @@ func (r *KustomizationReconciler) reconcile(
 	}
 
 	// setup the Kubernetes client for impersonation
-	impersonation := NewKustomizeImpersonation(kustomization, r.Client, r.StatusPoller, dirPath)
+	impersonation := NewKustomizeImpersonation(kustomization, r.Client, r.StatusPoller, r.DefaultServiceAccount)
 	kubeClient, statusPoller, err := impersonation.GetClient(ctx)
 	if err != nil {
 		return kustomizev1.KustomizationNotReady(
@@ -882,7 +884,7 @@ func (r *KustomizationReconciler) finalize(ctx context.Context, kustomization ku
 		kustomization.Status.Inventory.Entries != nil {
 		objects, _ := ListObjectsInInventory(kustomization.Status.Inventory)
 
-		impersonation := NewKustomizeImpersonation(kustomization, r.Client, r.StatusPoller, "")
+		impersonation := NewKustomizeImpersonation(kustomization, r.Client, r.StatusPoller, r.DefaultServiceAccount)
 		kubeClient, _, err := impersonation.GetClient(ctx)
 		if err != nil {
 			// when impersonation fails, log the stale objects and continue with the finalization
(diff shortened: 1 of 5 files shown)


squaremo:
> * When `--default-service-account=<SA name>` and/or `spec.ServiceAccountName` are specified, the controller no longer queries the API to extract and use the SA token. Instead, the controller relies on the Kubernetes impersonation feature to run the reconciliation under the `system:serviceaccount:<SA Namespace>:<SA Name>` identity.

I'm not clear on how this is a breaking change -- isn't the result the same?

stefanprodan:
> I'm not clear on how this is a breaking change -- isn't the result the same?

Indeed the result is the same, but the operations needed to achieve the result have drastically changed.

squaremo:
> Indeed the result is the same,

So does it break anything?

stefanprodan:
> So does it break anything?

Yes, if people have disabled the automated token creation for their service accounts, and they are setting the token in the attached secret by hand, then this PR will break their setup. As we no longer use tokens.

squaremo:
> if people have disabled the automated token creation for their service accounts, and they are setting the token in the attached secret by hand, then this PR will break their setup, as we no longer use tokens.

That's what you should say in the description (commit, docs, changelog) :-)

docs/spec/v1beta2/kustomization.md
```diff
@@ -617,11 +617,20 @@ spec:
...
 namespace, the reconciliation will fail since the account it runs under has no permissions to alter objects
 outside of the `webapp` namespace.
 
+### Enforce impersonation
+
+On multi-tenant clusters, platform admins can enforce impersonation with the
+`--default-service-account` flag.
+
+When the flag is set, all Kustomizations which don't have `spec.serviceAccountName` specified,
+they will use the service account name provided by `--default-service-account=<SA Name>`
```
hiddeco, line 631:
```suggestion
When the flag is set, all Kustomizations which don't have `spec.serviceAccountName` specified
will use the service account name provided by `--default-service-account=<SA Name>`
```

controllers/kustomization_impersonation.go
```diff
@@ -33,92 +32,58 @@ import (
...
 
 type KustomizeImpersonation struct {
-	workdir       string
-	kustomization kustomizev1.Kustomization
-	statusPoller  *polling.StatusPoller
 	client.Client
+	kustomization         kustomizev1.Kustomization
+	statusPoller          *polling.StatusPoller
+	defaultServiceAccount string
 }
```
souleb, line 40:
```suggestion
// NewKustomizeImpersonation creates a new KustomizeImpersonation.
```

hiddeco (approved):

controllers/kustomization_controller.go
```diff
@@ -86,6 +86,7 @@ type KustomizationReconciler struct {
 	StatusPoller          *polling.StatusPoller
 	ControllerName        string
 	NoCrossNamespaceRefs  bool
+	DefaultServiceAccount string
 }
```
souleb, line 91:
```suggestion
// KusomizationReconcilerOptions contains options for the KustomizationReconciler
```

stefanprodan: Thanks!

PR #561, opened 2022-02-04
add spec.suspended to Kustomization printer columns

schrej:
This adds spec.suspended to the default columns of `Kustomization`.

It's currently hard to find Kustomizations that are suspended, and I think that's a pretty common use-case when leveraging that feature.
It's currently a default column as it's pretty slim anyway, but it could also be set to only show for `-o wide`.

The PR also moves the `Status` column to the end, after `Age`, since the messages are very long in error cases, which makes the columns behind it hard to find, and breaks up the table header as well.


stefanprodan:
`flux get kustomization` does have the suspended field and the Git SHAs are trimmed. Any reason for not using Flux CLI?

schrej:
Oh, true, I forgot that `flux get` exists :grimacing:
But it's just habit to use `kubectl` as the primary method for interacting with resources, and I'd argue that it would be more convenient like this.

stefanprodan:
The suspend field is preset in almost all Flux APIs. Please open an issue in the flux2 with a proposal to change this in a consistent manner. If the maintainers of all Flux controllers are supportive of this change, then we need to open PRs everywhere and ship this as a whole in the next Flux minor release.

jflambert:
> `flux get kustomization` does have the suspended field and the Git SHAs are trimmed. Any reason for not using Flux CLI?

would help with `k9s` integration

PR #566, opened 2022-02-09
Clarify spec.path in API docs

kingdonb:
Fixes fluxcd/flux2#2357 - Docs accessibility

It's clear what was intended here if you know that Flux Kustomization has `spec.path` and GitRepository doesn't have any such a field, but that's presumptive about the user's knowledge. Let's not assume they've already learned and retained that.


stefanprodan (approved):
LGTM

Thanks @kingdonb

PR #570, opened 2022-02-13
Tolerate absence of resources in post-build substitution

### LINKED ISSUE #565: Allow ConfigMaps and Secrets in post-build substitution to be absent
Per earlier discussion in https://github.com/fluxcd/kustomize-controller/issues/558#issuecomment-1032632500, it would be convenient if users of [kustomize-controller's post-build susbstitution feature](https://fluxcd.io/docs/components/kustomize/kustomization/#variable-substitution) could opt in to allowing the referenced _ConfigMap_ or _Secret_ to be absent. At present, if [the controller can't read the referenced _ConfigMap_ or _Secret_, the reconciliation attempt fails](https://sourcegraph.com/github.com/fluxcd/kustomize-controller@110ee5b15b0b82d7745639daa130437d17311fa1/-/blob/controllers/kustomization_varsub.go?L50-L60). In some cases, especially where users use the variable expansion functions to supply a default value for an empty or missing placeholder value pulled from one of these referenced _ConfigMaps_ or _Secrets_, there's no reason to create the referenced Kubernetes object until it's time to supply such a value. When users later wish to remove the overridden value, they could then simply delete the referenced _ConfigMap_ or _Secret_, as opposed to editing its "data" field but needing to leave the object in place.

Consider adding a field to [the `api/v1beta2/SubstituteReference` struct](https://sourcegraph.com/github.com/fluxcd/kustomize-controller@110ee5b15b0b82d7745639daa130437d17311fa1/-/blob/api/v1beta2/kustomization_types.go?L196-L210). That field could be of type `bool` and named something like "Optional" or "TolerateAbsence." There's some precedent in the Kubernetes API's types such as [`core/v1.ConfigMapVolumeSource`](https://github.com/kubernetes/api/blob/67a3496afcb484d179407a5c699e216c10302a98/core/v1/types.go#L1625-L1627) and [`core/v1.ConfigMapProjection`](https://github.com/kubernetes/api/blob/67a3496afcb484d179407a5c699e216c10302a98/core/v1/types.go#L1652-L1654), along with their sibling types for _Secrets_. The default value of this field would be false, to retain the existing behavior.

Over in file _controllers/kustomization_varsub.go_ in the `substituteVariables` function, if [our call to the `(*client.Client).Get` method](https://sourcegraph.com/github.com/fluxcd/kustomize-controller@110ee5b15b0b82d7745639daa130437d17311fa1/-/blob/controllers/kustomization_varsub.go?L50) returns an error and this proposed `SubstituteReference.Optional` field is true, we'd check whether the error satisfies `k8s.io/apimachinery/pkg/api/errors.IsNotFound`. If it does, we'd skip the variable replacement and not report an error, as if the referenced _ConfigMap_ or _Secret_ existed but had an empty "Data" field.

This change would require documenting the new field in `SubstituteReference`, augmenting the `substituteVariables` function, and adding a few test cases in file _controllers/kustomize_varsub_test.go_ to exercise the case of referencing a _ConfigMap_ or _Secret_ that does not exist, but with the proposed new field set to true.

Whether to make this new field of type `*bool` like the Kubernetes API does remains an open question.

seh:
In a _Kustomization_'s post-build substitution sources, introduce a new "Optional" field to allow referencing a Kubernetes _ConfigMap_ or _Secret_ that may not exist at time of reconciliation. Treat substitution when the referenced object is missing as if the object had been present but empty, lacking any variable bindings.

Retain the longstanding behavior of interpreting references to Kubernetes objects being mandatory by default, such that reconciliation fails if such a referenced object does not exist. Only when the "Optional" field is set to true will reconciliation tolerate finding the referenced object to be missing.

Fixes #565.

### DIFF
--- a/docs/spec/v1beta2/kustomization.md
+++ b/docs/spec/v1beta2/kustomization.md
@@ -285,10 +285,10 @@ On multi-tenant clusters, platform admins can disable cross-namespace references
 
 If your repository contains plain Kubernetes manifests, the
 `kustomization.yaml` file is automatically generated for all the Kubernetes
-manifests in the `spec.path` of the Flux `Kustomization` and sub-directories.
-This expects all YAML files present under that path to be valid kubernetes
-manifests and needs non-kubernetes ones to be excluded using `.sourceignore`
-file or `spec.ignore` on `GitRepository` object.
+manifests in the directory tree specified in the `spec.path` field of the Flux `Kustomization`.
+All YAML files present under that path must be valid Kubernetes
+manifests, unless they're excluded either by way of the `.sourceignore`
+file or the `spec.ignore` field on the corresponding `GitRepository` object.
 
 Example of excluding CI workflows and SOPS config files:
 
@@ -1053,10 +1053,10 @@ spec:
 ### HashiCorp Vault
 
 Export the `VAULT_ADDR`  and `VAULT_TOKEN` environment variables to your shell,
-then use `sops` to encrypt a kubernetes secret (see [HashiCorp Vault](https://www.vaultproject.io/docs/secrets/transit)
+then use `sops` to encrypt a Kubernetes Secret (see [HashiCorp Vault](https://www.vaultproject.io/docs/secrets/transit)
 for more details on enabling the transit backend and [sops](https://github.com/mozilla/sops#encrypting-using-hashicorp-vault)).
 
-Then use `sops` to encrypt a kubernetes secret:
+Then use `sops` to encrypt a Kubernetes Secret:
 
 ```console
 $ export VAULT_ADDR=https://vault.example.com:8200
--- a/docs/spec/v1beta2/kustomization.md
+++ b/docs/spec/v1beta2/kustomization.md
@@ -748,6 +748,16 @@ With `spec.postBuild.substituteFrom` you can provide a list of ConfigMaps and Se
 from which the variables are loaded.
 The ConfigMap and Secret data keys are used as the var names.
 
+The `spec.postBuild.substituteFrom.optional` field indicates how the
+controller should handle a referenced ConfigMap or Secret being absent
+at renconciliation time. The controller's default behavior ― with
+`optional` unspecified or set to `false` ― has it fail reconciliation if
+the referenced object is missing. By setting the `optional` field to
+`true`, you can indicate that controller should use the referenced
+object if it's there, but also tolerate its absence, treating that
+absence as if the object had been present but empty, defining no
+variables.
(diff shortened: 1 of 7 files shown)


stefanprodan (commented):
Can you please update the API docs here https://github.com/fluxcd/kustomize-controller/blob/main/docs/spec/v1beta2/kustomization.md#variable-substitution and mention the Optional field and its behaviour. Please also add it to the example YAML in that section, under `cluster-vars`.

seh:
Please let me know if the documentation suits your style. I am amenable to adjusting the voice and level of detail.

docs/spec/v1beta2/kustomization.md
```diff
@@ -748,6 +748,16 @@ With `spec.postBuild.substituteFrom` you can provide a list of ConfigMaps and Se
 from which the variables are loaded.
 The ConfigMap and Secret data keys are used as the var names.
 
+The `spec.postBuild.substituteFrom.optional` field indicates how the
+controller should handle a referenced ConfigMap or Secret being absent
+at renconciliation time. The controller's default behavior―with
```
makkes, line 753:
```suggestion
at renconciliation time. The controller's default behavior ― with
```

docs/spec/v1beta2/kustomization.md
```diff
@@ -748,6 +748,16 @@ With `spec.postBuild.substituteFrom` you can provide a list of ConfigMaps and Se
...
+`optional` unspecified or set to false―has it fail reconciliation if
```
makkes, line 754:
```suggestion
`optional` unspecified or set to false ― has it fail reconciliation if
```

docs/spec/v1beta2/kustomization.md
```diff
@@ -748,6 +748,16 @@ With `spec.postBuild.substituteFrom` you can provide a list of ConfigMaps and Se
...
+the referenced object is missing. By setting the `optional` field to
+true, you can indicate that controller should use the referenced
```
makkes, line 756:
```suggestion
`true`, you indicate that controller should use the referenced
```

makkes (commented):
Great change, thanks @seh! Just a few nits, otherwise lgtm.

docs/spec/v1beta2/kustomization.md
seh: Ah, spaces around your em dashes, then? I can meet you there.

stefanprodan (approved):
LGTM

Thanks @seh 🎖️

PR #575, opened 2022-02-18
Update `pkg/runtime` and `apis/meta`

hiddeco:
This includes an update of the source-controller to v0.22.0, to pull in
the v1beta2 API which makes use of the same packages.


go.mod
```diff
@@ -208,3 +208,5 @@ replace github.com/opencontainers/runc => github.com/opencontainers/runc v1.0.3
 
 // Fix CVE-2021-41190
 replace github.com/opencontainers/image-spec => github.com/opencontainers/image-spec v1.0.2
+
+replace github.com/fluxcd/source-controller/api => github.com/fluxcd/source-controller/api v0.21.3-0.20220216150237-6804b676eb32
```
stefanprodan, line 212: I think we should delete this and set the api version to v.22.0 in the first section.

tests/fuzz/go.mod
```diff
@@ -3,3 +3,6 @@ module github.com/fluxcd/kustomize-controller/tests/fuzz
 // with fuzz dependencies.
 
 go 1.17
+
+replace github.com/fluxcd/kustomize-controller/api => ../../api
+replace github.com/fluxcd/kustomize-controller => ../../
```
stefanprodan, line 8: This shouldn't be needed, this go.mod is managed by https://github.com/fluxcd/kustomize-controller/commit/f224109d26e829fb55c251c96ce77bad6bc09f8e

stefanprodan (changes requested):

tests/fuzz/go.mod
hiddeco: We should likely apply that patch to the other controllers where we did this.

stefanprodan: Yes, as it is right now, we run the fuzzing with the `latest` version of all our dependencies, making fuzzing not accurate and error prone.

hiddeco: Reverted it, but added a note about what _actually_ manages the go.mod in the file.

stefanprodan (approved):
LGTM

PR #580, opened 2022-02-24
Use main go.mod for fuzz tests

pjbgf:
Fixes issues building fuzzers whilst also ensuring that they are linking back to current version of kustomizer-controller and ./api.


stefanprodan (approved):
LGTM

Thanks @pjbgf 🏅

PR #593, opened 2022-03-24
Add kubeconfig flags

pjbgf:
Two new flags were added to allow users to enable the use of user.Exec (`--insecure-kubeconfig-exec`) and InsecureTLS (`insecure-kubeconfig-tls`) in the kubeconfigs provided for remote apply.

Breaking change: both functionalities are no longer enabled by default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,9 @@ IMG ?= fluxcd/kustomize-controller:latest
 CRD_OPTIONS ?= crd:crdVersions=v1
 SOURCE_VER ?= v0.22.3
 
+# Use the same version of SOPS already referenced on go.mod
+SOPS_VER := $(shell go list -m all | grep go.mozilla.org/sops | awk '{print $$2}')
+
 # Repository root based on Git metadata
 REPOSITORY_ROOT := $(shell git rev-parse --show-toplevel)
 BUILD_DIR := $(REPOSITORY_ROOT)/build
@@ -14,7 +17,7 @@ GOBIN=$(BUILD_DIR)/gobin
 else
 GOBIN=$(shell go env GOBIN)
 endif
-export PATH:=${GOBIN}:${PATH}
+export PATH:=$(GOBIN):${PATH}
 
 # Allows for defining additional Go test args, e.g. '-tags integration'.
 GO_TEST_ARGS ?=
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -120,22 +125,26 @@ docker-deploy:
...
 
-ENVTEST = $(shell pwd)/bin/setup-envtest
+ENVTEST = $(GOBIN)/setup-envtest
 .PHONY: envtest
 setup-envtest: ## Download envtest-setup locally if necessary.
 	$(call go-install-tool,$(ENVTEST),sigs.k8s.io/controller-runtime/tools/setup-envtest@latest)
 
+SOPS = $(GOBIN)/sops
+$(SOPS): ## Download latest sops binary if none is found.
+	$(call go-install-tool,$(SOPS),go.mozilla.org/sops/v3/cmd/sops@v3.7.2)
```
hiddeco, line 146: Can we either get this version from the `go.mod`, or at least declare it as a var?

go.mod
```diff
@@ -30,10 +30,10 @@ require (
 	go.mozilla.org/sops/v3 v3.7.2
 	golang.org/x/net v0.0.0-20220225172249-27dd8689420f
 	google.golang.org/grpc v1.45.0
-	k8s.io/api v0.23.4
+	k8s.io/api v0.23.5
 	k8s.io/apiextensions-apiserver v0.23.4
-	k8s.io/apimachinery v0.23.4
-	k8s.io/client-go v0.23.4
+	k8s.io/apimachinery v0.23.5
+	k8s.io/client-go v0.23.5
```
stefanprodan, line 36: Please bump this in the api package, we really need to keep the Kubernetes packages in sync.

Makefile
pjbgf: Changed it to align with go mod version. PTAL

go.mod
pjbgf: Removed those changes accordingly.

stefanprodan (commented):
Please remove sops from the e2e workflow:

```
      - name: Setup SOPS
        uses: fluxcd/pkg/actions/sops@main
```

pjbgf:
> Please remove sops from the e2e workflow:

@stefanprodan  Added a new commit to this effect. PTAL

stefanprodan:
@pjbgf looks like the Makefile changes don't work, sops is not added to the PATH.

pjbgf:
@stefanprodan they have been fixed now, PTAL

stefanprodan (approved):
LGTM

Thanks @pjbgf 🏅

controllers/kustomization_controller.go
```diff
@@ -65,6 +65,7 @@ import (
 	sourcev1 "github.com/fluxcd/source-controller/api/v1beta2"
 
 	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1beta2"
+	fluxClient "github.com/fluxcd/pkg/runtime/client"
```
stefanprodan, line 68: Please move this in the fluxcd/pkg section and rename the alias to runtimeClient

stefanprodan (changes requested):

stefanprodan (approved):

PR #603, opened 2022-03-29
Add recommended settings to API docs

stefanprodan:
Add example of deploying applications from Git to production environments.

### DIFF
--- a/docs/spec/v1beta2/kustomization.md
+++ b/docs/spec/v1beta2/kustomization.md
@@ -242,6 +242,51 @@ const (
 )
 ```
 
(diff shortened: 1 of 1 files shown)


hiddeco (commented):
I think it would help if there would be some explanations as to why certain choices are made, to help the user answer certain questions as e.g. "why should I apply things with an interval of one hour?". As this could help them fine-tune their configurations.

stefanprodan:
@hiddeco I added explanations for each field, please suggest changes.

hiddeco (approved):

PR #620, opened 2022-04-15
controllers: use own Kustomize FS implementation

hiddeco:
For details, see: https://github.com/fluxcd/pkg/pull/262


stefanprodan (approved):
LGTM

Thanks @hiddeco 🥇

PR #619, opened 2022-04-13
controllers: improve decryptor and add tests

hiddeco:
- Refactored recursion while iterating over Kustomization files.
  References of files that have been visited are cached, and not
  visited again.
- Optimized various bits around (un)marshalling (encrypted) data, and
  YAML -> JSON -> YAML roundtrips are prevented where not required.
- Added support for decrypting INI Kustomize EnvSource references using
  the dedicated SOPS store for the format.
- Introduced support for decrypting Kustomize FileSources:
  https://pkg.go.dev/sigs.k8s.io/kustomize@v1.0.2/pkg/types#DataSources

PR #636, opened 2022-04-26
Use Secret generator keys for SOPS format hint

seh:
Rather than inspecting the source file name supplied to [_kustomize_'s _Secret_ generator](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/secretgenerator/) to determine the format of the SOPS-encrypted file content, instead inspect the _Secret_ key—when supplied separately from the source file name—as a more reliable heuristic.

Doing so allows _kustomization_ authors to name their SOPS-encrypted output files with a ".json" extension accurately reflecting the format in which SOPS writes its encrypted output, even if the encrypted content itself is not in JSON format.

See [preceding discussion in the "flux" channel](https://cloud-native.slack.com/archives/CLAJ40HV3/p1650920860479349) of the CNCF Slack workspace for the circuitous path I took to diagnose this change in behavior, with @hiddeco's help in pointing me to his recent #619 and suggesting the idea for the fix.

### DIFF
--- a/controllers/kustomization_decryptor_test.go
+++ b/controllers/kustomization_decryptor_test.go
@@ -695,7 +695,7 @@ func TestKustomizeDecryptor_DecryptResource(t *testing.T) {
 		},
 	}
 
-	t.Run("SOPS encrypted resource", func(t *testing.T) {
+	t.Run("SOPS-encrypted Secret resource", func(t *testing.T) {
 		g := NewWithT(t)
 
 		kus := kustomization.DeepCopy()
@@ -736,7 +736,7 @@ func TestKustomizeDecryptor_DecryptResource(t *testing.T) {
 		g.Expect(got.MarshalJSON()).To(Equal(secretData))
 	})
 
-	t.Run("SOPS encrypted binary Secret data field", func(t *testing.T) {
+	t.Run("SOPS-encrypted binary-format Secret data field", func(t *testing.T) {
 		g := NewWithT(t)
 
 		kus := kustomization.DeepCopy()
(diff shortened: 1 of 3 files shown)


controllers/suite_test.go
```diff
@@ -68,7 +67,6 @@ var (
 	k8sClient    client.Client
 	testEnv      *testenv.Environment
 	testServer   *testserver.ArtifactServer
-	testEventsH  kuberecorder.EventRecorder
```
seh, line 71: _staticcheck_ suggested that I remove this.

controllers/kustomization_decryptor_test.go
```diff
@@ -1159,7 +1169,7 @@ func TestKustomizeDecryptor_decryptSopsFile(t *testing.T) {
 
 				b, err := os.ReadFile(filepath.Join(tmpDir, f.name))
 				g.Expect(err).ToNot(HaveOccurred())
-				g.Expect(bytes.Compare(f.data, b) == 0).To(Equal(f.expectData))
+				g.Expect(bytes.Equal(f.data, b)).To(Equal(f.expectData))
```
seh, line 1172: _staticcheck_ suggested this change.

controllers/kustomization_decryptor_test.go
```diff
@@ -877,13 +882,13 @@ func TestKustomizeDecryptor_decryptKustomizationEnvSources(t *testing.T) {
 					GeneratorArgs: kustypes.GeneratorArgs{
 						Name: "envSecret",
 						KvPairSources: kustypes.KvPairSources{
-							FileSources: []string{"file.txt"},
-							EnvSources:  []string{"app.env", "key=../secret.env"},
```
seh, line 881: Note that _kustomize_ doesn't split the "EnvSources" field's values on an equals sign; it assumes that the whole value is always a file path.

hiddeco (approved):
Flawless execution of my instructions, and a big :100: for the additional nit fixes. Thanks a lot @seh :bow: :1st_place_medal:

stefanprodan (approved):
LGTM

Thanks @seh

PR #637, opened 2022-04-27
generator: recover from kustomize build panics

hiddeco:
Kustomize tends to panic in unpredicted ways due to (accidental) invalid object data; recover when this happens to ensure continuity of operations.


stefanprodan (approved):
LGTM

Thanks @hiddeco ☕

PR #638, opened 2022-04-27
Disallow remote bases usage in Kustomize overlays

stefanprodan:
Add an optional flag for disabling Kustomize [remote bases](https://github.com/kubernetes-sigs/kustomize/blob/a7f4db7fb41e17b2c826a524f545e6174b4dc6ac/examples/remoteBuild.md).

While the `--no-remote-bases` is set to `false` by default, Flux users are encouraged to enable it on production system for security and performance reasons. Using Kustomize remote bases means that kustomize-controller must clone the remote repositories on every reconciliation instead of using the source-controller artifacts cache. Allowing remote bases on multi-tenant clusters, means platform admins have no control over which repositories make up the desired state. When this flag is enabled, all resources must refer to local files included in the source artifact, meaning that only the Flux sources defined in Git can affect the cluster state.

### DIFF
--- a/controllers/kustomization_controller.go
+++ b/controllers/kustomization_controller.go
@@ -88,6 +88,7 @@ type KustomizationReconciler struct {
 	ControllerName        string
 	statusManager         string
 	NoCrossNamespaceRefs  bool
+	NoRemoteBases         bool
 	DefaultServiceAccount string
 	KubeConfigOpts        runtimeClient.KubeConfigOptions
 }
@@ -655,7 +656,7 @@ func (r *KustomizationReconciler) build(ctx context.Context, workDir string, kus
 		return nil, fmt.Errorf("error decrypting env sources: %w", err)
 	}
 
-	m, err := secureBuildKustomization(workDir, dirPath)
+	m, err := secureBuildKustomization(workDir, dirPath, !r.NoRemoteBases)
 	if err != nil {
 		return nil, fmt.Errorf("kustomize build failed: %w", err)
 	}
--- a/controllers/kustomization_generator.go
+++ b/controllers/kustomization_generator.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 	"sync"
 
+	"sigs.k8s.io/kustomize/api/filesys"
 	"sigs.k8s.io/kustomize/api/konfig"
 	"sigs.k8s.io/kustomize/api/krusty"
 	"sigs.k8s.io/kustomize/api/provider"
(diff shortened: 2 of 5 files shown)


controllers/kustomization_generator.go
```diff
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sigs.k8s.io/kustomize/api/filesys"
```
hiddeco, line 24: Should be moved to second import section so we keep things ordered.

hiddeco (approved):

controllers/kustomization_controller.go
```diff
@@ -88,6 +88,7 @@ type KustomizationReconciler struct {
 	ControllerName        string
 	statusManager         string
 	NoCrossNamespaceRefs  bool
+	NoRemoteBases         bool
```
makkes, line 91:
```suggestion
	AllowRemoteBases         bool
```
I think the kind of negative logic here might confuse users so a positive logic is more user-friendly from my point of view.

stefanprodan: > I think the kind of negative logic here might confuse users so a positive logic is more user-friendly from my point of view.

IMO the flags prefixed with `--no` tell people that a major feature that they rely on is going to be disabled. Having `allow` for something that was always available in Flux it's way more confusing.

makkes: I was actually thinking that a default of `false` for `AllowRemoteBases` was what we wanted in the service of security by default.

stefanprodan: > I was actually thinking that a default of false for AllowRemoteBases was what we wanted in the service of security by default.

Are you seriously considering breaking 99% of Flux users clusters?

main.go
```diff
@@ -88,6 +89,8 @@ func main() {
 	flag.DurationVar(&requeueDependency, "requeue-dependency", 30*time.Second, "The interval at which failing dependencies are reevaluated.")
 	flag.BoolVar(&watchAllNamespaces, "watch-all-namespaces", true,
 		"Watch for custom resources in all namespaces, if set to false it will only watch the runtime namespace.")
+	flag.BoolVar(&noRemoteBases, "no-remote-bases", false,
```
pjbgf, line 92: It may be beneficial to convey the security implications on the flag name (e.g. `--insecure-remote-bases`).

Now compounding the security and the performance impact, it would be good to consider this to be disabled by default, specially ahead of GA.
Alternatively, we could add a flag to `flux cli`, in which we always enforce the best security best practices across all controllers for a given Flux version. But that flag would not promise/honour backwards compatibility.

controllers/kustomization_controller.go
pjbgf: Also commented to this effect below. There may be a compromise between breaking users now (before GA) vs getting setting insecure defaults that will be around for the long-term. I suggested a potential approach for the security focus folks.

stefanprodan: This is not the right forum to propose something as impactful as this. Given that we didn't disabled cross-refs by default, doing so for remote bases would require a RFC where we gather feedback from the Flux community. The current implementation and naming convention follows the cross-refs one.

main.go
stefanprodan: I'm for including the `--no-remote-bases=true` into the security profile as described in [[RFC] Flux Multi-Tenancy Mode](https://github.com/fluxcd/flux2/pull/2093).

controllers/kustomization_controller.go
makkes: > > I was actually thinking that a default of false for AllowRemoteBases was what we wanted in the service of security by default.
> 
> Are you seriously considering breaking 99% of Flux users clusters?

😆 No, that was not my intention. I kind of thought about the flag in isolation and didn't think about the upgrade flow potentially breaking things w/o proper supervision (i.e. setting the flag accordingly to retain existing behaviour).

In that case it looks like we don't have much of a choice now.

main.go
pjbgf: Let's add this feature as is, and discuss on a separate forum about how to enable security focused users to adopt breaking changes more seamlessly - as well as the MT topics.

pjbgf (approved):
LGTM

PR #640, opened 2022-04-27
Fix field-manager name in docs

aliusmiles:
I assume that `flux-client-side-apply` value is leftover from client-side apply times?

Changed to `kustomize-controller` as this is the manager name showed by `kubectl describe` on Flux-managed resources.


aliusmiles:
Aha, thanks for clarification!

PR #644, opened 2022-04-29
decryptor: detect DockerConfigJsonKey as JSON out

hiddeco:
This ensures the Secret field gets formatted back into JSON, instead of
it being detected as binary output.


controllers/kustomization_decryptor.go
```diff
@@ -406,13 +409,13 @@ func (d *KustomizeDecryptor) decryptKustomizationEnvSources(visited map[string]s
...
-				if err := visitRef(filePath, formats.FormatForPath(key)); err != nil {
+				if err := visitRef(filePath, formatForPath(key)); err != nil {
 					return err
 				}
 			}
 			for _, envFile := range gen.EnvSources {
-				format := formats.FormatForPath(envFile)
-				if formats.FormatForPath(envFile) == formats.Binary {
+				format := formatForPath(envFile)
+				if formatForPath(envFile) == formats.Binary {
```
seh, line 418: It looks like both the previous statement and the new statement both repeat the call to determine the format for the "envFile" variable.
```suggestion
				if format == formats.Binary {
```

controllers/kustomization_decryptor_test.go
```diff
@@ -806,6 +806,59 @@ func TestKustomizeDecryptor_DecryptResource(t *testing.T) {
...
+		g.Expect(err).ToNot(HaveOccurred())
+
+		secret := resourceFactory.FromMap(map[string]interface{}{
+			"apiVersion": "v1",
+			"kind":       "Secret",
+			"metadata": map[string]interface{}{
+				"name":      "secret",
+				"namespace": "test",
+			},
+			"type": corev1.SecretTypeDockercfg,
```
seh, line 849:
```suggestion
			"type": corev1.SecretTypeDockerConfigJson,
```

controllers/kustomization_decryptor.go
```diff
@@ -731,3 +734,21 @@ func securePathErr(root string, err error) error {
 	}
 	return err
 }
+
+func formatForPath(path string) formats.Format {
+	switch {
+	case strings.HasSuffix(path, corev1.DockerConfigJsonKey):
+		return formats.Json
```
seh, line 741: Perhaps we should only look for this key when the _Secret_'s "type" field is "kubernetes.io/dockerconfigjson."

hiddeco: Think that with the current approach, it gives people some flexibility to e.g. have opaque Secrets with multiple `.dockerconfigjson` files, for whatever reason that may be required. In addition to simplifying the code :-).

controllers/kustomization_decryptor.go
```diff
@@ -92,6 +92,9 @@ var (
 		formats.Json:   []byte("\"mac\": \"ENC["),
 		formats.Yaml:   []byte("mac: ENC["),
 	}
+	// unsupportedFormat is used to signal no sopsFormatToMarkerBytes format was
+	// detected by detectFormatFromMarkerBytes.
+	unsupportedFormat = formats.Format(-1)
```
seh, line 97: Could this be a constant instead of a variable?

seh (approved):
Modulo the switch from `var` to `const`, this would have helped me when encrypting my Docker credentials. I used _sops --encrypt --input-type=json_ first, but found that Flux couldn't interpret the decrypted content correctly. With this patch, it looks like that would have worked as expected.

controllers/kustomization_decryptor.go
hiddeco: Yes, `maxEncryptedFileSize` as well :-)

PR #641, opened 2022-04-27
Support AWS KMS credentials using decryption secretRef

aryan9600:
Adds support for using `.spec.decryption` to refer to custom AWS KMS credentials and override global decryption configuration.

Signed-off-by: Sanskar Jaiswal <sanskar.jaiswal@weave.works>

PR #667, opened 2022-05-27
sops/awskms: Fix AWS KMS config creation

aryan9600:
This PR is a follow up to #641 which fixes a few things around AWS KMS config creation and expands related tests.
Signed-off-by: Sanskar Jaiswal <sanskar.jaiswal@weave.works>

### DIFF
--- a/internal/sops/awskms/keysource.go
+++ b/internal/sops/awskms/keysource.go
@@ -35,6 +35,8 @@ const (
 	stsSessionRegex = "[^a-zA-Z0-9=,.@-_]+"
 	// kmsTTL is the duration after which a MasterKey requires rotation.
 	kmsTTL = time.Hour * 24 * 30 * 6
+	// roleSessionNameLengthLimit is the AWS role session name length limit.
+	roleSessionNameLengthLimit = 64
 )
 
 // MasterKey is an AWS KMS key used to encrypt and decrypt sops' data key.
@@ -53,6 +55,8 @@ type MasterKey struct {
 	// EncryptionContext provides additional context about the data key.
 	// Ref: https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#encrypt_context
 	EncryptionContext map[string]string
+	// AWSProfile is the profile to use for loading configuration and credentials.
+	AwsProfile string
 
 	// credentialsProvider is used to configure the AWS config with the
 	// necessary credentials.
@@ -62,7 +66,7 @@ type MasterKey struct {
 	// to by default. This is mostly used for testing purposes as it can not be
 	// injected using e.g. an environment variable. The field is not publicly
 	// exposed, nor configurable.
-	epResolver aws.EndpointResolver
+	epResolver aws.EndpointResolverWithOptions
 }
 
 // CredsProvider is a wrapper around aws.CredentialsProvider used for authenticating
@@ -119,8 +123,9 @@ func (key *MasterKey) Encrypt(dataKey []byte) error {
 	}
 	client := kms.NewFromConfig(*cfg)
 	input := &kms.EncryptInput{
-		KeyId:     &key.Arn,
-		Plaintext: dataKey,
+		KeyId:             &key.Arn,
+		Plaintext:         dataKey,
+		EncryptionContext: key.EncryptionContext,
 	}
 	out, err := client.Encrypt(context.TODO(), input)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/sops/awskms/keysource.go
```diff
@@ -224,15 +227,25 @@ func NewMasterKeyFromArn(arn string, context map[string]string, awsProfile strin
...
 func (key MasterKey) createKMSConfig() (*aws.Config, error) {
-	// Use the credentialsProvider if present, otherwise default to reading credentials
-	// from the environment.
+	re, err := regexp.Compile(arnRegex)
+	if err != nil {
+		return nil, fmt.Errorf("failed to compile ARN regex: %w", err)
+	}
+	matches := re.FindStringSubmatch(key.Arn)
+	if matches == nil {
+		return nil, fmt.Errorf("No valid ARN found in '%s'", key.Arn)
```
hiddeco, line 236:
```suggestion
	return nil, fmt.Errorf("no valid ARN found in '%s'", key.Arn)
```

internal/sops/awskms/keysource.go
```diff
@@ -224,15 +227,25 @@ func NewMasterKeyFromArn(arn string, context map[string]string, awsProfile strin
 
 // createKMSConfig returns a Config configured with the appropriate credentials.
 func (key MasterKey) createKMSConfig() (*aws.Config, error) {
-	// Use the credentialsProvider if present, otherwise default to reading credentials
-	// from the environment.
+	re, err := regexp.Compile(arnRegex)
```
hiddeco, line 230: `MustCompile` is probably a better option.

internal/sops/awskms/keysource_test.go
```diff
@@ -282,56 +282,84 @@ aws_session_token: test-token
...
+		{
+			name: "master key with with proper configuration passes",
+			key: MasterKey{
+				credentialsProvider: credentials.NewStaticCredentialsProvider("test-id", "test-secret", "test-token"),
+				AwsProfile:          "test-profile",
+				Arn:                 "arn:aws:kms:us-west-2:107501996527:key/612d5f0p-p1l3-45e6-aca6-a5b005693a48",
+			},
+			assertFunc: func(g *WithT, cfg *aws.Config, err error) {
+				g.Expect(err).ToNot(HaveOccurred())
+				// g.Expect(cfg.Region).To(Equal("us-west-2"))
```
hiddeco, line 309: Left over?

hiddeco (commented):
Out of interest: any chance you can link me to the doc ref around profiles?

internal/sops/awskms/keysource.go
aryan9600: IMO we shouldn't call a function that can panic, even if the argument is a constant.

hiddeco: The constant is expected to be valid, panic'ing here is thus absolutely justified?

aryan9600: Yes fair point, I try to avoid calling functions that can panic, when they have a non-panic alternative, but I guess it's okay to use it here.

internal/sops/awskms/keysource.go
```diff
@@ -250,12 +267,13 @@ func (key MasterKey) createSTSConfig(config *aws.Config) (*aws.Config, error) {
...
-	stsRoleSessionNameRe, err := regexp.Compile(stsSessionRegex)
-	if err != nil {
-		return nil, fmt.Errorf("failed to compile STS role session name regex: %w", err)
-	}
+	stsRoleSessionNameRe := regexp.MustCompile(stsSessionRegex)
+
 	sanitizedHostname := stsRoleSessionNameRe.ReplaceAllString(hostname, "")
 	name := "sops@" + sanitizedHostname
+	if len(name) >= roleSessionNameLengthLimit {
+		name = name[:roleSessionNameLengthLimit]
```
hiddeco, line 275: This would not handle multi-byte strings correctly, but I am not sure how realistic this scenario is.

aryan9600: TBH, I don't think the AWS SDK would even accept a multi-byte string as a valid role name, so I don't think we need to worry about that while asserting length.

hiddeco: Just noticed this matches current upstream, so lets leave it.

hiddeco (approved):
LGTM, brilliant @aryan9600 :bow:

Issue #686, opened 2022-06-28
Healthcheck on Job ignores job status


kcighon:
Healthcheck of Kind: job on Kustomization ignores job status and moves to Ready True Applied.

```shell
► checking prerequisites
✗ flux 0.31.1 <0.31.2 (new version is available, please upgrade)
✔ Kubernetes 1.23.6+k3s1 >=1.20.6-0
► checking controllers
✔ helm-controller: deployment ready
► ghcr.io/fluxcd/helm-controller:v0.22.1
✔ kustomize-controller: deployment ready
► ghcr.io/fluxcd/kustomize-controller:v0.26.1
✔ notification-controller: deployment ready
► ghcr.io/fluxcd/notification-controller:v0.24.0
✔ source-controller: deployment ready
► ghcr.io/fluxcd/source-controller:v0.25.5
✔ all checks passed
```

- test job status
```shell
Every 2.0s: kubectl get job -n flux-system                   MYSERVER: Tue Jun 28 09:53:24 2022

NAME      COMPLETIONS   DURATION   AGE
testjob   0/1           5m53s      5m53s

```
- flux get ks testks
``` shell
NAMESPACE       NAME                                    REVISION                        SUSPENDED       READY   MESSAGE
flux-system     kustomization/testks                  main/6335077                  False           True    Applied revision: main/6335077 
```
- however test job still erroring
```shell
Every 2.0s: kubectl get pods -n flux-system                  MYSERVER: Tue Jun 28 09:52:36 2022

NAME                                      READY   STATUS    RESTARTS   AGE
helm-controller-9bb4968f9-zbq5f           1/1     Running   0          7m42s
kustomize-controller-655c6969df-wrdrq     1/1     Running   0          7m42s
source-controller-66d8b45dfb-nqftz        1/1     Running   0          7m42s
notification-controller-b8cd574d6-wg579   1/1     Running   0          7m42s
testjob-pgcnj                             0/1     Error     0          5m5s
testjob-vkcpv                             0/1     Error     0          4m54s
testjob-nm7wt                             0/1     Error     0          4m43s
testjob-2nrfp                             0/1     Error     0          4m32s
testjob-kj22j                             0/1     Error     0          4m20s

```
- testks.yaml
```yaml
---
apiVersion: kustomize.toolkit.fluxcd.io/v1beta2
kind: Kustomization
metadata:
  name: testks
  namespace: flux-system
spec:
  force: true
  interval: 60m0s
  healthChecks:
[...]

PR #687, opened 2022-06-28
Fix job wait by adding polling options to impersonation client

somtochiama:
Fixes: #686 

Use polling options (such as the internal custom job statusreaders) when creating a new status poller.


stefanprodan (approved):
LGTM

Thanks @somtochiama 🥇

PR #689, opened 2022-06-30
Retry downloading artifacts on not found errors

stefanprodan:
Changes:
- Extract the artifact operations such as download, verify and untar into a dedicated struct
- Introduce a dedicated type for artifact not found errors
- On artifact not found errors, log the error, update the ready status message and requeue the object
- Retry the artifact download at the interval set with `--requeue-dependency` (defaults to 30s)

This PR gracefully handles Flux upgrades, where instead of erroring out while source-controller rebuilds the artifacts cache, kustomize-contoller retries the fetch operations. Ref: https://github.com/fluxcd/flux2/issues/2884

### DIFF
--- a/controllers/kustomization_controller.go
+++ b/controllers/kustomization_controller.go
@@ -19,19 +19,13 @@ package controllers
 import (
 	"bytes"
 	"context"
-	"crypto/sha1"
-	"crypto/sha256"
 	"fmt"
-	"io"
-	"net/http"
-	"net/url"
 	"os"
 	"sort"
 	"strings"
 	"time"
 
 	securejoin "github.com/cyphar/filepath-securejoin"
-	"github.com/hashicorp/go-retryablehttp"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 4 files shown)


controllers/kustomization_downloader.go
```diff
@@ -0,0 +1,127 @@
...
+	httpClient.Logger = nil
+
+	return &ArtifactDownloader{httpClient: httpClient}
+}
+
+// Fetch downloads, verifies and extracts the artifact content to the specified directory.
+// If the artifact server responds with 5xx errors, the download operation is retried.
+// If the artifact server responds with 404, the returned error is of type ArtifactNotFoundError.
+// If the artifact server is unavailable for more than 3 minutes, the returned error contains the original status code.
+func (r *ArtifactDownloader) Fetch(artifact *sourcev1.Artifact, dir string) error {
```
hiddeco, line 61: I would either name this `Download`, or call the struct an `ArtifactFetcher`.

hiddeco (approved):
Other than the naming nit, this seems to be doing what is described. Thanks @stefanprodan :1st_place_medal:

controllers/kustomization_downloader.go
stefanprodan: Renamed to `ArtifactFetcher`

PR #690, opened 2022-06-30
Add example of how to use integer values in `postBuild.substitute`

stefanprodan:
Fix: https://github.com/fluxcd/flux2/issues/1839

### DIFF
--- a/docs/spec/v1beta2/kustomization.md
+++ b/docs/spec/v1beta2/kustomization.md
@@ -738,6 +738,31 @@ which will print out `${var}`.
 All the undefined variables in the format `${var}` will be substituted with string empty, unless a default 
 is provided e.g. `${var:=default}`.
 
+When using integers as values, for example:
+
+```yaml
+apiVersion: kustomize.toolkit.fluxcd.io/v1beta2
+kind: Kustomization
+metadata:
+  name: apps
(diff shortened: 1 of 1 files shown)


docs/spec/v1beta2/kustomization.md
```diff
@@ -738,6 +738,31 @@ which will print out `${var}`.
...
+metadata:
+  name: apps
+spec:
+  ...
+  postBuild:
+    substitute:
+      AWS_ACCOUNT_ID: "71111111117"
+```
+
+For the integer to be replaced as string, the variable needs to be escaped like so:
```
hiddeco, line 755:
```suggestion
For this integer to be replaced as string, the variable needs to be escaped:
```

stefanprodan:
Well this doesn't actually work, on my test cluster the integer is double quoted `account: '"12345"'` instead of `account: "12345"` . Not sure why people are saying this works cc @kingdonb

kingdonb:
It definitely worked back in September 2021. I think it stopped working around 0.18

I have seen a few threads since then where I suggested this as the resolution, and the reply has consistently been this does not work anymore, there's no workaround at present.

bogaertg:
any workaround to make it works again ?

PR #691, opened 2022-07-01
decryptor: recover from SOPS store panic

hiddeco:
Based on user reports, there seems to be a small chance for the
underlying SOPS store implementation to panic when a user provides input
and/or output format instructions which do not actually match the type
of the file. Recover from this to ensure continuity of operations.


hiddeco:
I have been unable to come up with a test case which actually reproduces the issue, due to not having a clear view of the environment of the user which reported the issue. But this catches things at the right point for any panic that may occur while working with the SOPS store, which should resolve things based on the stack trace that was shared.

stefanprodan (approved):
LGTM

Thanks @hiddeco

PR #700, opened 2022-07-19
Fix typo from pune to prune

anbotero:
Very small typo


stefanprodan (approved):
LGTM

Thanks @anbotero for the fix

PR #706, opened 2022-08-09
Add file path to sops decryption errors

stefanprodan:
If one of the `kustomization.yaml` files can't be unmarshal by the SOPS decryptor, users can't tell from the error message which file was at fault e.g.
```
error decrypting env sources: failed to unmarshal kustomization file: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go struct field Kustomization.patches of type types.Patch
```

This PR adds the file path to the error message.


darkowlzz (approved):
LGTM!

PR #710, opened 2022-08-17
Allow reordering of Kustomize resources

alexandermarston:
The Kustomize CLI command allows you to specify whether you want to use the 'legacy' sorting system of resources or none, this PR introduces this option into the FluxCD kustomize controller.


stefanprodan:
This has no effect on the cluster, we don't use the Kustomize sorting when doing the server-side apply. The apply logic is dictated by https://github.com/fluxcd/pkg/tree/main/ssa

PR #723, opened 2022-09-06
fuzz: Refactor Fuzzers based on Go native fuzzing

pjbgf:
The existing fuzzers are converted into the Go native format. This
works well with most fuzzers, apart from the one that fuzzes the
controller.

That specific test depends on funcs and vars that are defined on
`suite_test.go`, which is not supported when building fuzzers leveraging
Go native fuzz for oss-fuzz. That test is therefore ignored when
executed with `make fuzz-native`. But, all tests are covered with
`make fuzz-smoketest`, in order to support oss-fuzz.

Once the controller is rewritten we can revist this and ensure that
fuzzers does not require envtest nor embedded CRDs.

Relates to https://github.com/fluxcd/flux2/issues/2417.


stefanprodan (approved):

pjbgf:
The build failure is orthogonal to the changes but did not have the time to get to the bottom of it:
```
=== CONT  TestKustomizationReconciler_DependsOn
    kustomization_dependson_test.go:185:
        Timed out after 30.001s.
        Expected
            <bool>: false
        to be true
=== CONT  TestKustomizationReconciler_DependsOn/reconciles_when_source_is_found
    testing.go:1336: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
--- FAIL: TestKustomizationReconciler_DependsOn (31.01s)
    --- PASS: TestKustomizationReconciler_DependsOn/fails_due_to_source_not_found (0.00s)
    --- FAIL: TestKustomizationReconciler_DependsOn/reconciles_when_source_is_found (30.00s)
```

Main is [broken](https://github.com/fluxcd/kustomize-controller/actions/runs/2983383370/attempts/2), although the same commit at the time [worked](https://github.com/fluxcd/kustomize-controller/actions/runs/2983383370/attempts/1).

stefanprodan:
It's due to using deprecated APIs in tests, HPA v2beta1 was removed from Kubernetes 1.25 and we're running the tests on latest stable. I've fixed it in main now.

PR #726, opened 2022-09-10
Sort SOPS masterkeys so offline decrypt methods are tried first

bkreitch:
This is to sort master keys so PGP/age keys appear before the others and being used first. Thus if one of local methods is successful we avoid going to remote KMS/HCVault which may not be accessible from specific network and cause unnecessary timeouts or have some other cost.

Originally I've submitted [PR to SOPS project ](https://github.com/mozilla/sops/pull/1121) as solution for [this issue](https://github.com/mozilla/sops/issues/305), but as it seems it could be a while till that PR is reviewed and new version is released. And since kustomize-controller has some internal SOPS implementation same thing could be achieved before passing data to SOPS.

### DIFF
--- a/controllers/kustomization_decryptor.go
+++ b/controllers/kustomization_decryptor.go
@@ -25,6 +25,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


aryan9600:
I think it'd be better to get this merged to upstream SOPS rather than here, we have been trying to get rid of all our custom SOPS code by contributing it back upstream. cc: @hiddeco

stefanprodan:
I'm for merging this, it could take months before the next SOPS release.

hiddeco (approved):

PR #731, opened 2022-09-20
api: add custom validation for v1.Duration types

stefanprodan:
To solve discrepancies between parsing versus validation.

xref: https://github.com/kubernetes/apimachinery/issues/131

### DIFF
--- a/api/v1beta2/kustomization_types.go
+++ b/api/v1beta2/kustomization_types.go
@@ -47,12 +47,16 @@ type KustomizationSpec struct {
 	Decryption *Decryption `json:"decryption,omitempty"`
 
 	// The interval at which to reconcile the Kustomization.
+	// +kubebuilder:validation:Type=string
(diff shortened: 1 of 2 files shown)


api/v1beta2/kustomization_types.go
```diff
@@ -130,6 +134,8 @@ type KustomizationSpec struct {
 
 	// Timeout for validation, apply and health checking operations.
 	// Defaults to 'Interval' duration.
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Pattern="^([0-9]+(\\.[0-9]+)?(ms|s|m|h))+$"
```
hiddeco, line 138: Would omit `h` for reasons specified in https://github.com/fluxcd/source-controller/pull/903#discussion_r975027656

hiddeco (approved):

PR #738, opened 2022-10-04
Support alternative kustomization file names

### LINKED ISSUE #737: Kustomization files with name other than the default `kustomization.yaml` aren't processed properly
When using the file name `kustomization.yml` or `Kustomization`, k-c doesn't recognize these files even though it should. Instead it logs an error like here:

```
open /tmp/kustomization-1539084115/src/xx/xx/xx/kustomization.yaml: no such file or directory
```

**Workaround:**

Renaming the file to include the `a` so that it becomes `kustomization.yaml` works around this issue.

makkes:
`KustomizeGenerator.WriteFile` now detects alternative kustomization
file names such as `kustomization.yml` and `Kustomization`.

closes #737

### DIFF
--- a/controllers/kustomization_generator.go
+++ b/controllers/kustomization_generator.go
@@ -24,12 +24,12 @@ import (
 	"strings"
 	"sync"
 
-	"sigs.k8s.io/kustomize/api/filesys"
 	"sigs.k8s.io/kustomize/api/konfig"
 	"sigs.k8s.io/kustomize/api/krusty"
 	"sigs.k8s.io/kustomize/api/provider"
 	"sigs.k8s.io/kustomize/api/resmap"
 	kustypes "sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
 	"sigs.k8s.io/yaml"
 
 	"github.com/fluxcd/pkg/apis/kustomize"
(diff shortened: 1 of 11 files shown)


controllers/kustomization_generator_test.go
```diff
@@ -57,3 +61,33 @@ func Test_secureBuildKustomization_rel_basedir(t *testing.T) {
 	_, err := secureBuildKustomization("testdata/relbase", "testdata/relbase/clusters/staging/flux-system", false)
 	g.Expect(err).ToNot(HaveOccurred())
 }
+
+func TestGeneratorWriteFile(t *testing.T) {
```
hiddeco, line 65: As a follow up / extension, it would probably be good to also check if an existing Kustomize file is still overwritten with the patches from the spec of the object.

makkes: done

controllers/kustomization_generator.go
```diff
@@ -24,7 +24,7 @@ import (
 	"strings"
 	"sync"
 
-	"sigs.k8s.io/kustomize/api/filesys"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
```
hiddeco, line 27: This change needs to be properly sorted and requires a `make tidy`.

hiddeco (commented):
Changes look good to me, but think smashing (especially the last commit) would be better to keep main free from noise.

makkes:
> Changes look good to me, but think smashing (especially the last commit) would be better to keep main free from noise.

Sure thing, was planning to do that, anyway. Just wanted to simplify review. Will squash now.

hiddeco (approved):

PR #739, opened 2022-10-06
Refactor: Extract decrytor to internal package

stefanprodan:
Move the decryption logic and related unit tests from `controllers` to `internal/decrytor`.


pjbgf (approved):
LGTM

PR #749, opened 2022-10-21
api: allow configuration of `h` unit for timeouts

hiddeco:
To allow waiting for `HelmRelease` resources with an exceptionally high timeout of `1h` or more.

Related to: https://github.com/fluxcd/helm-controller/pull/549

### DIFF
--- a/api/v1beta2/kustomization_types.go
+++ b/api/v1beta2/kustomization_types.go
@@ -134,7 +134,7 @@ type KustomizationSpec struct {
 	// Timeout for validation, apply and health checking operations.
 	// Defaults to 'Interval' duration.
 	// +kubebuilder:validation:Type=string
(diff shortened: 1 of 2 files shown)


hiddeco:
@stefanprodan wondering if in a future patch it would be better to create separate values for some of the timeouts, as my assumption is that a validation and/or apply timeout should not take more than a couple minutes. Where as a health check might take longer.

In extend to this, I am wondering if it would not be better to take note of the start timestamp a health checks start, to then repetitively run health checks while re-queuing until they are healthy and/or it runs into the timeout compared to the timestamp. This would improve continuity.

stefanprodan (approved):

stefanprodan:
Having a separate timeout for the health check makes sense, maybe `waitTimeout` or something like that.

PR #754, opened 2022-10-28
Add support for Kustomize components

### LINKED ISSUE #753: Add support for Kustomize components
We are using Flux's `Kustomization` for its `dependsOn` functionality to install the workloads in a specific order, ex: `Kyverno` first, then CRDs (`ServiceMonitors`, `ExternalSecrets` etc.) and last team applications. It works very well.

Kustomize supports [components](https://kubectl.docs.kubernetes.io/guides/config_management/components/) and part of the motivation is (described in the [KEP](https://github.com/kubernetes/enhancements/tree/f905a70adc9f9f629f77830a9c8c58a8830939df/keps/sig-cli/1802-kustomize-components#motivation)):

> The problem is that modular applications cannot always be expressed in a tall hierarchy while preserving all combinations of available features. Doing so would require putting each feature in an overlay, and making overlays for independent features inherit from each other. However, this is semantically incorrect, cannot not scale as the number of features grows, and soon results in duplicate manifests and kustomizations.
>
> Instead, such applications are much better expressed as a collection of components, i.e., reusable pieces of configuration logic that are defined in a common place and that distinct overlays can then mix-and-match. This approach abides by the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle and increases ease of maintenance.

This is a very useful feature and what we need for our use-case. Unfortunately, it is bit hard to use with Flux's `Kustomization`.

In our case we have a few environments (clusters):
```
.
├── environments
│   ├── base
│   │   ├── flux-system
│   │   │   ├── gotk-components.yaml
│   │   │   ├── gotk-sync.yaml
│   │   │   └── kustomization.yaml
│   │   ├── infrastructure-sources.yaml
│   │   ├── kustomization.yaml
│   │   └── monitoring-kustomization.yaml
│   ├── dev
│   │   ├── flux-system
│   │   │   └── gotk-sync.yaml
│   │   └── kustomization.yaml
│   └── prod
│       ├── flux-system
│       │   └── gotk-sync.yaml
│       └── kustomization.yaml
```

<details>
  <summary>The rest of the structure</summary>

```
└── infrastructure
    ├── monitoring
    │   ├── cortex
    │   │   ├── bkt-claim.yaml
    │   │   ├── cortex-hr.yaml
    │   │   └── kustomization.yaml
    │   ├── grafana
    │   │   ├── grafana-hr.yaml
    │   │   └── kustomization.yaml
    │   ├── kustomization.yaml
    │   ├── kube-prometheus-stack
    │   │   ├── kube-prometheus-stack-hr.yaml
    │   │   ├── kustomization.yaml
    │   │   └── namespace.yaml
    │   └── loki
    │       ├── bkt-claim.yaml
    │       ├── kustomization.yaml
    │       ├── loki-hr.yaml
    │       └── promtail-hr.yaml
    └── sources
        ├── cortex-project.yaml
        ├── grafana-helm-charts.yaml
        ├── kustomization.yaml
        ├── prometheus-community.yaml
```
</details>

We want to enable some optional features for Grafana in the dev environment, let's say OAuth and add a ingress.

If Flux's `Kustomization` supported components that would be very easy to implement, I suppose it could look like this:

```yaml
kind: Kustomization
metadata:
  name: monitoring
  namespace: flux-system
spec:
  components:
    - ../../components/grafana-oauth
    - ../../components/grafana-ingress
```
(preferable with `x-kubernetes-patch-merge-key` and `x-kubernetes-patch-strategy` for the `components` so it can be strategic merged)

Today the only option is patching the Flux `Kustomization` and creating a new overlay:

`environments/dev/monitoring-kustomization.yaml`
```yaml
kind: Kustomization
metadata:
  name: monitoring
  namespace: flux-system
spec:
  path: ./components-hack/monitoring-dev
```

`components-hack/monitoring-dev/kustomization.yaml`
```yaml
kind: Kustomization
resources:
  - ../../infrastructure/monitoring
components:
    - ../../components/grafana-oauth
    - ../../components/grafana-ingress
```

This is a bit cumbersome and adds unnecessary complexity. If Flux's `Kustomization` supported `components` we could avoid all this.

klausenbusk:
Fix #753

This has to be merged after: https://github.com/fluxcd/pkg/pull/412.


klausenbusk:
@stefanprodan can you have a look? :)

internal/generator/generator.go
```diff
@@ -79,6 +79,8 @@ func (kg *KustomizeGenerator) WriteFile(dirPath string) (string, error) {
 		})
 	}
 
+	kus.Components = kg.kustomization.Spec.Components
```
stefanprodan, line 82: What if components are already set in `kustomization.yaml`? Like with `Images` we need to merge the two arrays and add only the items which are not in original config.

stefanprodan: Please add tests for both use-cases.

klausenbusk: Oh, I missed that. I have added a test.

internal/generator/testdata/components/components/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - configmap.yaml
+components:
+  - componentA
```
klausenbusk, line 7: I dunno if this directory should be committed? It isn't required for the test nor is `configmap.yaml`, but both is required if you want to run kustomize in the directory (e.g. `kubectl kustomize`).

api/v1beta2/kustomization_types.go
```diff
@@ -153,6 +153,11 @@ type KustomizationSpec struct {
 	// +kubebuilder:validation:Enum=none;client;server
 	// +optional
 	Validation string `json:"validation,omitempty"`
+
+	// Components specifies relative paths to specifications of other Components
+	// via relative paths, absolute paths, or URLs.
```
stefanprodan, line 158: Absolute paths and URLs shouldn't be allowed here

internal/generator/testdata/components/components/kustomization.yaml
stefanprodan: If this not used in tests then why commit it?

klausenbusk: I have removed the files not needed for the test.

api/v1beta2/kustomization_types.go
klausenbusk: I have removed it from the description, but nothing prevents the user from doing it anyways (assuming `--no-remote-bases=false`).

stefanprodan: We need to block this, so far all the Kustomize features exposed in a Flux Kustomization have this restriction, e.g. you can't refer to external patches they must be inlined.

stefanprodan:
Given that Kustomize components are v1alpha1 and they can be dropped at any time from upstream, we need to tell users not to rely on this feature in production. We also need to mark the field as experimental in Flux API via a comment. Please add a section [here](https://github.com/fluxcd/kustomize-controller/blob/main/docs/spec/v1beta2/kustomization.md#override-kustomize-config) and document the components field and its restrictions. Thanks!

api/v1beta2/kustomization_types.go
klausenbusk: I don't see how this is easily doable. According to [kustomize's remote targets documentation](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md) a remote target can look like:
- `https://github.com/kubernetes-sigs/kustomize//examples/multibases/dev/?ref=v1.0.6`
- `git@github.com:owner/repo`
- `ssh://git@github.com/owner/repo`

Kustomize has [a parser](https://github.com/kubernetes-sigs/kustomize/blob/84bd402cc0662c5df3f109c4f80c22611243c5f9/api/internal/git/repospec.go#L81-L100) for this, but I can't use it, as it is a `internal` package.

klausenbusk: After a bit more digging I found the [host parser logic](https://github.com/kubernetes-sigs/kustomize/blob/84bd402cc0662c5df3f109c4f80c22611243c5f9/api/internal/git/repospec.go#L231-L239) and copied it with `file://` removed.

Looks good? :)

stefanprodan: Yeap, thanks!

internal/generator/generator.go
```diff
@@ -234,3 +241,21 @@ func adaptSelector(selector *kustomize.Selector) (output *kustypes.Selector) {
...
+	for _, p := range []string{
+		// Order matters here.
+		"git::", "gh:", "ssh://", "https://", "http://",
+		"git@", "github.com:", "github.com/"} {
+		if len(p) < len(path) && strings.ToLower(path[:len(p)]) == p {
+			return false
+		}
+	}
+
+	if filepath.IsAbs(path) || filepath.IsAbs(strings.TrimPrefix(strings.ToLower(path), "file://")) {
```
klausenbusk, line 257: I think this is unneeded after looking at bit more at [the code](https://github.com/kubernetes-sigs/kustomize/blob/84bd402cc0662c5df3f109c4f80c22611243c5f9/api/internal/git/repospec.go#L130).
```suggestion
	if filepath.IsAbs(path) {
```

stefanprodan (commented):
Please add components to the docs here https://github.com/fluxcd/kustomize-controller/blob/main/docs/spec/v1beta2/kustomization.md#override-kustomize-config

docs/spec/v1beta2/kustomization.md
```diff
@@ -654,6 +655,27 @@ spec:
...
+  name: podinfo
+  namespace: flux-system
+spec:
+  # ...omitted for brevity
+  components:
+  - ingress
+  - tls
+```
+
+**Note:** The component paths must be local and relative.
```
stefanprodan, line 677: Please add here a warning that while components are alpha upstream they are considered experimental in Flux, we offer no guarantees that these will work in the future.

internal/generator/generator.go
klausenbusk: I went full circle on this one, the current implementation is fine, as `filepath.IsAbs` is only checking the string, it isn't checking on on the fs.

docs/spec/v1beta2/kustomization.md
klausenbusk: Like so?

stefanprodan: Great! Thanks you

stefanprodan:
@klausenbusk please run `make generate manifests api-docs` and force push the changes.

stefanprodan:
Yes we need to get the components mapping in `fluxcd/pkg/kustomize` so that Flux users can build manifests locally and diff them with the CLI. Sorry for this, I should’ve mention this earlier.

klausenbusk:
pkg PR created: https://github.com/fluxcd/pkg/pull/412. The `github.com/fluxcd/pkg/kustomize` dependency must be updated when said PR is in.

klausenbusk:
I guess a new pkg kustomize tag is needed, before this can get in.

stefanprodan:
@klausenbusk I've tagged `kustomize/v0.11.0` please update go.mod and run `make tidy`.

klausenbusk:
Rebased and bumped `github.com/fluxcd/pkg/kustomize` in `go.mod`.

PR #760, opened 2022-11-09
build: Bump gpg to alpine's edge

pjbgf:
LibKSBA is a dependency to GnuPG, which has a CVE that is yet to be patched on Alpine's stable channel. This PR installs GnuPG from the edge channel, and should be reverted once libksba's version 1.6.2 is in main.

https://pkgs.alpinelinux.org/packages?name=libksba&branch=edge
https://gnupg.org/blog/20221017-pepe-left-the-ksba.html

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,9 @@ RUN xx-go build -trimpath -a -o kustomize-controller main.go
 
 FROM alpine:3.16
 
-RUN apk add --no-cache ca-certificates tini git openssh-client gnupg
+# Uses GnuPG from edge to patch CVE-2022-3515.
+RUN apk add --no-cache ca-certificates tini git openssh-client && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -33,7 +33,8 @@ RUN xx-go build -trimpath -a -o kustomize-controller main.go
 
 FROM alpine:3.16
 
-RUN apk add --no-cache ca-certificates tini git openssh-client gnupg
+RUN apk add --no-cache ca-certificates tini git openssh-client && \
+	apk add --no-cache gnupg --repository=https://dl-cdn.alpinelinux.org/alpine/edge/main
```
hiddeco, line 37: Might want to add a note here on why we are pulling from edge.

hiddeco (approved):

stefanprodan:
Which scanner reports this CVE for the kustomize-controller image? I don't see it here https://artifacthub.io/packages/helm/fluxcd-community/flux2?modal=security-report

PR #761, opened 2022-11-10
Don't override the reconcile error on status patching

stefanprodan:
Fix bug introduced in #745

### DIFF
--- a/controllers/kustomization_controller.go
+++ b/controllers/kustomization_controller.go
@@ -175,7 +175,10 @@ func (r *KustomizationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 	// Finalise the reconciliation and report the results.
 	defer func() {
-		retErr = r.finalizeStatus(ctx, obj, patcher)
+		// Patch finalizers, status and conditions.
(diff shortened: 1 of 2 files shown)


controllers/kustomization_controller.go
```diff
@@ -175,7 +175,10 @@ func (r *KustomizationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 	// Finalise the reconciliation and report the results.
 	defer func() {
-		retErr = r.finalizeStatus(ctx, obj, patcher)
+		// Patch finalizers, status and conditions.
+		if err := r.finalizeStatus(ctx, obj, patcher); err != nil {
+			log.Error(err, "Status update failed")
```
hiddeco, line 180: Instead of logging the error (and thereby ignoring the fact that the final persistence failed), I would append to `retErr` using [`errors.NewAggregrate`](https://pkg.go.dev/sigs.k8s.io/kind/pkg/errors#NewAggregate) so the reconciliation as a whole is retried (preventing the reflection getting lost until next interval).

darkowlzz (approved):
LGTM!

PR #765, opened 2022-11-14
Remove nsswitch.conf creation from Dockerfile

stefanprodan:
Creating a minimal nsswitch.conf in Alpine is no longer necessary due to https://git.alpinelinux.org/aports/commit/?id=348653a9ba0701e8e968b3344e72313a9ef334e4


pjbgf (approved):
LGTM

thanks @stefanprodan! :bow:

PR #771, opened 2022-11-23
build: Fix cifuzz tests and improve fuzz tests' reliability

pjbgf:
Establish conventions which aligns with what is supported upstream today, whilst expanding on documentation to ensure folks have pointers on how to debug/check for issues going forwards.


stefanprodan (approved):
LGTM

Thanks @pjbgf

PR #774, opened 2022-12-07
Document behaviour of atomic fields with server-side apply

### LINKED ISSUE #772: Patching Node manifest taints removes existing taints
We're experimenting with using flux to control labels, annotations, and taints for our nodes. It seems like the server side apply field manager isn't getting set correctly for array values and is wiping out taints managed by other controllers.

flux version: v0.30.2
kustomize-controller version: v0.25.0
kubernetes version: 1.23.0

Given the following node patch spec that we want to merge into the existing node definition:

```yaml
apiVersion: v1
kind: Node
metadata:
  name: dev-admiral
  labels:
    ecosystem: platform
  annotations:
    kustomize.toolkit.fluxcd.io/prune: disabled
    kustomize.toolkit.fluxcd.io/ssa: merge
spec:
  taints:
    - effect: NoSchedule
      key: migration-safety
```

The resulting managedFields spec for the node after flux has reconciled the manifest becomes:

```yaml
managedFields:
  - apiVersion: v1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          f:kustomize.toolkit.fluxcd.io/prune: {}
          f:kustomize.toolkit.fluxcd.io/ssa: {}
        f:labels:
          f:kustomize.toolkit.fluxcd.io/name: {}
          f:kustomize.toolkit.fluxcd.io/namespace: {}
          f:ecosystem: {}
      f:spec:
        f:taints: {}
    manager: kustomize-controller
    operation: Apply
    time: "2022-11-30T23:01:51Z"
  ...
```

The important part in my mind is the 
```yaml
      f:spec:
        f:taints: {}
```

The existing taints, and I assume any other array field I tried to patch with this approach, are wiped out.

What I would expect to see in the "kustomize-controller" field manager entry is this instead only for the specified taints:
```yaml
      f:spec:
        f:taints:
          k:{"key":"migration-safety"}:
            .: {}
            f:effect: {}
```

somtochiama:
Closes #772 

### DIFF
--- a/docs/spec/v1beta2/kustomization.md
+++ b/docs/spec/v1beta2/kustomization.md
@@ -281,6 +281,11 @@ kustomize.toolkit.fluxcd.io/ssa: merge
 **Note:** The fields defined in manifests will always be overridden,
 the above procedure works only for adding new fields that don’t overlap with the desired state.
 
+For lists fields which are atomic (e.g `spec.tolerations` in PodSpec), Kubernetes doesn't allow different managers 
+for such fields, therefore any changes to these fields will be undone, even if you specify a manager. 
+For more context, please see the Kubernetes enhancement doc:
+[555-server-side-apply](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/555-server-side-apply/README.md#lists).
+
 ## Garbage collection
 
 To enable garbage collection, set `spec.prune` to `true`.


stefanprodan (approved):
LGTM

Thanks @somtochiama

PR #789, opened 2023-01-30
Disable caching of Secrets and ConfigMaps

hiddeco:
You can re-enable caching by starting the controller with the argument `--feature-gates=CacheSecretsAndConfigMaps=true`.

xref: https://github.com/fluxcd/flux2/issues/3426


stefanprodan (approved):
LGTM

Thanks @hiddeco

Issue #801, opened 2023-02-27
kustomize-controller getting OOMKilled


vytautaskubilius:
After deploying a handful of Kustomizations I have noticed that the `kustomize-controller` pods were being OOMKilled on every reconciliation. At the moment our deployment has 12 Kustomizations in total, and the only way I've found to keep the pods stable was to increase their memory limit to 4Gi. The pod now seems to consistently sit at ~3.5Gi of used memory:

```
kubectl -n flux-system top pods
NAME                                           CPU(cores)   MEMORY(bytes)
helm-controller-67b65ff599-8tp9l               2m           64Mi
image-automation-controller-7f9b499997-v92q8   5m           31Mi
image-reflector-controller-65f575dbd-5rf2q     13m          91Mi
kustomize-controller-bddb7f558-h44ff           1m           3311Mi
notification-controller-654c5d586b-w2x84       1m           13Mi
source-controller-59498d9f8d-pkztj             13m          79Mi
```

`flux stats -A` doesn't appear to show anything out of ordinary:

```
flux stats -A
RECONCILERS          	RUNNING	FAILING	SUSPENDED	STORAGE
GitRepository        	1      	0      	0        	54.0 KiB
OCIRepository        	0      	0      	0        	-
HelmRepository       	6      	0      	0        	3.6 MiB
HelmChart            	6      	0      	0        	768.6 KiB
Bucket               	0      	0      	0        	-
Kustomization        	12     	0      	0        	-
HelmRelease          	6      	0      	0        	-
Alert                	0      	0      	0        	-
Provider             	0      	0      	0        	-
Receiver             	0      	0      	0        	-
ImageUpdateAutomation	1      	0      	0        	-
ImagePolicy          	11     	0      	0        	-
ImageRepository      	11     	0      	0        	-
```

While the pod is stable for the time being, in the next few weeks we will be deploying more applications (~250 in total), and I'm worried that Flux won't be able to scale if memory usage increases linearly with the amount of Kustomizations that we will be deploying.

I'm attaching a [heap profile](https://fluxcd.io/flux/gitops-toolkit/debugging/#collecting-a-profile) from the pod taken while it was reporting high memory usage in case it's helpful - [`heap.tar.gz`](https://github.com/fluxcd/kustomize-controller/files/10841644/heap.tar.gz).

Is the `kustomize-controller` pod supposed to consume this much memory? If not, is there any more data I could collect and provide to help with troubleshooting?

[...]

PR #803, opened 2023-03-02
kstatus: allow disabling of caching cluster reader

hiddeco:
This commit allows the disabling of the caching cluster reader used by
the status poller while waiting and/or checking the health of resources.
Potentially reducing the memory usage of the controller on large scale
clusters, at the cost of an increase in direct API calls.

The feature can be enabled using
`--feature-gates=DisableStatusPollerCache=true`.

Potentially helps aid scenario described in #801


stefanprodan (approved):
LGTM

Thanks @hiddeco 🏅

PR #810, opened 2023-03-08
Allow decryption secrets in other namespaces

toelke:
This patch changes the API for `Kustomize` slightly in that it allows decryption secrets to be referenced from another namespace.
I have tested this for backwards-compatibility and found no issues.


hiddeco:
See https://github.com/fluxcd/flux2/discussions/3639#discussioncomment-5194031

toelke:
Ah. I had only searched issues and PRs, not discussions.
Well, let's take a look at kyverno then.

Thank you and sorry for the noise!

Issue #508, opened 2021-12-04
Allow Common Labels and Common Annotations Transformers on Kustomization Apply


jonathan-innis:
Currently, there doesn't seem to be a way to append a given label to all Kubernetes objects that my Kustomization applies. If I want to leverage the `.spec.patches` property in `Kustomization` to achieve this, I either have to be willing to completely overwrite my `/metadata/labels` field or I have to already have the `label` field available in each `metadata` section of all objects in my git repo as described [here](https://stackoverflow.com/questions/55573724/create-a-patch-to-add-a-kubernetes-annotation)

### Proposal
Enable a `spec.commonLabels` and `spec.commonAnnotations` similar to the [CommonLabel](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/transformerconfigs/README.md#labels-transformer) and [CommonAnnotation](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/transformerconfigs/README.md#annotations-transformer) transformers available natively with Kustomize

stefanprodan:
The major issue with the common labels and annotations is that they are applied to pod selectors which are immutable thus breaking the sync.

jonathan-innis:
@stefanprodan This is not an issue with `spec.commonAnnotations` since it only updates the metadata and not the selectors for any resources. Feasibly, we could alter the behavior of the CommonLabel transformer slightly so we don't update the selectors of resources, though this obviously deviates slightly from Kustomize

JamWils:
@stefanprodan I'm a little confused about the explanation above.  In what way does it break the sync?  Do you mean that we would have to destroy the pods to apply new common labels if they are added to the flux kustomization?

stefanprodan:
@JamWils that's how Kubernetes works, pod labels are immutable, so you need to delete the Deployments/StatefulSets/DaemonSets to change them. See the docs here https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates

JamWils:
I 100% agree with that.  However, if I change the label on a pod manually and then push then the kustomization controller will replace the pod with the new label.  If we had a commonLabels on the flux kustomization wouldn't we expect the same behavior?  I.e. I add a common label `test: this` and then I would see that common label applied to the resources kustomize is working with.

AndersBennedsgaard:
[...]

PR #817, opened 2023-03-21
Add common labels and annotations patching capabilities 

stefanprodan:
This PR adds an optional field to the API called `CommonMetadata` for setting labels and/or annotations to all resources part of a Kustomization.

Example:

```yaml
kind: Kustomization
metadata:
  name: my-app
  namespace: apps
spec:
  # ...omitted for brevity
  commonMetadata:
    labels:
      app.kubernetes.io/part-of: my-app
    annotations:
      a8r.io/owner: my-team
```

Note that any existing label or annotation will be overridden if its key matches a common one.

⚠️ The main difference to the Kustomize [commonLabels](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/commonlabels/) and [commonAnnotations](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/commonannotations/), is that the controller sets the labels and annotations only to the top level `metadata` field, without patching the Kubernetes Deployment `spec.template` or the Service `spec.selector`. 

Closes: #508


hiddeco (approved):

haarchri:
@transacid check this feature

PR #818, opened 2023-03-24
Verify Digest of Artifact

hiddeco:
This commit changes the verification to the `Digest` field of the Artifact, which was introduced in source-controller v0.35.0. As the `Checksum` is expected to be removed in the next release of the source-controller.


stefanprodan (approved):
LGTM

Thanks @hiddeco

PR #821, opened 2023-03-29
Add reconciler sharding capability based on label selector

stefanprodan:
Implementation of the work done in https://github.com/fluxcd/pkg/pull/525

Part of https://github.com/fluxcd/flux2/issues/3731


main.go
```diff
@@ -141,10 +146,15 @@ func main() {
 		LeaseDuration:                 &leaderElectionOptions.LeaseDuration,
 		RenewDeadline:                 &leaderElectionOptions.RenewDeadline,
 		RetryPeriod:                   &leaderElectionOptions.RetryPeriod,
-		LeaderElectionID:              fmt.Sprintf("%s-leader-election", controllerName),
+		LeaderElectionID:              leaderelection.GenerateID(controllerName, watchOptions.LabelSelector),
```
hiddeco, line 149: You need to check for an `LabelSelector != ""` and compose it until I have fixed `GenerateID` to deal with empty string arguments, as otherwise it will be suffixed with `-e3b0c4429`.

hiddeco: Example: https://github.com/fluxcd/source-controller/blob/52d7bab10d3744bc9631c158f9f23be6fc224382/main.go#L335-L338

hiddeco (approved):

Issue #755, opened 2022-11-01
GA: Promote `Kustomization` to `kustomize.toolkit.fluxcd.io/v1`


pjbgf:
Items to be reviewed plus any outstanding work required to bump API to `v1`:
- [ ] Documentation
- [x] API Validation
- [x] Test Coverage

PR #822, opened 2023-03-29
GA: Promote Kustomization API to `kustomize.toolkit.fluxcd.io/v1`

stefanprodan:
## API changes

The `Kustomization` kind was promoted from v1beta2 to v1 (GA) and deprecated fields were removed.

The `kustomizations.source.toolkit.fluxcd.io` CRD contains the following versions: 
- v1 (storage version)
- v1beta2 (deprecated)
- v1beta1 (deprecated)

### Upgrade

The `Kustomization` v1 API is backwards compatible with v1beta2 with the following exceptions:
- the deprecated field `.spec.validation` was removed 
- the deprecated field `.spec.patchesStrategicMerge` was removed (replaced by `.spec.patches`)
- the deprecated field `.spec.patchesJson6902 ` was removed (replaced by `.spec.patches`)

To upgrade from v1beta2, after deploying the new CRD and controller, set  `apiVersion: kustomize.toolkit.fluxcd.io/v1` in the YAML files that contain `Kustomization` definitions and remove the deprecated fields if any. Bumping the API version in manifests can be done gradually. It is advised to not delay this procedure as the beta versions will be removed after 6 months.

Closes: #755
Closes: #664


docs/spec/v1/kustomization.md
```diff
@@ -0,0 +1,1752 @@
...
+```yaml
+kind: Kustomization
+secretGenerator:
+  - name: ghcr-auth
+    type: kubernetes.io/dockerconfigjson
+    files:
+      - .dockerconfigjson=ghcr.dockerconfigjson.encrypted
+```
+
+### Recommend settings
```
aryan9600, line 1367: should this be not here and have its own H2 heading?

docs/spec/v1/kustomization.md
```diff
@@ -0,0 +1,1750 @@
+# Kustomization
+
```
makkes, line 2: We need to set a menu weight here as we did in [notification-controller](https://github.com/fluxcd/notification-controller/pull/498/files#diff-912aec2e27ddfd4a4a99a1c3882d3e9b5d6cdcae48d7566eaa09920c0327c432R3) so that the API reference navbar entry is properly rendered at the bottom.

docs/api/kustomize.md
```diff
@@ -2,16 +2,16 @@
 <p>Packages:</p>
 <ul class="simple">
 <li>
-<a href="#kustomize.toolkit.fluxcd.io%2fv1beta2">kustomize.toolkit.fluxcd.io/v1beta2</a>
+<a href="#kustomize.toolkit.fluxcd.io%2fv1">kustomize.toolkit.fluxcd.io/v1</a>
```
makkes, line 5: To align the structure within docs/api/ with NC and SC we should move this file (docs/api/kustomize.md) into docs/api/v1/ (I know it's only one API version at this point but it makes it easier for contributors to navigate this directory structure when it's the same across all Flux repos).

stefanprodan: I have updated the docs and moved the generated ones to `/v1`, please make these changes in the website PR to verify all is good.

docs/spec/v1/kustomization.md
stefanprodan: Done!

api/v1/groupversion_info.go
```diff
@@ -0,0 +1,33 @@
...
+package v1
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+var (
+	// GroupVersion is group version used to register these objects.
+	GroupVersion = schema.GroupVersion{Group: "kustomize.toolkit.fluxcd.io", Version: "v1beta2"}
```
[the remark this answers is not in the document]
stefanprodan: This explains the deprecation warnings I see in test logs. Good catch Sunny!

hiddeco (approved):
This looks good enough to me to be released.

In terms of the spec doc, I think it could use another proof read to see if more interlinking between sections is possible, and to align tone of voice between different sections. In addition to reducing the usage of "kustomize-controller" as at some point based on the context it can just be called "controller", and other nits like this. But not blocking to move forward, as structure wise it's :100:.

Thanks @stefanprodan and @aryan9600 :bow: :1st_place_medal:

aryan9600:
> I think it could use another proof read to see if more interlinking between sections is possible
i completely agree. one particular instance that comes to mind is how can we make the [sops guide](https://fluxcd.io/flux/guides/mozilla-sops/) and the decryption section more compatible/co-existent?

docs/spec/v1/kustomization.md
```diff
@@ -939,10 +949,13 @@ data:
   sops.vault-token: <BASE64>
 ```
 
-## Recommended settings
+## Working with Kustomizations
+
+### Recommended settings
```
aryan9600, line 954: `Recommended settings` used to be inside this section, but we moved it out so that it could have a greater emphasis (ref: https://github.com/fluxcd/kustomize-controller/pull/822#discussion_r1152266697)

hiddeco:
Another thing I ran into is that we first talk about

> By default, a Kustomization apply runs under the cluster admin account [...]

but then later have a section on how this default can be changed. I think this should eventually be rewritten in a way which brings the override more to the foreground, as people who read lazily may have had their alarm bells ring before they actually notice this.

docs/spec/v1/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+This is the v1 API specification for defining continuous delivery pipelines
+of Kubernetes objects generated with Kustomize.
+
+## Specification
+
+- [Kustomization CRD](kustomization.md)
+    + [Example](kustomization.md#example)
+    + [Writing a Kustomization spec](kustomization.md#writing-a-kustomization-spec)
+    + [Recommended settings](kustomization.md#recommended-settings)
```
aryan9600, line 11:
```suggestion
```

aryan9600 (changes requested):

aryan9600 (approved):

PR #838, opened 2023-04-21
Fix SOPS azkv envCred

apeschel:
At the moment, the envCred logic can't actually set the Azure credentials.

This commit fixes the logic so that the environment variables can actually be used to set the Azure credentials.

There are other issues that come up from this block of code, but those can be dealt with separately.


apeschel:
I ran this locally and verified it fixes the issue. It does run into rate limiting with the Azure API, but that is a separate issue to solve.

hiddeco (approved):

stefanprodan (approved):

PR #850, opened 2023-05-04
update k8s to v1.27 and kustomize to v5

aryan9600:
Changes made due to breaking changes in controller-runtime:
* `Watches` accepts a plain `client.Object` instead of a `source.Kind` object.
* Some fields in `controller.Options` are now in `config.Controller` which is embedded in `controller.Options`.
* `handler.MapFunc` now accepts a context.
* `ctrlcache.Options` uses `ByObject` instead of `SelectorByObject`.


internal/controllers/kustomization_indexers.go
```diff
@@ -29,8 +30,8 @@ import (
 	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1"
 )
 
-func (r *KustomizationReconciler) requestsForRevisionChangeOf(indexKey string) func(obj client.Object) []reconcile.Request {
-	return func(obj client.Object) []reconcile.Request {
+func (r *KustomizationReconciler) requestsForRevisionChangeOf(indexKey string) handler.MapFunc {
+	return func(ctx context.Context, obj client.Object) []reconcile.Request {
```
hiddeco, line 34: The `panic` can be replaced by a logged error using `ctx` and an early return.

main.go
```diff
@@ -148,21 +148,24 @@ func main() {
...
+			Cache: &ctrlclient.CacheOptions{
+				DisableFor: disableCacheFor,
+			},
+		},
+		Cache: ctrlcache.Options{
+			ByObject: map[ctrlclient.Object]ctrlcache.ByObject{
 				&kustomizev1.Kustomization{}: {Label: watchSelector},
 			},
-		}),
+			Namespaces: []string{watchNamespace},
```
aryan9600, line 167: for it to watch all namespaces, the list should be nil/empty, but I don't think `[""]` is the same as that

internal/controllers/kustomization_indexers.go
```diff
@@ -29,20 +31,21 @@ import (
...
+		log := ctrl.LoggerFrom(ctx)
 		repo, ok := obj.(interface {
 			GetArtifact() *sourcev1.Artifact
 		})
 		if !ok {
-			panic(fmt.Sprintf("Expected an object conformed with GetArtifact() method, but got a %T", obj))
+			log.Error(fmt.Errorf("expected an object conformed with GetArtifact() method, but got a %T", obj), "object type mismatch")
```
hiddeco, line 41:
```suggestion
			log.Error(fmt.Errorf("expected an object conformed with GetArtifact() method, but got a %T", obj), "failed to get requests for revision change")
```

maybe?

internal/controllers/kustomization_indexers.go
```diff
@@ -29,20 +31,21 @@ import (
...
+			return nil
```
hiddeco, line 42: The `List` error below can also be captured in logs now.

hiddeco (approved):

PR #856, opened 2023-05-11
Upgrade alpine v0.34.0 fix lib curl

SohailWahab-hub:

PR #858, opened 2023-05-11
Upgrade alpine - fix lib curl cve for tag v0.340

SohailWahab-hub:


hiddeco:
This is now the third PR you are creating while we already marked it as a duplicate in https://github.com/fluxcd/kustomize-controller/pull/856. Please refrain from opening any more.

PR #897, opened 2023-06-23
Use kustomization namespace for empty dependency source namespace

devholic:
Kustomize controller uses the namespace of the kustomization resource if the sourceRef is empty. However, this policy doesn't applied to dependencies. This can be problematic if the same named `Sources` without explicit namespace is in different namespace.

This commit fixes this issue by using kustomization's namespace when checking dependencies if the namespace in sourceRef is empty.


stefanprodan (approved):
LGTM

Great find, thanks @devholic for the fix 🥇

PR #917, opened 2023-07-04
ci: Group all GH action updates under the same PR

stefanprodan:
Beta feature: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -4,6 +4,13 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     labels: ["area/ci", "dependencies"]
+    groups:
+      # Group all updates together, so that they are all applied in a single PR.
+      # Grouped updates are currently in beta and is subject to change.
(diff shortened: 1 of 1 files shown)


.github/dependabot.yaml
```diff
@@ -4,6 +4,11 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     labels: ["area/ci", "dependencies"]
+    # Group all GH action updates under the same PR
+    groups:
```
hiddeco, line 8:
```suggestion
    groups:
        # Group all updates together, so that they are all applied in a single PR.
        # Grouped updates are currently in beta and is subject to change.
        # xref: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups
```

is what I did in helm-controller, as we will probably have to look at the docs once in awhile.

hiddeco (approved):

Issue #919, opened 2023-07-05
Flux sends notifications when kustomization reconcile is set to disabled


craigbryson:
### Describe the bug

As per a slack discussion, we have a Service in kubernetes that is managed by flagger, the definition for which is reconciled by Flux, it seems Flagger is adding the `kustomize.toolkit.fluxcd.io/reconcile=disabled` annotation, which is the correct behaviour. We receive a notification at every reconcile interval informing us that the Service was skipped.

The notification controller should not send notifications about ignored objects.

### Steps to reproduce

Deploy a service via Flux with the kustomize.toolkit.fluxcd.io/reconcile=disabled annotation.
Set up notification controller to send all events (in our case we use MS Teams)
See that an event is emitted for the disabled service when the kustomization is reconciled.

### Expected behavior

The notification controller should not send notifications about ignored objects.

### OS / Distro

N/A

### Flux version

v0.41.1

### Flux check

► checking prerequisites
✗ flux 0.41.1 <2.0.0-rc.5 (new version is available, please upgrade)
✔ Kubernetes 1.26.3 >=1.20.6-0
► checking controllers
✔ helm-controller: deployment ready
► ghcr.io/fluxcd/helm-controller:v0.31.1
✔ image-automation-controller: deployment ready
► ghcr.io/fluxcd/image-automation-controller:v0.31.0
✔ image-reflector-controller: deployment ready
► ghcr.io/fluxcd/image-reflector-controller:v0.26.0
✔ kustomize-controller: deployment ready
► ghcr.io/fluxcd/kustomize-controller:v0.35.0
✔ notification-controller: deployment ready
► ghcr.io/fluxcd/notification-controller:v0.33.0
✔ source-controller: deployment ready
► ghcr.io/fluxcd/source-controller:v0.36.0
► checking crds
✔ alerts.notification.toolkit.fluxcd.io/v1beta2
✔ buckets.source.toolkit.fluxcd.io/v1beta2
✔ gitrepositories.source.toolkit.fluxcd.io/v1beta2
✔ helmcharts.source.toolkit.fluxcd.io/v1beta2
✔ helmreleases.helm.toolkit.fluxcd.io/v2beta1
✔ helmrepositories.source.toolkit.fluxcd.io/v1beta2
✔ imagepolicies.image.toolkit.fluxcd.io/v1beta2
✔ imagerepositories.image.toolkit.fluxcd.io/v1beta2
[...]

PR #920, opened 2023-07-06
Exclude skipped resources from apply events

stefanprodan:
Fix: #919

### DIFF
--- a/internal/controller/utils.go
+++ b/internal/controller/utils.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+
+	"github.com/fluxcd/pkg/ssa"
 )
 
 // MkdirTempAbs creates a tmp dir and returns the absolute path to the dir.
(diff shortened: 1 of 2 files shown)


internal/controller/utils.go
```diff
@@ -36,3 +38,16 @@ func MkdirTempAbs(dir, pattern string) (string, error) {
...
 	return tmpDir, nil
 }
+
+// HasChanged evaluates the given action and returns true
+// if the action type matches a resource mutation or deletion.
+func HasChanged(action ssa.Action) bool {
+	switch action {
+	case ssa.SkippedAction:
+		return false
+	case ssa.UnchangedAction:
```
hiddeco, line 48: What about `ssa.UnknownAction`?

stefanprodan: UnknownAction would be accompanied by an error, which always should alert.

hiddeco (approved):

PR #930, opened 2023-07-24
Handle delete before adding finalizer

darkowlzz:
In `Reconcile()` method, move the object deletion above add finalizer. Finalizers can't be set when an object is being deleted.

Found by @hiddeco while working on helm-controller. Refer https://github.com/kubernetes-sigs/cluster-api/commit/cef1cf1d9312088a785c520f49970b1c609ab9ad .


hiddeco (approved):
Backporting this change shouldn’t necessarily be required as we do not have multiple reconcilers handling this resource.

It could however become a problem if an arbitrary 3rd party registers a finalizer on the object.

Thanks @darkowlzz 🙇

PR #933, opened 2023-07-31
Enable fail-fast behavior for health checks

stefanprodan:
Fail the health check as soon as a resource becomes stalled without waiting for the timeout to expire.

For example, if `wait: true` and `timeout: 20m` and a Deployment has reached its deadline progressing in 5m, the controller will not wait for another 15m, it will fail the reconciliation when the Deployment rollout has stalled. 

Note that the fail-fast behavior does not currently work with HelmReleases as these don't have a stalled condition.  We expect to ship stalled conditions in the HelmRelease API v2beta2.

Fix: https://github.com/fluxcd/flux2/discussions/3980

This behavior can be disabled using the `DisableFailFastBehavior` feature flag.

### DIFF
--- a/internal/controller/kustomization_controller.go
+++ b/internal/controller/kustomization_controller.go
@@ -89,6 +89,7 @@ type KustomizationReconciler struct {
 	statusManager         string
 	NoCrossNamespaceRefs  bool
 	NoRemoteBases         bool
+	FailFast              bool
 	DefaultServiceAccount string
 	KubeConfigOpts        runtimeClient.KubeConfigOptions
 }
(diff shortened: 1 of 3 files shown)


hiddeco (commented):
While this pull request by itself looks good to me, I am wondering if there is any thought and/or reasoning around this being a global flag versus a field in the API?

stefanprodan:
@hiddeco I think failing fast is how heath checking should behave. The feature flag is temporary, if no issues will arise, we'll remove it in a future minor version.

hiddeco (approved):

makkes (approved):
🎉 Would we be able to add a test for this feature?

stefanprodan:
@makkes fail-fast tests where added in the ssa package.

PR #972, opened 2023-09-18
Add optional built-in substitution of last attempted revision (1.1.x)

arikkfir:
The purpose of this new feature (which is CLI-conditional, and defaults to false in order to preserve compatibility) is to allow to refer to the revision that is currently being applied in the resource manifests, via the standard substitutions mechanism.

This PR is for the `release/v1.1.x` branch; a corresponding PR for the `main` branch exists in #971


hiddeco:
We backport changes automatically using when required, and given this is a new feature, it would not require a backport to an existing minor.

arikkfir:
Ah got it, thanks 🙏

PR #986, opened 2023-10-09
Address miscellaneous issues throughout code base

hiddeco:


internal/decryptor/decryptor.go
```diff
@@ -508,7 +508,7 @@ func (d *Decryptor) sopsDecryptFile(path string, inputFormat, outputFormat forma
 	if err != nil {
 		return err
 	}
-	err = os.WriteFile(path, out, 0o644)
+	err = os.WriteFile(path, out, 0o600)
```
stefanprodan, line 511: I propose we revert this change so we can merge this PR. We can do this change in dedicated PR and add test to see how it impacts kustomize build.

hiddeco: Dropped the commit.

stefanprodan (approved):
LGTM

PR #990, opened 2023-10-11
Add changelog entry for v1.1.1

stefanprodan:
Cherrypick from https://github.com/fluxcd/kustomize-controller/pull/989


hiddeco (approved):
When cherry-picking, it's wise to use `-x` to automatically include the commit reference it is picked from.

stefanprodan:
Ok let me redo it

PR #1017, opened 2023-11-25
Support additional Field Managers in the disallow list

### LINKED ISSUE #982: Support Additional Field Managers
In working with this controller and messing around with `kubectl edit` and Lens to make changes to resources I have found that Lens will update the list of `managedFields` with it's own entry where `manager` is set to `node-fetch`. As this is not part of the below code Flux does not undo the changes made (assuming they do not conflict with flux managed fields) and we are left with the resource being in an unknown state where the changes have come from outside the cluster.

It would be great to be able to configure the below list so that we can have other managers in this list:
https://github.com/fluxcd/kustomize-controller/blob/5e11b653daa4fc47f97857786c6cf2a3d43b9ccb/internal/controller/kustomization_controller.go#L684-L705

Nalum:
Setup a new flag to pass in the names of managers that should be overwritten in the same way that `kubectl` changes are.

Closes #982

### DIFF
--- a/internal/controller/kustomization_disallowed_managers_test.go
+++ b/internal/controller/kustomization_disallowed_managers_test.go
@@ -1,0 +1,156 @@
+/*
+Copyright 2023 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/fluxcd/pkg/apis/meta"
+	"github.com/fluxcd/pkg/testserver"
+	sourcev1 "github.com/fluxcd/source-controller/api/v1"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1"
+)
+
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -94,6 +94,7 @@ func main() {
 		httpRetry             int
 		defaultServiceAccount string
 		featureGates          feathelper.FeatureGates
+		fieldManagers         []string
```
stefanprodan, line 97:
```suggestion
		disallowedFieldManagers         []string
```

stefanprodan: Also please prefix with `disallowed` in all other files and vars.

stefanprodan (commented):
@Nalum please add tests for this, similar to kustomization_force_test.go

internal/controller/kustomization_disallowed_managers_test.go
```diff
@@ -0,0 +1,155 @@
+/*
+Copyright 2021 The Flux authors
```
stefanprodan, line 2:
```suggestion
Copyright 2023 The Flux authors
```

internal/controller/kustomization_disallowed_managers_test.go
```diff
@@ -0,0 +1,155 @@
...
+	})
+
+	t.Run("update configmap with new data", func(t *testing.T) {
+		configMap := corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      id,
+				Namespace: id,
+			},
+		}
+		_ = k8sClient.Patch(context.Background(), &configMap, client.RawPatch(types.MergePatchType, []byte(`{"data":{"bad-key":"overridden field manager"}}`)), &client.PatchOptions{FieldManager: overrideManagerName})
```
stefanprodan, line 137: Let's check the errors with `g.Expect(err).NotTo(HaveOccurred())` for both patches.

internal/controller/kustomization_disallowed_managers_test.go
```diff
@@ -0,0 +1,155 @@
...
+		_ = k8sClient.Patch(context.Background(), &configMap, client.RawPatch(types.MergePatchType, []byte(`{"data":{"key2":"not overridden field manager"}}`)), &client.PatchOptions{FieldManager: "good-name"})
+
+		g.Eventually(func() bool {
```
stefanprodan, line 140: No need for `Eventually` as the above patches are not async.

Nalum:
@stefanprodan would this change be picked up automatically in the docs or should I update them? Assuming I should update them they are located in the flux2 repo?

stefanprodan:
@Nalum please squash into one or two commits and force push. I will make the docs changes in website before the release, there are other changes besides this to the controller global configuration.

Nalum:
Think this is good for testing now @stefanprodan 👍

PR #1028, opened 2023-12-13
Log HTTP errors to provide faster feedback

hiddeco:
This configures a logger on the archive fetcher, to make HTTP errors surface faster when it fails to retrieve the artifact of a source.

Part of: https://github.com/fluxcd/flux2/issues/4472


stefanprodan (approved):
LGTM

Thanks @hiddeco

PR #1030, opened 2023-12-13
Trim errors reported in status conditions

stefanprodan:
Fix: https://github.com/fluxcd/flux2/issues/4470

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,9 +22,9 @@ require (
 	github.com/fluxcd/pkg/apis/event v0.6.0
 	github.com/fluxcd/pkg/apis/kustomize v1.2.0
 	github.com/fluxcd/pkg/apis/meta v1.2.0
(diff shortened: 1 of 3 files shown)


internal/controller/utils.go
```diff
@@ -51,3 +51,17 @@ func HasChanged(action ssa.Action) bool {
...
 }
+
+// TrimError returns the error message trimmed to a maximum length of 15K characters
+// which should fit in the status condition message.
+func TrimError(err error) string {
+	if err == nil {
+		return ""
+	}
+	errMsg := err.Error()
+	maxLength := 15000
```
hiddeco, line 62:
```suggestion
	const maxLength = 15000
```

What made you pick `15000` instead of `32765`?

stefanprodan: Some errors end up twice in the status and it could reach the max size of the whole object.

PR #1077, opened 2024-02-01
Release v1.2.2

darkowlzz:
Part of https://github.com/fluxcd/flux2/issues/4556.

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -5,4 +5,4 @@ resources:
 images:
   - name: fluxcd/kustomize-controller
     newName: fluxcd/kustomize-controller
-    newTag: v1.2.1
+    newTag: v1.2.2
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,41 @@
...
+## 1.2.2
+
+**Release date:** 2024-02-01
+
+This patch release comes with various bug fixes and improvements. The kustomize
+build regression for empty directory has been fixed. The regression due to which
+the namespaced objects without a namespace specified resulted in `not found`
+error instead of `namespace not specified` has also been fixed. And the
+regression due to which Roles and ClusterRoles were reconciled over and over due
+to the normalization of Roles and ClusterRoles has also been fixed.
```
stefanprodan, line 14:
```suggestion
This patch release comes with various bug fixes and improvements.

Reconciling empty directories and  directories without Kubernetes manifests no longer results in an error. This
regressing bug was introduced with the controller upgrade to Kustomize v5.3 and has been fixed in this patch release.

The regression due to which the namespaced objects without a namespace specified resulted in `not found`
error instead of `namespace not specified` has also been fixed. And the
regression due to which Roles and ClusterRoles were reconciled over and over due
to the normalization of Roles and ClusterRoles has also been fixed.
```

PR #1085, opened 2024-02-09
chore: set explicitly user id and group id

mloiseleur:
# Motivation

1. Kubernetes [documentation says](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container) on `runAsGroup`:

>  If this field is omitted, the primary group ID of the containers will be root(0).

2. Security scanner, Kubescape,  recommends with [C-0013](https://hub.armosec.io/docs/c-0013) to set those values explicitly. 

If this PR is accepted, I can open similar PR on the other components.

### DIFF
--- a/config/manager/deployment.yaml
+++ b/config/manager/deployment.yaml
@@ -30,11 +30,12 @@ spec:
           allowPrivilegeEscalation: false
           readOnlyRootFilesystem: true
           runAsNonRoot: true
+          runAsUser: 65534
+          runAsGroup: 65534
           capabilities:
             drop: [ "ALL" ]
           seccompProfile:
             type: RuntimeDefault
-
         ports:
           - containerPort: 8080
             name: http-prom


souleb:
It seems that it wouldn't work on `Openshift` as you need to use a Security Context Constraints(SCC) instead.

mloiseleur:
\o. I was not expecting this.
In current OpenShift documentation, in [examples](https://docs.openshift.com/container-platform/4.14/authentication/managing-security-context-constraints.html#security-context-constraints-example_configuring-internal-oauth) on SCC it says when runAsUser is explicitely set:
> A container or pod that requests a specific user ID will be accepted by OpenShift Container Platform only when a service account or a user is granted access to a SCC that allows such a user ID.

So I guess I have to close this PR and keep my divergence locally.

Thanks for your review.

Issue #728, opened 2022-09-12
validation error: cannot convert int64 to string


cdenneen:
Taking something like this (while validates fine locally) errors in kustomize-controller:
```yaml
- name: fpm-metrics
  image: "hipages/php-fpm_exporter:1"
  imagePullPolicy: IfNotPresent
  readinessProbe:
    httpGet:
      path: /metrics
      port: fpm-metrics
    periodSeconds: 5
  livenessProbe:
    httpGet:
      path: /metrics
      port: fpm-metrics
    periodSeconds: 5
  startupProbe:
    failureThreshold: 120
    periodSeconds: 1
    httpGet:
      path: /metrics
      port: fpm-metrics
  env:
    # FastCGI address where FPM listens on, we're connecting over TCP
    - name: PHP_FPM_SCRAPE_URI
      value: tcp://127.0.0.1:9666/status
    # Address on which to expose metrics and web interface.
    - name: PHP_FPM_WEB_LISTEN_ADDRESS
      value: 6999
```

quoting `value: "6999"` is only way to fix it.

xeruf:
may be related to https://github.com/calcom/cal.com/issues/12201

Issue #554, opened 2022-01-31
substituteFrom value used in a yaml template are not being rendered correctly when used with quotes.  


jbehrends:
I believe this the correct project for this bug report, but if it's not I'm more than happy to resubmit it somewhere else.

I have a Kustomization (kustomize.toolkit.fluxcd.io/v1beta2) which pulls values from a configMap using postBuild: substituteFrom:
The values are then used in a HelmRelease (helm.toolkit.fluxcd.io/v2beta1) called by this Kustomization.  One of the values is an AWS account number which is defined as a string in the configMap.  The variable is then used in the HelmRelese manifest and wrapped in quotes.  Once the Kustomize controller renders this template and applies it to Kubernetes, the double quotes get stripped away.

Example:

Snippet of my Kustomization file:
```
---
apiVersion: kustomize.toolkit.fluxcd.io/v1beta2
kind: Kustomization
...
spec:
  postBuild:
    substituteFrom:
      - kind: ConfigMap
        name: myvalues
``` 

Configmap that's being called above:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: myvalues
data:
  AWS_ACCOUNT_ID: "887265102384"     # Not our ACTUAL account number. :)
...
``` 

Snippet of my Helmrelease yaml file thats called as part of the above Kustomization
```
---
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: myhelmrelease
...
spec:
  values:
      customMetadata:    # Experiment with adding quotes multiple ways.
        - name: aws_account_id1
          value: '887265102384'   # Passing in a static value with quotes as a control
        - name: aws_account_id2
          value: "${AWS_ACCOUNT_ID}"   # Value wrapped in double quotes
        - name: aws_account_id3
          value: '${AWS_ACCOUNT_ID}'   # Value wrapped in single quotes   
        - name: aws_account_id4
          value: "'${AWS_ACCOUNT_ID}'"   # Value wrapped in double then single quotes
        - name: aws_account_id5
          value: '"${AWS_ACCOUNT_ID}"'   # Value wrapped in single then double quotes
        - name: aws_account_id6
          value: ${AWS_ACCOUNT_ID}       # Value without any quotes     
```

Snipped of Helmrelease manifest with values that ends up in my cluster:
```
$ kubectl get helmrelease myhelmrelease -o yaml
...
      customMetadata:
      - name: aws_account_id1
[...]

PR #1129, opened 2024-04-06
Document how to use numbers and booleans in post build substitutions

stefanprodan:
When using post build substitutions with vars that are numbers or booleans, the reconciliation may fail if the substitution
is for a field that must be of type string. This PR adds a section to the API docs on how to use double quotes vars, and adds unit tests for this procedure.

Fix: https://github.com/fluxcd/flux2/issues/4154
Fix: https://github.com/fluxcd/flux2/issues/1839
Fix: #1096
Fix: #728
Fix: #554

### DIFF
--- a/docs/spec/v1/kustomizations.md
+++ b/docs/spec/v1/kustomizations.md
@@ -624,11 +624,14 @@ kind: Kustomization
 metadata:
   name: apps
 spec:
-  ...
   postBuild:
     substitute:
       var_substitution_enabled: "true"
 ```
+
+**Note:** When using numbers or booleans as values for variables, they must be
+enclosed in double quotes vars to be treated as strings, for more information see
+[substitution of numbers and booleans](#post-build-substitution-of-numbers-and-booleans).
 
 You can replicate the controller post-build substitutions locally using
 [kustomize](https://github.com/kubernetes-sigs/kustomize)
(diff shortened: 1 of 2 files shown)


souleb (approved):

stefanprodan:
The problem with this `"'` is that when the var comes as a string it will be invalid, for example some AWS IDs can be int or string.

ahothan:
not really a problem because for string, the variable value should just be defined in ConfigMap as usual without need for any quoting (or with the regular quoting)

stefanprodan:
@ahothan your solution doesn't work, using `"'${{number}'"` for metadata labels results in: `Invalid value: \"'123'\": a valid label must be a string`. I think it works for you, because you only do substitutions for Helm values.

ahothan:
we would not be using "'${number}'" at all

example of definitions in configmap that should all result in a string being rendered

number: "'100'"
condition: "'true'"
text: "anytext_that_is_not_int_or_bool"

and the reference from any flux yaml file (can be in values of HelmRelease or in any part of k8s manifest):

text1: ${text}
text2: ${number:='100'}
text3: ${cond}

rendered:

text1: anytext_that_is_not_int_or_bool
text2: "100"
text3: "true"

I have not tried it on a metadata label but it should work

ruiengana:
I think this works better...

var="100"
value: ${v="${var}"}

renders

value: "100"

better than having ${quote}${var}${quote} I guess...

nick4fake:
So... there is still no clean solution? This doesn't work for many cases, like substitutions from configmaps

stefanprodan:
@nick4fake I don't see why substitutions from ConfigMaps would not work with the solution present in the docs. You can configure it like this:

```yaml
spec:
  postBuild:
    substitute:
      quote: '"'
    substituteFrom:
      - kind: ConfigMap
        name: cluster-vars
```

pkit:
I'm not sure it works:
```
kind: Kustomization
....
patches:
  - patch: |-
      apiVersion: apps/v1
      kind: StatefulSet
      metadata:
        name: batch-worker
      spec:
        replicas: ${replicas}
        ....
```
```
.spec.replicas: expected numeric (int or float), got string
```
The funny thing is:
```
template:
          spec:
            containers:
              - name: batch-worker-container
                 env:
                    - name: WORKER_COUNT
                       value: ${quote}${replicas}${quote}
```
```
var substitution failed for 'batch-worker': YAMLToJSON: variable substitution failed: missing closing brace
```
And the values are:
```
  postBuild:
    substitute:
      quote: '"'
      replicas: "1"
```
So it fails as a string and as int too. lol

Issue #1133, opened 2024-04-09
Add `namePrefix` and `nameSuffix` to Flux Kustomization API


stefanprodan:
Allow setting name prefix and suffix in the Flux Kustomization API and apply the transformers during build.

PR #1134, opened 2024-04-09
Implement name prefix/suffix transformers

stefanprodan:
Closes: #1133 

`.spec.namePrefix` and `.spec.nameSuffix` are optional fields used to specify a prefix and suffix
to be added to the names of all the resources in the Kustomization.

```yaml
kind: Kustomization
metadata:
  name: app
spec:
  # ...omitted for brevity
  namePrefix: "prefix-"
  nameSuffix: "-suffix"
```

### DIFF
--- a/api/v1/kustomization_types.go
+++ b/api/v1/kustomization_types.go
@@ -99,6 +99,20 @@ type KustomizationSpec struct {
 	// +optional
 	HealthChecks []meta.NamespacedObjectKindReference `json:"healthChecks,omitempty"`
 
+	// NamePrefix will prefix the names of all managed resources.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=200
+	// +kubebuilder:validation:Optional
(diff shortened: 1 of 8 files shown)


api/v1/kustomization_types.go
```diff
@@ -99,6 +99,20 @@ type KustomizationSpec struct {
 	// +optional
 	HealthChecks []meta.NamespacedObjectKindReference `json:"healthChecks,omitempty"`
 
+	// NamePrefix will prefix the names of all managed resources.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=200
+	// +kubebuilder:validation:Optional
```
souleb, line 105: It seems that there isn't any difference between `// +kubebuilder:validation:Optional` and  `// +optional`

stefanprodan: This is how is set on others fields. If we decided to drop one, this is not the place to do it.

PR #1144, opened 2024-04-23
api: Refer condition type constants from `fluxcd/pkg/apis`

gdasson:
With reference to this PR: https://github.com/fluxcd/pkg/pull/723, I have referenced the kustomize constants from pkg now.


stefanprodan (commented):
Please undo the changes to v1beta1 and v1beta2, these API versions are considered readonly.

gdasson:
@stefanprodan : fixed as suggested.

PR #1241, opened 2024-09-11
feat: implement partial apply

cwrau:
this allows the user to create for example a HelmRelease for an operator and the corresponding resource in the same commit

TODO:

- [x] tests
- [ ] documentation


stefanprodan:
This is not something that I would be in favor of, we have `dependsOn` which allows you to properly apply things in order. 

This kind of change would require a RFC, you need to find a Flux maintainer to sponsor your proposal, see here the RFC process https://github.com/fluxcd/flux2/tree/main/rfcs.

PR #1283, opened 2024-11-08
Enable decryption of secrets generated by Kustomize components

vlasov-y:
Resolves #1235

PR #1286, opened 2024-11-12
Added decryption of Kustomize patches and refactor SOPS tests

vlasov-y:
Hi :hand: 
In addition to the previous PR, I've added decryption for the resource and patch files before the build process. I also found the test data for SOPS to be quite messy, so I restructured it entirely and streamlined the controller test code.

Followup: #1283


vlasov-y:
@stefanprodan tagging you because you have reviewed #1283 for me :sweat_smile:

internal/decryptor/decryptor.go
```diff
@@ -447,21 +453,45 @@ func (d *Decryptor) decryptKustomizationEnvSources(visited map[string]struct{})
...
+			if patch.Path == "" {
+				continue
+			}
+			// Determine the format for the patch, defaulting to YAML if not specified.
+			format := formatForPath(patch.Path)
+			// Visit the patch reference and attempt to decrypt it.
+			if err := visitRef(patch.Path, format, false); err != nil {
+				return err
+			}
+		}
```
stefanprodan, line 494: Why is this needed? Besides the secret/configMap generators, all the other YAML encrypted resources are decrypted in-memory before apply.

vlasov-y: I will give a short example that will be close to the case I have faced.
We have victoria-metrics HelmRelease in bases. It contains many services, also it is partially encrypted, because of basicAuth embedded in externalWrite URL.
In overlays, we have many values blocks for the same HelmRelease defined in separate merge patch files. One of them is alert-manager.sops.yaml that has Slack token.
When flux runs kustomization and merges them _before_ the decryption - it is failed on kubernetes apply stage, because SOPS metadata fields are mixed/overwriten etc. and hashes do not match anymore.
That is why I needed patches to be decreted **before** the kustomization processing.

vlasov-y: There is an example https://github.com/vlasov-y/kustomize-controller/tree/main/internal/controller/testdata/sops/patches

stefanprodan: Ok but for resources this will fail the moment you use remote bases no?

vlasov-y: I thought about that and that Is why I have added extra arg to visitRef

https://github.com/vlasov-y/kustomize-controller/blob/aefd7aa811cc7afa19a9005504393f03e3734bb8/internal/decryptor/decryptor.go#L418
https://github.com/vlasov-y/kustomize-controller/blob/aefd7aa811cc7afa19a9005504393f03e3734bb8/internal/decryptor/decryptor.go#L479

stefanprodan: Hmm wouldn't `securePaths(root, sourcePath)` fail before it gets to `ignoreNotRegular`? I think we need a test with remote bases to verify they get ignored, you can use `https://raw.githubusercontent.com/fluxcd/kustomize-controller/refs/heads/main/config/default/namespace.yaml`

stefanprodan: How about we decrypt only the patches, if I remember correctly those can't be remote

vlasov-y: > How about we decrypt only the patches, if I remember correctly those can't be remote

You are right, they cannot be.
I thought about that, but technically resources also can be encrypted... Maybe let it be for both?

I have added a fix and a test for remote base.
Unfortunately, namespace object is not created from test customization somewhy, so I make a trick and change both kind and name in patches :) That is a bit strange, but it works.
https://github.com/fluxcd/kustomize-controller/blob/ecbb3c91c012e9bfe7c070c7942ce930dcb77ca8/internal/controller/testdata/sops/remote/kustomization.yaml#L13

stefanprodan: > I thought about that, but technically resources also can be encrypted... Maybe let it be for both?

I would like to avoid touching the resources, these get decrypted after kustomize build. If we decrypt the patches, then the final secret should be fine, as the merged keys are in clean (those coming from patches) and the ones encrypted have the right SOPS metadata.

vlasov-y: > > I thought about that, but technically resources also can be encrypted... Maybe let it be for both?
> 
> I would like to avoid touching the resources, these get decrypted after kustomize build. If we decrypt the patches, then the final secret should be fine, as the merged keys are in clear text (those coming from patches) and the ones encrypted have the right SOPS metadata.

I have added another one test and example.
This bug, by the way, exists in the mainline as well.
If the resource is encrypted with `encrypted_regex` that overlap with any other key in the same manifest (added by patch) - sops decryption will fail. So resource has to be decrypted before the merge.

In that mix test we have a manifest with only part of it encrypted (by sops encrypted_regex "env")
https://github.com/fluxcd/kustomize-controller/blob/96206e3f82948d94c19b84d3a4b5b6698de687b6/internal/controller/testdata/sops/mix/pod.yaml#L1-L15

Also we have a patch, that adds new fields and that fields have key with name "env". SOPS will try to decrypt this field as well, but will fail

https://github.com/fluxcd/kustomize-controller/blob/fbd02cee72bae63fb2a2aca344f6abd8db6059d2/internal/controller/testdata/sops/mix/patch.yaml#L1-L15

vlasov-y: Hi @stefanprodan, sorry for bothering, just kindly reminding about this PR 😃

stefanprodan: This will need to wait for other maintainers to review, given the implications of going over all resources and running decryption on disk this could introduce CVEs in Flux.

vlasov-y: > This will need to wait for other maintainers to review, given the implications of going over all resources and running decryption on disk this could introduce CVEs in Flux.

Okay, whom can I ping here and ask to review?

stefanprodan: I'll add this to the dev meeting agenda for next week.

stefanprodan: Another thing about resources, these can point to other overlays, is your check function skipping dirs?

stefanprodan: Parsing all the resources has a major impact on the controller performance. The `sopsDecryptFile` loads the file in-memory to check if it needs decryption, while this was fine for env sources, with the current change this will explode the memory usage, as now we load every single manifests. There are Flux users with 10K+ manifests in a sigle Kustomization.

vlasov-y: > Parsing all the resources has a major impact on the controller performance. The `sopsDecryptFile` loads the file in-memory to check if it needs decryption, while this was fine for env sources, with the current change this will explode the memory usage and CPU usage (GC needs to work twice as much), as now we load every single manifests. There are Flux users with 10K+ manifests in a sigle Kustomization.

Okay, lets break this changes into parts. What if I remove decryption of resources from the PR, we merge decryption of patches and I am creating a new PR for resources only (if needed). Anyway, I am waiting for results of your next dev meeting. Can I attend?

vlasov-y: > Parsing all the resources has a major impact on the controller performance. The `sopsDecryptFile` loads the file in-memory to check if it needs decryption, while this was fine for env sources, with the current change this will explode the memory usage and CPU usage (GC needs to work twice as much), as now we load every single manifests. There are Flux users with 10K+ manifests in a sigle Kustomization.

There are some points I keep in mind thinking about the case:
1. https://github.com/fluxcd/kustomize-controller/pull/1286#discussion_r1840096168 - this issue exists anyway
2. but issue from p.1 can be bypassed if we move secrets from the resource to patches (assuming we perform patches decryption)
3. We load all resources to memory during the `kustomization build` anyway, do not we?
4. Can we decrypt resources basing on filename? like one matching `\.(sops|enc)` regex? And move that regex to some key in kustomize.toolkit.fluxcd.io/Kustomization/v1, wdyt?

stefanprodan: > We load all resources to memory during the kustomization build anyway, do not we?

Exactly, and this PR doubles the memory usage and the Go GC CPU usage.

> Can we decrypt resources basing on filename?

I'm not for changing the API to cope with a SOPS upstream issue, I'm Ok with decrypting the patches but after merge SOPS should handle mixed mode.

vlasov-y: > I'm not for changing the API to cope with a SOPS upstream issue, I'm Ok with decrypting the patches but after merge SOPS should handle mixed mode.

That is not a SOPS issues. SOPS decrypts the file that it encrypts without any issue, but we alter that file, we may adding keys that match `encrypted_regex`. Merging this PR with patches decryption only won't resolve mix test case completely, but will allow to easily bypass such limitation moving encrypted values to patch files. I think that will make life much easier already.

stefanprodan: @vlasov-y the next dev meeting is in 2 days, feel free to join https://fluxcd.io/community/#meetings

I understand now how SOPS fails, it's due to the regex. I think the best way forward is to decrypt the patches only and document this workaround, where the secret will have an empty stringData entry will all keys coming from patches.

vlasov-y: > @vlasov-y the next dev meeting is in 2 days, feel free to join https://fluxcd.io/community/#meetings

Thanks!

>  I think the best way forward is to decrypt the patches only and document this workaround

Thanks, agree. I have tried to benchmark decryption of resources and to find a better algorithm, tried streaming reading of files, but it is slower. Anyway, making patches being decrypted and document regex issue will be more than enough to live with Kustomization + SOPS in comfort.

vlasov-y: Updated documentation describing the case.
Removed decryption of resources.
Removed incompatible sops tests.

vlasov-y: @stefanprodan can we merge it?

docs/spec/v1/kustomizations.md
```diff
@@ -788,6 +790,86 @@ data:
   sops.vault-token: <BASE64>
 ```
 
+#### Important case: SOPS decryption encrypted_regex conflict
```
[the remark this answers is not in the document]
vlasov-y: Added same text here https://github.com/fluxcd/website/pull/2107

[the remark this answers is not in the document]
stefanprodan: Great! You can remove now this section and tidy up your commits, then it's good for merge.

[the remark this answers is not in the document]
vlasov-y: Done. I have squashed all my commits into one.

stefanprodan (approved):
LGTM

Thanks @vlasov-y 🏅

stefanprodan:
@vlasov-y I've published a release candidate of kustomize-controller that contains the changes in this PR. You use these images until Flux 2.5 is released:

- `ghcr.io/fluxcd/kustomize-controller:rc-9974a49b`
- `docker.io/fluxcd/kustomize-controller:rc-9974a49b`

PR #1313, opened 2024-12-13
Upgrading Helm API version to v2

luis-giraldo:


matheuscscp (approved):
Thanks @luis-giraldo!

Issue #662, opened 2022-05-25
Allow disabling a Kustomization's finalizer


julienduchesne:
Discussed here: https://cloud-native.slack.com/archives/CLAJ40HV3/p1653481382666929 but opening this for further discussion

The `prune` setting of Kustomizations has a two-fold behavior. It will delete resources which are removed from source, as is expected from a GitOps tool and it also deletes all resources managed by the Kustomization through its finalizer, when it is deleted.

We'd like to be able to keep pruning but disable the finalizer for the following reasons:

1. Prevent a cascading deletion in case someone deletes the `Kustomization`resource from the source (git). Flux manages itself on our clusters and it can be unexpected for users that deleting or modifying a single resource will cascade to others
2. Allow easier management of Kustomizations. They currently need to be suspended to be interacted with. Our usecase: we have a "global" Kustomization that manages other Kustomizations, which are located in each namespace. The global Kustomization is set to NOT prune to avoid a cascade deletion of the whole cluster on its deletion. If we got rid of the finalizer, we could manage everything with prune. This would allow us to do operations like renaming all Kustomizations in the cluster without any manual intervention
3. Safeguard against accidental deletion. I understand that production clusters should have access restrictions in place. However, in dev for example, a user may wish to disable Flux and delete the `Kustomization`, not knowing that this will delete all of the namespace's resources

I'd propose a new Kustomization setting: `finalizerMode` with three modes: `none` (disabled), `delabel` (remove flux labels on managed resources, `prune` (current mode)

makkes:
Please read https://fluxcd.io/docs/components/kustomize/kustomization/#garbage-collection.

stefanprodan:
Having annotations as a bandaid for not setting proper RBAC is not something I would consider. If users don't understand how Flux works, a better way to protect the cluster from destructive actions, would be to remove write & delete actions in RBAC, so expect for cluster admins, no other user could edit or delete Flux objects.

zxbyoyoyo:
I also encountered the same problem, when setting `prune=true` there are two behaviors,

1. deleting a single resource in Git will delete it from K8s;
2. when the Kustomize object is deleted, all managed resources will be deleted from K8s.

If I want to use the first behavior and don't need the second behavior, I need to add annotation `kustomize.toolkit.fluxcd.io/prune: disabled` to each resource to avoid not being recycled, which is very unfriendly , I also recommend providing a new field for the second behavior to support global control at the fluxcd level

[...]

PR #1314, opened 2024-12-15
Allow control of finalization garbage collection with `.spec.deletionPolicy`

erikgb:
The `prune` setting of Kustomizations has a two-fold behavior. When set to `true` FLux will:
1. Ensure stale managed resources are deleted.
2. Garbage collect managed resources when Kustomization is deleted (using a finalizer).

We would like to have 1, but not 2, and this is currently not possible. This PR introduces a new API field allowing control over finalization garbage collection independently of `prune`.

### Some background/motivation for this proposed enhancement:

We are using Flux to provision review environments in a multi-tenant environment (shared cluster), and our tenants are not allowed to create Kubernetes namespaces directly. Instead they have to use a custom namespace-scoped resource as a controller for the namespace (and RBAC). This indirect management of namespaces makes it hard to create Kustomizations inside review namespaces. To overcome this, we use the `targetNamespace` field. This works well when the review namespace is created. But when the review namespace is teared down by our review environment automation, there is a finalization race condition between Flux and our namespace controller. If Flux looses this race, the Kustomization is stuck in non-finalized state (because the namespace is gone) and requires manual intervention from a cluster-admin. This is something we definitely would like to avoid! Since the namespace is controlled and eventually deleted by our namespace controller, no resources are in fact orphaned if we disable FLux garbage collection.

Orphaning resources is generally a bad idea, but is sometimes required. 😉 
~~Once we agree on the naming, I will submit a PR to update the docs in the [website repo](https://github.com/fluxcd/website).~~

Close https://github.com/fluxcd/kustomize-controller/issues/662

### DIFF
--- a/api/v1/kustomization_types.go
+++ b/api/v1/kustomization_types.go
@@ -33,6 +33,10 @@ const (
 	MergeValue                = "Merge"
 	IfNotPresentValue         = "IfNotPresent"
 	IgnoreValue               = "Ignore"
+
+	DeletionPolicyMirrorPrune = "MirrorPrune"
+	DeletionPolicyDelete      = "Delete"
+	DeletionPolicyOrphan      = "Orphan"
 )
 
 // KustomizationSpec defines the configuration to calculate the desired state
@@ -94,6 +98,14 @@ type KustomizationSpec struct {
 	// Prune enables garbage collection.
 	// +required
 	Prune bool `json:"prune"`
+
+	// DeletionPolicy can be used to control garbage collection when this
+	// Kustomization is deleted. Valid values are ('MirrorPrune', 'Delete',
+	// 'Orphan'). 'MirrorPrune' mirrors the Prune field (orphan if false,
+	// delete if true). Defaults to 'MirrorPrune'.
+	// +kubebuilder:validation:Enum=MirrorPrune;Delete;Orphan
+	// +optional
+	DeletionPolicy string `json:"deletionPolicy,omitempty"`
 
 	// A list of resources to be included in the health assessment.
 	// +optional
@@ -285,6 +297,14 @@ func (in Kustomization) GetRetryInterval() time.Duration {
 // reconciled again.
 func (in Kustomization) GetRequeueAfter() time.Duration {
 	return in.Spec.Interval.Duration
+}
+
+// GetDeletionPolicy returns the deletion policy and default value if not specified.
+func (in Kustomization) GetDeletionPolicy() string {
+	if in.Spec.DeletionPolicy == "" {
+		return DeletionPolicyMirrorPrune
+	}
+	return in.Spec.DeletionPolicy
 }
 
 // GetDependsOn returns the list of dependencies across-namespaces.
--- a/docs/spec/v1/kustomizations.md
+++ b/docs/spec/v1/kustomizations.md
@@ -169,6 +169,39 @@ kustomize.toolkit.fluxcd.io/prune: disabled
 For details on how the controller tracks Kubernetes objects and determines what
 to garbage collect, see [`.status.inventory`](#inventory).
 
+### Deletion policy
+
+`.spec.deletionPolicy` is an optional field that allows control over
+garbage collection when a Kustomization object is deleted. The default behavior
+is to mirror the configuration of [`.spec.prune`](#prune).
+
+Valid values:
+
+- `MirrorPrune` (default) - The managed resources will be deleted if `prune` is
+  `true` and orphaned if `false`.
+- `Delete` - Ensure the managed resources are deleted before the Kustomization
+   is deleted.
+- `Orphan` - Leave the managed resources when the Kustomization is deleted.
+
+For special cases when the managed resources are removed by other means (e.g.
+the deletion of the namespace specified with
+[`.spec.targetNamespace`](#target-namespace)), you can set the deletion policy
+to `Orphan`:
+
+```yaml
+---
+apiVersion: kustomize.toolkit.fluxcd.io/v1
+kind: Kustomization
+metadata:
+  name: app
+  namespace: default
+spec:
+  # ...omitted for brevity
+  targetNamespace: app-namespace
+  prune: true
+  deletionPolicy: Orphan
+```
+
 ### Interval
 
 `.spec.interval` is a required field that specifies the interval at which the
--- a/internal/controller/kustomization_controller.go
+++ b/internal/controller/kustomization_controller.go
@@ -956,10 +956,17 @@ func (r *KustomizationReconciler) prune(ctx context.Context,
 	return false, nil
 }
 
+func finalizerShouldDeleteResources(obj *kustomizev1.Kustomization) bool {
+	if obj.GetDeletionPolicy() == kustomizev1.DeletionPolicyMirrorPrune {
+		return obj.Spec.Prune
+	}
+	return obj.Spec.DeletionPolicy == kustomizev1.DeletionPolicyDelete
+}
+
 func (r *KustomizationReconciler) finalize(ctx context.Context,
 	obj *kustomizev1.Kustomization) (ctrl.Result, error) {
 	log := ctrl.LoggerFrom(ctx)
-	if obj.Spec.Prune &&
+	if finalizerShouldDeleteResources(obj) &&
 		!obj.Spec.Suspend &&
 		obj.Status.Inventory != nil &&
 		obj.Status.Inventory.Entries != nil {
--- a/internal/controller/kustomization_deletion_policy_test.go
+++ b/internal/controller/kustomization_deletion_policy_test.go
@@ -1,0 +1,164 @@
+/*
+Copyright 2024 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/fluxcd/pkg/apis/meta"
+	"github.com/fluxcd/pkg/testserver"
+	sourcev1 "github.com/fluxcd/source-controller/api/v1"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1"
+)
+
+func TestKustomizationReconciler_DeletionPolicyDelete(t *testing.T) {
+	tests := []struct {
+		name           string
+		prune          bool
+		deletionPolicy string
+		wantDelete     bool
+	}{
+		{
+			name:           "should delete when deletionPolicy overrides pruning disabled",
+			prune:          false,
+			deletionPolicy: kustomizev1.DeletionPolicyDelete,
+			wantDelete:     true,
+		},
+		{
+			name:           "should delete when deletionPolicy mirrors prune and pruning enabled",
+			prune:          true,
+			deletionPolicy: kustomizev1.DeletionPolicyMirrorPrune,
+			wantDelete:     true,
+		},
+		{
+			name:           "should orphan when deletionPolicy overrides pruning enabled",
+			prune:          true,
+			deletionPolicy: kustomizev1.DeletionPolicyOrphan,
+			wantDelete:     false,
+		},
+		{
+			name:           "should orphan when deletionPolicy mirrors prune and pruning disabled",
+			prune:          false,
+			deletionPolicy: kustomizev1.DeletionPolicyMirrorPrune,
+			wantDelete:     false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			g := NewWithT(t)
+			id := "gc-" + randStringRunes(5)
+			revision := "v1.0.0"
+
+			err := createNamespace(id)
+			g.Expect(err).NotTo(HaveOccurred(), "failed to create test namespace")
+
+			err = createKubeConfigSecret(id)
+			g.Expect(err).NotTo(HaveOccurred(), "failed to create kubeconfig secret")
+
+			manifests := func(name string, data string) []testserver.File {
+				return []testserver.File{
+					{
+						Name: "config.yaml",
+						Body: fmt.Sprintf(`---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: %[1]s
+data:
+  key: "%[2]s"
+`, name, data),
+					},
+				}
+			}
+
+			artifact, err := testServer.ArtifactFromFiles(manifests(id, id))
+			g.Expect(err).NotTo(HaveOccurred())
+
+			repositoryName := types.NamespacedName{
+				Name:      fmt.Sprintf("gc-%s", randStringRunes(5)),
+				Namespace: id,
+			}
+
+			err = applyGitRepository(repositoryName, artifact, revision)
+			g.Expect(err).NotTo(HaveOccurred())
+
+			kustomizationKey := types.NamespacedName{
+				Name:      fmt.Sprintf("gc-%s", randStringRunes(5)),
+				Namespace: id,
+			}
+			kustomization := &kustomizev1.Kustomization{
+				ObjectMeta: metav1.ObjectMeta{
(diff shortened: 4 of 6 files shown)


api/v1/kustomization_types.go
```diff
@@ -95,6 +99,13 @@ type KustomizationSpec struct {
 	// +required
 	Prune bool `json:"prune"`
 
+	// DeletionPolicy can be used to control garbage collection when this
+	// Kustomization is deleted. Will mirror value of Prune if unspecified.
+	// Valid values are ('MirrorPrune' (default), 'Delete')
```
stefanprodan, line 104:
```suggestion
	// Valid values are ('MirrorPrune' (default), 'Delete', 'Orphan')
```

stefanprodan (commented):
Please add a new section to the API docs right after https://github.com/fluxcd/kustomize-controller/blob/main/docs/spec/v1/kustomizations.md#prune

internal/controller/kustomization_controller.go
```diff
@@ -956,10 +956,17 @@ func (r *KustomizationReconciler) prune(ctx context.Context,
 	return false, nil
 }
 
+func finalizerShouldDeleteDependents(obj *kustomizev1.Kustomization) bool {
```
stefanprodan, line 959:
```suggestion
func finalizerShouldDeleteResources(obj *kustomizev1.Kustomization) bool {
```

internal/controller/kustomization_controller.go
```diff
@@ -956,10 +956,17 @@ func (r *KustomizationReconciler) prune(ctx context.Context,
...
+	if obj.Spec.DeletionPolicy == nil || *obj.Spec.DeletionPolicy == kustomizev1.DeletionPolicyMirrorPrune {
+		return obj.Spec.Prune
+	}
```
stefanprodan, line 962: Please add a function to the API package called `GetDeletionPolicy()`, similar to `GetRetryInterval()`, that returns the default value when the field value is set to string empty. Then change the logic here to `if obj.GetDeletionPolicy() == kustomizev1.DeletionPolicyMirrorPrune`.

api/v1/kustomization_types.go
```diff
@@ -95,6 +99,13 @@ type KustomizationSpec struct {
...
+	// +kubebuilder:validation:Enum=MirrorPrune;Delete;Orphan
+	// +optional
+	DeletionPolicy *string `json:"deletionPolicy,omitempty"`
```
stefanprodan, line 107:
```suggestion
	DeletionPolicy string `json:"deletionPolicy,omitempty"`
```

No need to set a pointer to a string.

internal/controller/kustomization_deletion_policy_test.go
```diff
@@ -0,0 +1,220 @@
...
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1"
+)
+
+func TestKustomizationReconciler_DeletionPolicyDelete(t *testing.T) {
```
matheuscscp, line 39: Please test also `MirrorPrune` with the different possible values for `prune`.

matheuscscp (commented):
Thanks for this contribution! I think it's a really valuable one.

internal/controller/kustomization_deletion_policy_test.go
erikgb: @matheuscscp, will do! Are there any recommendations or established patterns in this project to avoid so much duplicated test code? If not, I am pretty good at copy-modify. 😉

internal/controller/kustomization_controller.go
erikgb: This should be fixed now. I will write docs and add more tests before rebasing and squashing everything into a single commit.

internal/controller/kustomization_deletion_policy_test.go
matheuscscp: We always try to do the usual table testing widespread in Go projects. However if the test cases are too different then it's OK to have separate `func TestXXX` tests for them. In this specific case you could for example group together the cases where deletion is expected, and make another group where orphaning is expected.

erikgb: Thanks! Do you have a good example of how you write table-driven tests in Flux projects? Assuming there are multiple ways to do this.

docs/spec/v1/kustomizations.md
```diff
@@ -169,6 +169,46 @@ kustomize.toolkit.fluxcd.io/prune: disabled
...
+But if another garbage collection mechanism is in play, it might make sense
+to orphan resources in the context of Flux.
+
+One example is in multi-tenant clusters where a tenant might not be allowed to
+provision namespaces directly, but have to do it indirectly through a custom
+resource type.
+This setup requires use of [`.spec.targetNamespace`](#target-namespace) to get
+anything into a new namespace. When deleting both the Kustomization and the
+namespace controller resource, there is a risk of the namespace being gone
+before Flux initializes the Kustomization finalization.
```
tenstad, line 194: Time of completion is what's relevant?

```suggestion
before Flux completes the Kustomization finalization.
```

api/v1/kustomization_types.go
```diff
@@ -95,6 +99,13 @@ type KustomizationSpec struct {
 	// +required
 	Prune bool `json:"prune"`
 
+	// DeletionPolicy can be used to control garbage collection when this
+	// Kustomization is deleted. Will mirror value of Prune if unspecified.
+	// Valid values are ('MirrorPrune' (default), 'Delete', 'Orphan')
```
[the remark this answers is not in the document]
tenstad: The mapping could be described in the docs instead, that might be helpful (and sufficient).

internal/controller/kustomization_deletion_policy_test.go
erikgb: Thanks! I have updated the tests now. I have some issues running the tests on my workstation (unrelated to this change). I get the following error:

````
panic: Failed to create Vault instance: could not connect to docker: reached retry deadline: Get "http://127.0.0.1:32769/v1/sys/init": dial tcp 127.0.0.1:32769: connect: connection refused
````

So would it be possible to approve workflows to check the tests (and CI in general)?

I also want to rebase this PR before merge. But I'll wait until it's "approved" (almost ready to merge).

File: api/v1/kustomization_types.go (hunk 7 above)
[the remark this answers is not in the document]
Comment (erikgb) [reply api/v1/kustomization_types.go:104]: > I do like the convention of a final Defaults to statement

I agree in general. But the default is a bit "special" in this case.

File: docs/spec/v1/kustomizations.md [hunk 8]
```diff
@@ -169,6 +169,46 @@ kustomize.toolkit.fluxcd.io/prune: disabled
...
+
+`.spec.deletionPolicy` is an optional field that is allowing control over
+garbage collection when a Kustomization object is deleted. The default behavior
+is to mirror the value of [`.spec.prune`](#prune).
+
+Valid values include:
+
+- `Delete` - Ensure the managed resources are deleted before the Kustomization
+   is finalized.
+- `Orphan` - Leave the managed resources when the Kustomization is deleted.
```
stefanprodan, line 182: We need to add MirrorPrune first and explain what it does.

docs/spec/v1/kustomizations.md
```diff
@@ -169,6 +169,46 @@ kustomize.toolkit.fluxcd.io/prune: disabled
...
+
+One example is in multi-tenant clusters where a tenant might not be allowed to
+provision namespaces directly, but have to do it indirectly through a custom
+resource type.
+This setup requires use of [`.spec.targetNamespace`](#target-namespace) to get
+anything into a new namespace. When deleting both the Kustomization and the
+namespace controller resource, there is a risk of the namespace being gone
+before Flux completes the Kustomization finalization.
+To avoid the Kustomization from being stuck in a non-finalized state, it can
+make sense to orphan managed resources:
```
stefanprodan, line 196:
```suggestion
For special cases when the managed resources are removed by other means (e.g. the deletion of the namespace specified with `.spec.targetNamespace`), you can set the deletion policy to `Orphan`:
```

api/v1/kustomization_types.go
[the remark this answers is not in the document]
tenstad: Don't see the harm in writing the default _value_ (not behaviour) when the sentence before it describes the behavior of that value.

I think the connection between the two current statements can be improved. _mirror value of Prune_ relates to _MirrorPrune_ and _unspecified_ relates to _default_, but it is not that explicit, which can lead to confusion.

> Will mirror value of Prune if unspecified.

> Valid values are ('MirrorPrune' (default), 'Delete', 'Orphan')

docs/spec/v1/kustomizations.md
erikgb: Made an attempt. Please take a look.

internal/controller/kustomization_deletion_policy_test.go
```diff
@@ -0,0 +1,255 @@
+/*
+Copyright 2021 The Flux authors
```
stefanprodan, line 2:
```suggestion
Copyright 2024 The Flux authors
```

erikgb:
Ok, it seems like we are approaching something here. 😸  I have just rebased and squashed all commits. The PR should be ready to merge if you are satisfied with the proposed changes.

docs/spec/v1/kustomizations.md
```diff
@@ -169,6 +169,39 @@ kustomize.toolkit.fluxcd.io/prune: disabled
 For details on how the controller tracks Kubernetes objects and determines what
 to garbage collect, see [`.status.inventory`](#inventory).
 
+### Deletion policy
+
+`.spec.deletionPolicy` is an optional field that is allowing control over
```
tenstad, line 174:
```suggestion
`.spec.deletionPolicy` is an optional field that allows control over
```

docs/spec/v1/kustomizations.md
```diff
@@ -169,6 +169,39 @@ kustomize.toolkit.fluxcd.io/prune: disabled
...
+garbage collection when a Kustomization object is deleted. The default behavior
+is to mirror the value of [`.spec.prune`](#prune).
```
tenstad, line 176: Not totally sure about this one, but it's not the concrete true/false value that is mirrored. Should that be addressed here?

```suggestion
is to mirror the configuration of [`.spec.prune`](#prune).
```

docs/spec/v1/kustomizations.md
```diff
@@ -169,6 +169,39 @@ kustomize.toolkit.fluxcd.io/prune: disabled
...
+
+Valid values:
+
+- `MirrorPrune` (default) - The managed resources will be deleted if `prune` is
+  `true` and orphaned if `false`.
+- `Delete` - Ensure the managed resources are deleted before the Kustomization
+   is finalized.
+- `Orphan` - Leave the managed resources when the Kustomization is deleted.
```
tenstad, line 184: Keep it simple and non-technical, and equal between the two sentances?
```suggestion
   is deleted.
- `Orphan` - Leave the managed resources when the Kustomization is deleted.
```

docs/spec/v1/kustomizations.md
```diff
@@ -169,6 +169,39 @@ kustomize.toolkit.fluxcd.io/prune: disabled
...
+[`.spec.targetNamespace`](#target-namespace)), you can set the deletion policy
+to `Orphan`:
+
+```yaml
+---
+apiVersion: kustomize.toolkit.fluxcd.io/v1
+kind: Kustomization
+metadata:
+  name: app
+  namespace: tenant
```
tenstad, line 197: Not relevant here, might as well remove?
```suggestion
```

api/v1/kustomization_types.go
```diff
@@ -95,6 +99,14 @@ type KustomizationSpec struct {
 	// +required
 	Prune bool `json:"prune"`
 
+	// DeletionPolicy can be used to control garbage collection when this
+	// Kustomization is deleted. Valid values are ('MirrorPrune', 'Delete',
+	// 'Orphan'). 'MirrorPrune' mirrors the Prune field (false = 'Ophan',
+	// true = 'Delete'). Defaults to 'MirrorPrune'.
```
tenstad, line 105: If becoming a bit too technical and type-specific, writing it out in text can also work. Up to you!

```suggestion
	// 'Orphan'). 'MirrorPrune' mirrors the Prune field (orphan if false,
	// delete if true). Defaults to 'MirrorPrune'.
```

tenstad (commented):
Looking good! Just some minor nitpicks/thouhts regarding text.

docs/spec/v1/kustomizations.md
matheuscscp: I'm ok with removing it but I'd prefer to keep it, otherwise a new user reading this quickly might think `Kustomization` is cluster-scoped. :shrug:

docs/spec/v1/kustomizations.md
```diff
@@ -169,6 +169,39 @@ kustomize.toolkit.fluxcd.io/prune: disabled
...
+spec:
+  # ...omitted for brevity
```
matheuscscp, line 199:
```suggestion
```

internal/controller/kustomization_deletion_policy_test.go
```diff
@@ -0,0 +1,255 @@
...
+
+			resultK := &kustomizev1.Kustomization{}
+			resultConfig := &corev1.ConfigMap{}
+
+			g.Eventually(func() bool {
+				_ = k8sClient.Get(context.Background(), client.ObjectKeyFromObject(kustomization), resultK)
+				return resultK.Status.LastAppliedRevision == revision
+			}, timeout, time.Second).Should(BeTrue())
+
+			t.Run("creates objects", func(t *testing.T) {
```
matheuscscp, line 130: I believe we can move all these assertions outside the test case. The test cases should be only the ones in the table on the beginning.

internal/controller/kustomization_deletion_policy_test.go
```diff
@@ -0,0 +1,255 @@
...
+				g.Expect(k8sClient.Get(context.Background(), types.NamespacedName{Name: id, Namespace: id}, resultConfig)).Should(Succeed())
+			})
+
+			t.Run("deletes managed objects", func(t *testing.T) {
```
matheuscscp, line 134: Ditto.

internal/controller/kustomization_deletion_policy_test.go
```diff
@@ -0,0 +1,255 @@
...
+
+			resultK := &kustomizev1.Kustomization{}
+			resultConfig := &corev1.ConfigMap{}
+
+			g.Eventually(func() bool {
+				_ = k8sClient.Get(context.Background(), client.ObjectKeyFromObject(kustomization), resultK)
+				return resultK.Status.LastAppliedRevision == revision
+			}, timeout, time.Second).Should(BeTrue())
+
+			t.Run("creates objects", func(t *testing.T) {
```
matheuscscp, line 240: Ditto.

internal/controller/kustomization_deletion_policy_test.go
```diff
@@ -0,0 +1,255 @@
...
+				g.Expect(k8sClient.Get(context.Background(), types.NamespacedName{Name: id, Namespace: id}, resultConfig)).Should(Succeed())
+			})
+
+			t.Run("orphans managed objects", func(t *testing.T) {
```
matheuscscp, line 244: Ditto.

internal/controller/kustomization_deletion_policy_test.go
```diff
@@ -0,0 +1,255 @@
...
+				}, timeout, time.Second).Should(BeTrue())
+
+				err = k8sClient.Get(context.Background(), client.ObjectKeyFromObject(resultConfig), resultConfig)
+				g.Expect(apierrors.IsNotFound(err)).To(BeTrue())
+			})
+		})
+	}
+}
+
+func TestKustomizationReconciler_DeletionPolicyOrphan(t *testing.T) {
```
matheuscscp, line 148: Looking at these two top-level tests I think the difference between them is not huge. I think we could move all the cases into a single top-level test function and add a boolean to the test case struct to indicate whether deletion is expected or not and have and if-else that tests both cases at the end of the test code. Sorry for misleading you on the opposite direction before!

matheuscscp (commented):
Great stuff, I'm ready to give my approval after all the open discussions are resolved!

internal/controller/kustomization_deletion_policy_test.go
erikgb: No, problem. I will fix it! The same thought struck me when changing this.

docs/spec/v1/kustomizations.md
erikgb: This was copied from another example in the docs. Without it, a reader might wonder if this is a valid Kustomization.

internal/controller/kustomization_deletion_policy_test.go
erikgb: Done

erikgb:
@matheuscscp All discussions should now be resolved. Can you please approve CI to see if I made any mistakes in the test refactoring?

matheuscscp (approved):
Great contribution! Thanks @erikgb

tenstad (approved):

stefanprodan (approved):
LGTM

Thanks @erikgb 🏅

PR #1338, opened 2025-01-21
Add OCI origin revision to events

matheuscscp:
Part of: https://github.com/fluxcd/flux2/issues/5148

I have tested this e2e using GitHub:

oci-bad-commit-status

oci-good-commit-status

### DIFF
--- a/internal/controller/kustomization_controller.go
+++ b/internal/controller/kustomization_controller.go
@@ -193,7 +193,7 @@ func (r *KustomizationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 				time.Since(reconcileStart).String(),
 				obj.Spec.Interval.Duration.String())
 			log.Info(msg, "revision", obj.Status.LastAttemptedRevision)
-			r.event(obj, obj.Status.LastAppliedRevision, eventv1.EventSeverityInfo, msg,
+			r.event(obj, obj.Status.LastAppliedRevision, obj.Status.LastAppliedOriginRevision, eventv1.EventSeverityInfo, msg,
 				map[string]string{
 					kustomizev1.GroupVersion.Group + "/" + eventv1.MetaCommitStatusKey: eventv1.MetaCommitStatusUpdateValue,
 				})
@@ -234,7 +234,7 @@ func (r *KustomizationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		if acl.IsAccessDenied(err) {
 			conditions.MarkFalse(obj, meta.ReadyCondition, apiacl.AccessDeniedReason, "%s", err)
 			log.Error(err, "Access denied to cross-namespace source")
-			r.event(obj, "unknown", eventv1.EventSeverityError, err.Error(), nil)
+			r.event(obj, "", "", eventv1.EventSeverityError, err.Error(), nil)
 			return ctrl.Result{RequeueAfter: obj.GetRetryInterval()}, nil
 		}
 
@@ -249,6 +249,8 @@ func (r *KustomizationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		log.Info(msg)
 		return ctrl.Result{RequeueAfter: r.requeueDependency}, nil
 	}
+	revision := artifactSource.GetArtifact().Revision
+	originRevision := getOriginRevision(artifactSource)
 
 	// Check dependencies and requeue the reconciliation if the check fails.
 	if len(obj.Spec.DependsOn) > 0 {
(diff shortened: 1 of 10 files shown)


internal/controller/kustomization_controller.go
```diff
@@ -965,6 +972,11 @@ func finalizerShouldDeleteResources(obj *kustomizev1.Kustomization) bool {
 
 func (r *KustomizationReconciler) finalize(ctx context.Context,
 	obj *kustomizev1.Kustomization) (ctrl.Result, error) {
+
+	// Here the origin revision is not known because the artifact was not fetched,
+	// it's too soon in the Reconcile() method.
+	const originRevision = ""
```
matheuscscp, line 978: Does this sound right?

matheuscscp (commented):
@stefanprodan Please take a look at the comments :pray:

internal/controller/kustomization_controller.go
stefanprodan: If we add `status.LastAppliedOriginRevision` than we can just read it if it's there, like we do for revision now.

internal/controller/kustomization_controller.go
```diff
@@ -193,7 +193,7 @@ func (r *KustomizationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 				time.Since(reconcileStart).String(),
 				obj.Spec.Interval.Duration.String())
 			log.Info(msg, "revision", obj.Status.LastAttemptedRevision)
-			r.event(obj, obj.Status.LastAppliedRevision, eventv1.EventSeverityInfo, msg,
+			r.event(obj, obj.Status.LastAppliedRevision, "not-sure-what-to-put-here", eventv1.EventSeverityInfo, msg,
```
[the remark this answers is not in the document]
matheuscscp: Or rather, defaults to empty, and has the `omitempty` JSON annotation in the struct.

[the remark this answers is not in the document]
matheuscscp: fixed, thanks

internal/controller/kustomization_controller.go
matheuscscp: fixed, thanks

internal/controller/kustomization_controller.go
```diff
@@ -234,7 +234,7 @@ func (r *KustomizationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		if acl.IsAccessDenied(err) {
 			conditions.MarkFalse(obj, meta.ReadyCondition, apiacl.AccessDeniedReason, "%s", err)
 			log.Error(err, "Access denied to cross-namespace source")
-			r.event(obj, "unknown", eventv1.EventSeverityError, err.Error(), nil)
+			r.event(obj, "unknown", "unknown", eventv1.EventSeverityError, err.Error(), nil)
```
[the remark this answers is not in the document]
matheuscscp: fixed, thanks

internal/controller/kustomization_controller.go
```diff
@@ -1108,3 +1119,23 @@ func (r *KustomizationReconciler) patch(ctx context.Context,
...
+// or the empty string if it's not present, or if the artifact itself
+// is not present.
+func getOriginRevision(src sourcev1.Source) string {
+	a := src.GetArtifact()
+	if a == nil {
+		return ""
+	}
+
+	for _, key := range []string{
+		"org.opencontainers.image.revision",
```
stefanprodan, line 1133: Let's add this to the API package as:

```go
const ArtifactOriginRevisionAnnotation = "org.opencontainers.image.revision"
```

And instead of a for loop, we just do `a.Metadata[api.ArtifactOriginRevisionAnnotation]`

matheuscscp: And keep the loop...

stefanprodan: What's point of having a loop over an array with a single entry?

stefanprodan: Also the annotation has nothing to do with SC, this is convention used in the CLI now in KC. SC does not look at annotations it just copies whatever is in the upstream artifact.

matheuscscp: > What's point of having a loop over an array with a single entry?

Communicating the possibility of more items being added to that array in the future. But I agree that, since this is an internal implementation detail that can change later, we can remove the loop for now.

> Also the annotation has nothing to do with SC, this is a convention used in the CLI now in KC. SC does not look at annotations, it just copies whatever is in the upstream artifact.

I don't think this constant belongs in KC's API. I agree that this is an API, but certainly not one that KC is implementing/providing. KC is just the consumer of this API. If source-controller is not the right place, maybe somewhere else is, e.g. pkg/apis/oci.

stefanprodan: We have it pkg/oci but I don’t want to import a huge package here just for one static var. Let’s set the static var in the controller internal pkg then.

matheuscscp: fixed, thanks

internal/controller/kustomization_origin_revision_test.go
```diff
@@ -0,0 +1,123 @@
+/*
+Copyright 2022 The Flux authors
```
stefanprodan, line 2:
```suggestion
Copyright 2025 The Flux authors
```

matheuscscp: fixed, thanks

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🥇

PR #1344, opened 2025-01-24
[RFC-0009] Add CEL custom healthchecks

matheuscscp:
Part of: https://github.com/fluxcd/flux2/pull/5151


api/v1/kustomization_types.go
```diff
@@ -179,6 +179,11 @@ type KustomizationSpec struct {
 	// Components specifies relative paths to specifications of other Components.
 	// +optional
 	Components []string `json:"components,omitempty"`
+
+	// HealthCheckExprs is a list of healthcheck expressions for evaluating the
+	// health of custom resources according to RFC 0009.
```
stefanprodan, line 184:
```suggestion
	// health of custom resources using Common Expression Language (CEL).
```

matheuscscp: fixed, thanks

internal/controller/kustomization_wait_test.go
```diff
@@ -275,3 +276,87 @@ parameters:
...
+			Interval: metav1.Duration{Duration: 2 * time.Minute},
+			Path:     "./",
+			SourceRef: kustomizev1.CrossNamespaceSourceReference{
+				Name:      repositoryName.Name,
+				Namespace: repositoryName.Namespace,
+				Kind:      sourcev1.GitRepositoryKind,
+			},
+			TargetNamespace: id,
+			Prune:           true,
+			Timeout:         &metav1.Duration{Duration: time.Second},
```
matheuscscp, line 335: I copied this 1sec timeout from the test right above. This timeout doesn't work, because:

https://github.com/fluxcd/kustomize-controller/blob/59676860b8cb5bb25db94efc7dea8b75f5966998/api/v1/kustomization_types.go#L285-L294

And the result is that this new test takes at least 30secs to run.

go.mod
```diff
@@ -20,11 +20,11 @@ require (
...
 	github.com/fluxcd/pkg/apis/acl v0.5.0
 	github.com/fluxcd/pkg/apis/event v0.15.0
-	github.com/fluxcd/pkg/apis/kustomize v1.8.0
-	github.com/fluxcd/pkg/apis/meta v1.9.0
+	github.com/fluxcd/pkg/apis/kustomize v1.8.1-0.20250123112748-c55030369b58
+	github.com/fluxcd/pkg/apis/meta v1.9.1-0.20250127154341-0b0eba6810f4
 	github.com/fluxcd/pkg/http/fetch v0.14.0
 	github.com/fluxcd/pkg/kustomize v1.15.0
-	github.com/fluxcd/pkg/runtime v0.52.0
+	github.com/fluxcd/pkg/runtime v0.52.1-0.20250127144843-68389f802c21
```
matheuscscp, line 27: Need to import the proper tag after https://github.com/fluxcd/pkg/pull/858 is merged.

internal/controller/kustomization_controller.go
```diff
@@ -978,12 +987,18 @@ func (r *KustomizationReconciler) finalize(ctx context.Context,
 		!obj.Spec.Suspend &&
 		obj.Status.Inventory != nil &&
 		obj.Status.Inventory.Entries != nil {
+
 		objects, _ := inventory.List(obj.Status.Inventory)
 
+		statusPoller, pollingOpts, err := r.getPollerAndOptions(obj)
+		if err != nil {
+			return ctrl.Result{}, err
```
matheuscscp, line 995: Not sure if I should be returning an error here, why is the `inventory.List()` call right above swallowing the error?

matheuscscp: Yeah I think we should not even call `getPollerAndOptions` here and just use the default poller+options from the reconciler.

go.mod
```diff
@@ -14,34 +14,34 @@ require (
...
+	github.com/spf13/pflag v1.0.6
+	golang.org/x/net v0.34.0
+	k8s.io/api v0.32.1
+	k8s.io/apimachinery v0.32.1
+	k8s.io/client-go v0.32.1
 	k8s.io/utils v0.0.0-20241210054802-24370beab758
-	sigs.k8s.io/controller-runtime v0.19.3
-	sigs.k8s.io/kustomize/api v0.18.0
+	sigs.k8s.io/controller-runtime v0.20.1
+	sigs.k8s.io/kustomize/api v0.19.0
```
stefanprodan, line 44: This is pinned below so you need to bump Kustomize packages there, make sure to update the comment as well

matheuscscp: fixed, thanks

docs/spec/v1/kustomizations.md
```diff
@@ -339,6 +339,67 @@ spec:
...
+  prune: true
+  sourceRef:
+    kind: GitRepository
+    name: flux-system
+  timeout: 1m
+  wait: true # Tells the controller to wait for all resources to be ready by performing health checks.
+  healthCheckExprs:
+    - apiVersion: bitnami.com/v1alpha1
+      kind: SealedSecret
+      inProgress: has(status.observedGeneration) && status.observedGeneration != metadata.generation
```
stefanprodan, line 387: Remove `inProgress` to be in sync with the library

api/v1/kustomization_types.go
```diff
@@ -179,6 +179,11 @@ type KustomizationSpec struct {
 	// Components specifies relative paths to specifications of other Components.
 	// +optional
 	Components []string `json:"components,omitempty"`
+
+	// HealthCheckExprs is a list of healthcheck expressions for evaluating the
+	// health of custom resources using Common Expression Language (CEL).
```
stefanprodan, line 184:
```suggestion
	// HealthCheckExprs is a list of healthcheck expressions for evaluating the
	// health of custom resources using Common Expression Language (CEL).
	// The expressions are evaluated only when .spec.wait or .spec.healthCheck are specified.
```

docs/spec/v1/kustomizations.md
matheuscscp: fixed, thanks

api/v1/kustomization_types.go
matheuscscp: fixed, thanks

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🥇

PR #1374, opened 2025-02-24
Revert controller-runtime to v0.19

stefanprodan:
Revert controller-runtime from v0.20 to v0.19.  The Kubernetes [Aggregated Discovery](https://github.com/kubernetes/enhancements/issues/3352)  was enabled in [controller-runtime#2901](https://github.com/kubernetes-sigs/controller-runtime/pull/2901) and looks like this feature is buggy resulting in old API versions being returned as the preferred version randomly. 

This was observed on lots of cluster by Flux users, including on Kubernetes 1.32, 1.31 and OpenShift 4.16.

RC images:

- `ghcr.io/fluxcd/kustomize-controller:rc-deaabaa6`
- `docker.io/fluxcd/kustomize-controller:rc-deaabaa6`


matheuscscp (approved):

erikgb:
I have tested the RC image, and I still see some issues - but at a much lower scale. Example from an ephemeral cluster:

````
$ k logs -n flux-system kustomize-controller-548d7b9444-9ndxq | grep deprecated
{"level":"info","ts":"2025-02-24T15:10:29.814Z","logger":"KubeAPIWarningLogger","msg":"kyverno.io/v2beta1 ClusterCleanupPolicy is deprecated; use kyverno.io/v2 ClusterCleanupPolicy"}
````

stefanprodan:
@erikgb that warning should also be in all the other versions of Flux, we've reverted to what we've been using 2.4

erikgb:
Yes, you are correct. But at a much lower scale. Here is an extract from our Splunk installation:
image

Almost all the log events are from after we upgraded to Flux 2.5.

Issue #1373, opened 2025-02-24
Flux 2.5 seems to use old/deprecated versions of API


erikgb:
After upgrading to Flux 2.5 in one of our early clusters we observed a new type of error. Example:

````
health check failed after 3m0.016091887s: timeout waiting for: [SubNamespace/strm/strm-kafka-dev2 status: 'Unknown': looking up status.observedGeneration from resource: .status.observedGeneration accessor error: ok is of the type string, expected map[string]interface, SubNamespace/strm/strm-cluster-monitoring status: 'Unknown': looking up status.observedGeneration from resource: .status.observedGeneration accessor error: ok is of the type string, expected map[string]interface, SubNamespace/strm/strm-kafka status: 'Unknown': looking up status.observedGeneration from resource: .status.observedGeneration accessor error: ok is of the type string, expected map[string]interface]
````

It seems like this error occurs because Flux somehow uses an old/deprecated version of the [SubNamespace API](https://github.com/cybozu-go/accurate/blob/main/config/crd/bases/accurate.cybozu.com_subnamespaces.yaml). This wasn't an issue on Flux 2.4 and the error seems to occur a bit randomly.

After investigating this a bit more, it seems like kustomize-controller is also using other old/deprecated API versions. This example is from a bootstrap of our "fleet" repo on an ephemeral k3s cluster:

````
$ k logs -n flux-system kustomize-controller-5995c9c7f6-dg9qc | grep deprecated
{"level":"info","ts":"2025-02-24T11:12:10.399Z","logger":"KubeAPIWarningLogger","msg":"external-secrets.io/v1alpha1 ExternalSecret is deprecated; use external-secrets.io/v1beta1 ExternalSecret"}
{"level":"info","ts":"2025-02-24T11:12:10.404Z","logger":"KubeAPIWarningLogger","msg":"v1beta2 Kustomization is deprecated, upgrade to v1"}
{"level":"info","ts":"2025-02-24T11:12:15.402Z","logger":"KubeAPIWarningLogger","msg":"external-secrets.io/v1alpha1 ExternalSecret is deprecated; use external-secrets.io/v1beta1 ExternalSecret"}
{"level":"info","ts":"2025-02-24T11:12:15.407Z","logger":"KubeAPIWarningLogger","msg":"v1beta2 Kustomization is deprecated, upgrade to v1"}
[...]

PR #1377, opened 2025-02-25
Use the dynamic RESTMapper from controller-runtime v0.19

stefanprodan:
This PR replaces the RESTMapper shipped in controller-runtime v0.20 for health check operations. 

The Kubernetes [Aggregated Discovery](https://github.com/kubernetes/enhancements/issues/3352)  was enabled in [controller-runtime#2901](https://github.com/kubernetes-sigs/controller-runtime/pull/2901) and looks like this feature is buggy resulting in old API versions being returned as the preferred version randomly. While running the health checks, the controller would query old versions resulting in spurious logs about deprecated versions and in some cases, the health check would even fail.

This was observed on lots of cluster by Flux users, including on Kubernetes 1.32, 1.31 and OpenShift 4.16.

Depends on: https://github.com/fluxcd/pkg/pull/873
Supersedes: https://github.com/fluxcd/kustomize-controller/pull/1374
Fixes: https://github.com/fluxcd/kustomize-controller/issues/1373

Test images:
- `ghcr.io/fluxcd/kustomize-controller:rc-8b6f4bc6`
- `docker.io/fluxcd/kustomize-controller:rc-8b6f4bc6`


internal/controller/kustomization_wait_test.go
```diff
@@ -325,6 +328,11 @@ data: {}
 		Spec: kustomizev1.KustomizationSpec{
 			Interval: metav1.Duration{Duration: 2 * time.Minute},
 			Path:     "./",
+			KubeConfig: &meta.KubeConfigReference{
+				SecretRef: meta.SecretKeyReference{
+					Name: "kubeconfig",
```
matheuscscp, line 333: Just curious, why was this needed?

matheuscscp (approved):

internal/controller/kustomization_wait_test.go
stefanprodan: This is how all KC tests run, using the envTest kubeConfig generated at startup. Previously we were bypassing impersonation...

PR #1396, opened 2025-03-17
Skip emitting events for suspended Kustomizations

norman-zon:
There are edge cases where a reconciliation can be triggered on a suspended Kustomization, mainly through changes in a source. In these cases a "no-op" reconciliation occurs, which can be confusing for the user. This PR prevents the controller from updating the status and issuing events.

Fix: https://github.com/fluxcd/flux2/issues/5174

### DIFF
--- a/internal/controller/kustomization_controller.go
+++ b/internal/controller/kustomization_controller.go
@@ -189,6 +189,11 @@ func (r *KustomizationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		// Record Prometheus metrics.
 		r.Metrics.RecordDuration(ctx, obj, reconcileStart)
 
+		// Do not proceed if the Kustomization is suspended
+		if obj.Spec.Suspend {
+			return
+		}
+
 		// Log and emit success event.
 		if conditions.IsReady(obj) {
 			msg := fmt.Sprintf("Reconciliation finished in %s, next run in %s",


norman-zon:
If you want to have a test for this, I think I need some help how to reproduce this case in testing.

internal/controller/kustomization_controller.go
```diff
@@ -191,9 +191,15 @@ func (r *KustomizationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 		// Log and emit success event.
 		if conditions.IsReady(obj) {
-			msg := fmt.Sprintf("Reconciliation finished in %s, next run in %s",
-				time.Since(reconcileStart).String(),
-				obj.Spec.Interval.Duration.String())
+			var msg string
+			if obj.Spec.Suspend {
```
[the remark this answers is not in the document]
norman-zon: thx. absolutely makes sense

stefanprodan:
Please squash the 2 commits into one, signoff and force push. To sign off the commit please see https://github.com/fluxcd/kustomize-controller/pull/1396/checks?check_run_id=38894311508

stefanprodan (approved):
LGTM

Thanks @norman-zon

internal/controller/kustomization_controller.go
```diff
@@ -181,6 +181,11 @@ func (r *KustomizationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 	// Finalise the reconciliation and report the results.
 	defer func() {
+		// Do not proceed if the Kustomization is suspended
+		if obj.Spec.Suspend {
+			return
+		}
+
```
stefanprodan, line 188: This breaks the finalization and resources. Please move this just before `if conditions.IsReady(obj) {`.

stefanprodan (changes requested):

internal/controller/kustomization_controller.go
norman-zon: Thanks for the feedback. I updated the commit accordingly.

stefanprodan (approved):
LGTM

Thanks @norman-zon

Issue #1408, opened 2025-03-29
How to handle `healthCheckExprs` for resources that don't initially have a `status.conditions`


zevisert:
It could be any field, not just `status`, but in my case - when bootstrapping a new cluster with karpenter, Karpenter's EC2Nodeclass is created before the controller that manages it is live. In that case, we see errors like 

```
timeout waiting for: [EC2NodeClass/default status: 'Unknown': failed to evaluate the CEL expression 'status.conditions.filter(e, e.type == 'Ready').all(e, e.status == 'False')': no such attribute(s): status.conditions]
```

Once the controller starts, the resource grows a `status.conditions` field, but doesn't initially have one.

matheuscscp:
Also, see here the order with which the expressions are evaluated: https://fluxcd.io/flux/components/kustomize/kustomizations/#health-check-expressions

And btw I just noticed we already documented this case where expressions fail due to missing fields here:

> A common error is writing expressions that reference fields that do not exist in the custom resource. This will cause the controller to wait for the resource to be ready until the timeout is reached. To avoid this, make sure your CEL expressions are correct. The [CEL Playground](https://playcel.undistro.io/) is a useful resource for this task. The input passed to each expression is the custom resource object itself.

But it would be nice to have this FAQ in the CEL Cheatsheet as well 🙏 

And could you please also open a PR here mentioning `has()`? 🙏 

https://github.com/fluxcd/kustomize-controller/blob/860d5f833928a5c188c31d2c3cf8307e05b02377/docs/spec/v1/kustomizations.md?plain=1#L391-L397

zevisert:
Thanks @matheuscscp for the speedy reply! Seeing that you renamed the issue, I want to make it clear that the resource I'm working with doesn't have a `status` at all, nevermind `status.conditions`. I'm new to CEL expressions, but I did find the CEL cheatsheet and `has(...)` before I opened this issue. It could be that it's a feature request for Karpenter to provide an empty status like in your example to the playground. Here's a more fleshed-out/accurate example on the [CEL playground]. I didn't originally post this since the error message I get from CEL isn't exactly the same as what I see from the kustomize controller - I wasn't sure if maybe there was some extra magic top-level object in the context or anything.

What I really need is something like `has($.status)` where `$` (like with [jsonpath]) refers to the root input object

[...]

PR #1413, opened 2025-03-31
docs: mention CEL macros when discussing health check expressions

zevisert:
A quick one line addition to the docs coming from #1408


stefanprodan (approved):
LGTM

Issue #1420, opened 2025-04-10
What is the meaning of `sourceRef`?


Sleepful:
AFAIK it is meant to reference a `GitRepository` kind, one such example is in the [docs](https://fluxcd.io/flux/components/kustomize/kustomizations/#example):
```yml
kind: GitRepository
metadata:
  name: podinfo
  namespace: default
```
...
```yml
kind: Kustomization
metadata:
  name: podinfo
  namespace: default
spec:
  sourceRef:
    kind: GitRepository
    name: podinfo
```
so here the `sourceRef.name` is pointing to the `GitRepository` from above.

But I am confused by examples using `sourceRef.name: flux-system`:

```yml
  sourceRef:
    kind: GitRepository
    name: flux-system
```

Which `GitRepository` resource would this be pointing at?

I would expect it to point to a resource like the following, but it is not present in the examples:

```yml
kind: GitRepository
metadata:
  name: flux-system
  namespace: default
```

**Edit**: it seems that the `flux-system` repo is generated by `bootstrap`, and this is why it is not shown in the examples. Do you think I could add this to the docs?

PR #1421, opened 2025-04-10
Add explanation to the usage of `spec.sourceRef.name: flux-system`

Sleepful:
I want to add more clarity to the underlying reference in this property.

It had left me wondering such that I opened an issue:

https://github.com/fluxcd/kustomize-controller/issues/1420

I made sense of it after lurking the code, and now I want to add some extra clarity for the next passerby.

Ty,
Jose


Sleepful:
@stefanprodan hi, should I close this PR?

stefanprodan:
This belongs in the bootstrap docs.

PR #1426, opened 2025-04-18
[RFC-0010] Introduce object-level workload identity for KMS decryption

matheuscscp:
Depends on: https://github.com/fluxcd/pkg/pull/919

Part of: https://github.com/fluxcd/flux2/issues/5022

Fixes: #324

This PR also introduces a token cache in kustomize-controller for caching the cloud provider access tokens used for interacting with the respective KMS services. This token cache will report the following metrics:

* Counter `gotk_token_cache_events_total`
  * Help: `Total number of cache retrieval events for a Gitops Toolkit resource reconciliation.`
  * Labels: `event_type`, `kind`, `name`, `namespace`, `operation`. The values of `event_type` can be `cache_miss` or `cache_hit`. The value of `operation` is `decrypt_with_<cloud_provider>`.
* Counter `gotk_token_cache_requests_total`
  * Help: `Total number of cache requests partioned by success or failure.`
  * Labels: `status`. The values of `status` can be `success` or `failure`.
* Counter `gotk_token_cache_evictions_total`
  * Help: `Total number of cache evictions.`
  * No labels.
* Gauge `gotk_token_cached_items`
  * Help: `Total number of items in the cache.`
  * No labels.

I tested this PR in the following setups:

EKS:
* Single-tenant Pod Identity
* Single-tenant IRSA
* Multi-tenant IRSA
* Single-tenant IRSA with the SOPS feature for assuming a role
* Multi-tenant IRSA  with the SOPS feature for assuming a role

AKS:
* Single-tenant workload identity
* Multi-tenant workload identity

GKE:
* Single-tenant workload identity
* Multi-tenant workload identity with direct access
* Multi-tenant workload identity with GCP service account impersonation
* `.secretRef` with GCP service account JSON key (this is unrelated to workload identity but this code path is changing, so I tested that it still works)


docs/spec/v1/kustomizations.md
```diff
@@ -847,6 +847,15 @@ It has two fields:
 - `.secretRef.name`: The name of the secret that contains the keys to be used for
   decryption. This field can be omitted when using the
   [global decryption](#controller-global-decryption) option.
+- `.serviceAccountName`: The name of the service account used for
+  secret-less authentication with KMS services from cloud providers.
+  If a static credential for a given cloud provider is defined
+  inside the secret referenced by `.secretRef`, that static
+  credential takes priority.
```
matheuscscp, line 854: We want to allow both `secretRef` and `serviceAccountName` to be set because a user could be using a secret for a given cloud provider while using automatic/default authentication for another provider, all in the same Kustomization with SOPS-encrypted secrets using KMS keys from both cloud providers. They should still be able to do the same thing when using `serviceAccountName`, except that all the automatic authentication will now use the specified service account instead of the default one (the controller's).

internal/decryptor/decryptor.go
```diff
@@ -253,18 +264,63 @@ func (d *Decryptor) ImportKeys(ctx context.Context) error {
...
+			serviceAccount := types.NamespacedName{
+				Name:      d.kustomization.Spec.Decryption.ServiceAccountName,
+				Namespace: d.kustomization.GetNamespace(),
+			}
+			opts = append(opts, auth.WithServiceAccount(serviceAccount, d.client))
+		}
+
+		if d.tokenCache != nil {
+			involvedObject := cache.InvolvedObject{
+				Kind:      kustomizev1.KustomizationKind + "_decryption",
```
matheuscscp, line 305: For kubeconfig in the future we can use `_kubeconfig`, then we can distinguish between the two use cases in the token cache metrics

main.go
```diff
@@ -73,6 +75,10 @@ func init() {
 }
 
 func main() {
+	const (
+		tokenCacheDefaultMaxSize = 0
```
stefanprodan, line 79: Let's enable cacheing by default here and in all other controllers. The cache implementation has been shipped for some time now in Flux Operator and no issues were found in production envs. 

```suggestion
		tokenCacheDefaultMaxSize = 100
```

matheuscscp: Fixed, thanks!

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #1436, opened 2025-04-29
Fix performance regression due to using client without cache

### LINKED ISSUE #1428: Slower Reconciles with v1.5.1
When upgrading from kustomize-controller `1.5.0` to `1.5.1`, we're seeing a significant slowdown in reconcile times.

Using one Kustomization as an example, which manages ~1500 NetworkPolicies:

Average reconcile time on `1.5.0`: 25s
Average reconcile time on `1.5.1`: 2m30s

Example Kustomization setup:

```yaml
kind: Kustomization
metadata:
  name: example-kustomize
spec:
  force: false
  interval: 60m
  path: ./path/to/netpols
  prune: true
  sourceRef:
    kind: GitRepository
    name: example-repo
```

matheuscscp:
Fixes #1428

### DIFF
--- a/internal/controller/kustomization_controller.go
+++ b/internal/controller/kustomization_controller.go
@@ -48,6 +48,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	"github.com/fluxcd/cli-utils/pkg/kstatus/polling"
 	"github.com/fluxcd/cli-utils/pkg/kstatus/polling/engine"
 	"github.com/fluxcd/cli-utils/pkg/object"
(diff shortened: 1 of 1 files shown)


internal/controller/kustomization_controller.go
```diff
@@ -1160,6 +1180,24 @@ func (r *KustomizationReconciler) patch(ctx context.Context,
 	return nil
 }
 
+func (r *KustomizationReconciler) getClientAndPoller(
```
stefanprodan, line 1183: Please add a comment to this function, explain what it does

matheuscscp: Fixed, thanks!

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

Issue #1306, opened 2024-12-09
Kustomization reconciliation fails on multiple $patch: delete uses in a single strategic merge patch


tdemin:
# Environment

* Flux v2.4.0 (kustomize-controller v1.4.0)
* Kubernetes v1.29.3 (seemingly irrelevant)

# Description

If multiple `$patch: delete` strategic merge patches are provided in a single `spec.patches.patch` in a `kustomize.toolkit.fluxcd.io/v1.Kustomization` resource, kustomize-controller fails to reconcile with the following status:

```
flux-system   kustomization.kustomize.toolkit.fluxcd.io/debug         3m20s   False     kustomize build failed: recovered from kustomize build panic: runtime error: invalid memory address or nil pointer dereference
```

kustomize-controller log reports the following:

```json
{"level":"error","ts":"2024-12-09T18:41:35.677Z","msg":"Reconciliation failed after 108.007784ms, next try in 1h0m0s","controller":"kustomization","controllerGroup":"kustomize.toolkit.fluxcd.io","controllerKind":"Kustomization","Kustomization":{"name":"debug","namespace":"flux-system"},"namespace":"flux-system","name":"debug","reconcileID":"5eef40e9-8bf0-4a65-a178-eae10b04f877","revision":"main@sha1:4258282a038ab9506e1b37b3a44cc5ab4d38075e","error":"kustomize build failed: recovered from kustomize build panic: runtime error: invalid memory address or nil pointer dereference"}
```

The Kustomization resource needs to look like that:

```yaml
apiVersion: kustomize.toolkit.fluxcd.io/v1
kind: Kustomization
metadata:
  name: whatever
spec:
  patches:
    - patch: |-
        $patch: delete
        apiVersion: v1
        kind: Whatever
        metadata:
          name: whatever1
        ---
        $patch: delete
        apiVersion: v1
        kind: Whatever
        metadata:
          name: whatever2
```

# Reproduction steps

0. Create a fresh Kubernetes cluster and install Flux v2.4.0 in whatever way convenient.
1. Create a Kustomization with the following files:

```
% cat > kustomization.yaml
resources:
  - configmaps.yml
% cat > configmaps.yml
apiVersion: v1
kind: Namespace
metadata:
  name: debug
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: debug1
  namespace: debug
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: debug2
  namespace: debug
---
[...]

PR #1466, opened 2025-06-14
Add test for multiple patch delete in strategic merge patches

cappyzawa:
This test verifies that multiple $patch: delete directives in a single strategic merge patch work correctly. The test currently fails due to a panic in kustomize (upstream issue kubernetes-sigs/kustomize#5552), but will pass once the fix is included in a future kustomize release.

Ref: https://github.com/fluxcd/kustomize-controller/issues/1306

PR #1470, opened 2025-06-29
Update kustomize to v5.7.0 and add regression test for multiple patch delete

cappyzawa:
Update kustomize dependency from v5.6.0 to v5.7.0 and add regression test for multiple `$patch: delete` directives in strategic merge patches.

Multiple `$patch: delete` strategic merge patches in a single patch file cause panic in kustomize v5.6.0. This has been fixed in v5.7.0.

The regression test is based on https://github.com/fluxcd/kustomize-controller/pull/1466.

Resolves https://github.com/fluxcd/kustomize-controller/issues/1306


matheuscscp (approved):
LGTM! :rocket: 

Thanks very much @cappyzawa!

Issue #1490, opened 2025-07-21
Lockstep upgrade using Kustomization dependsOn


matheuscscp:
Over the years users have requested the ability to customize the logic used for checking dependencies with the `.spec.dependsOn` field. For example, users would like to consider a dependency ready only when it's annotated with a custom `version` annotation that matches the same annotation in the Kustomization object that depends on this dependency, i.e. when both Kustomization objects have `.metadata.annotations.version` with the same value.

## Proposal

We propose introducing the optional field `.spec.dependsOn[].readyExpr` where users can specify a Common Expression Language (CEL) expression that allows for customizing the dependency check logic, evaluating to a boolean. The context fed to this expression would be the Kustomization object itself in the variable `self` and the dependency currently being checked in the variable `dep`, both being the entire Kustomization objects as retrieved from the Kubernetes API.

Example for matching the `version` annotation:

```yaml
apiVersion: kustomize.toolkit.fluxcd.io/v1
kind: Kustomization
metadata:
  name: applier
  namespace: frontend
  annotations:
    version: v1.3.2
spec:
  ... # other fields omitted for brevity
  dependsOn:
  - name: applier
    namespace: backend
    readyExpr: self.metadata.annotations.version == dep.metadata.annotations.version
```

By default, a dependency object specified in the `.spec.dependsOn` field containing `.readyExpr` would have the entire dependency check logic specified by `.readyExpr`, i.e. the hard-coded behavior would *not* be evaluated when `.readyExpr` is specified. To make `.readyExpr` be evaluated *in addition* to the hard-coded logic, we propose introducing an opt-in feature gate called `AdditiveCELDependencyCheck`. Therefore, in order to have both the hard-coded and CEL expression logics evaluated for all dependency checks (in case the CEL expression is specified in the dependency check), users must set the feature gate `AdditiveCELDependencyCheck` to `true`.

## Implementation details

The `DependsOn` field in the `KustomizationSpec` struct currently has the type `[]meta.NamespacedObjectReference`, where `meta` is the `github.com/fluxcd/pkg/apis/meta` package. To implement this feature we should replace the type of `DependsOn` with `[]Dependency`, where:

```go
// Dependency defines a ResourceSet dependency on a Kubernetes resource.
type Dependency struct {
	meta.NamespacedObjectReference `json:",inline"`

	ReadyExpr string `json:"readyExpr,omitempty"`
}
```

[...]

PR #1491, opened 2025-07-22
Extend the readiness evaluation of dependencies with CEL expressions

stefanprodan:
This PR extends the Kustomization API with the `.spec.dependsOn[].readyExpr` optional field that can be used to define a CEL expression to determine the readiness of a Kustomization dependency. 

This is helpful for when custom logic is needed to determine if a dependency is ready. For example, when performing a lockstep upgrade, the `readyExpr` can be used to verify that a dependency has a matching version label before proceeding with the reconciliation of the dependent Kustomization.

```yaml
kind: Kustomization
metadata:
  name: app-backend
  namespace: apps
  labels:
    app/version: v1.2.3
---
kind: Kustomization
metadata:
  name: app-frontend
  namespace: apps
  labels:
    app/version: v1.2.3
spec:
  dependsOn:
    - name: app-backend
      readyExpr: >
        dep.metadata.labels['app/version'] == self.metadata.labels['app/version'] &&
        dep.status.conditions.filter(e, e.type == 'Ready').all(e, e.status == 'True') &&
        dep.metadata.generation == dep.status.observedGeneration
```

The CEL expression contains the following variables:

- `dep`: The dependency Kustomization object being evaluated.
- `self`: The Kustomization object being reconciled.

**Note:** When `readyExpr` is specified, the built-in readiness check is replaced by the logic defined in the CEL expression. You can configure the controller to run both the CEL expression evaluation and the built-in readiness check, with the `AdditiveCELDependencyCheck` [feature gate](https://fluxcd.io/flux/components/kustomize/options/#feature-gates).

Closes: #1490


api/v1/reference_types.go
```diff
@@ -40,9 +42,31 @@ type CrossNamespaceSourceReference struct {
...
+// DependencyReference defines a Kustomization dependency on another Kustomization resource.
+type DependencyReference struct {
+	// Name of the referent.
+	// +required
+	Name string `json:"name"`
+
+	// Namespace of the referent, defaults to the namespace of the Kustomization
+	// resource object that contains the reference.
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
```
matheuscscp, line 63: nit: Why not embed `meta.NamespacedObjectReference` with `json:",inline"`?

api/v1/reference_types.go
```diff
@@ -40,9 +42,31 @@ type CrossNamespaceSourceReference struct {
...
+
+	// ReadyExpr is a CEL expression that can be used to assess the readiness
+	// of a dependency. When specified, the built-in readiness check
+	// is replaced by the logic defined in the CEL expression.
+	// To make the CEL expression additive to the built-in readiness check,
```
matheuscscp, line 68: nit: The word `additive` here all by itself can sound a bit vague. Since here is a space where we don't have to spare words (like in the case of choosing a name for a feature gate), we can probably be a bit more verbose/redundant here and say that "the hard-coded logic and the CEL expression are `AND'ed` together".

docs/spec/v1/kustomizations.md
```diff
@@ -487,6 +487,51 @@ is running before deploying applications inside the mesh.
 **Note:** Circular dependencies between Kustomizations must be avoided,
 otherwise the interdependent Kustomizations will never be applied on the cluster.
 
+#### Dependency Ready Expression
+
+`.spec.dependsOn.readyExpr` is an optional field that can be used to define a CEL expression
```
matheuscscp, line 492:
```suggestion
`.spec.dependsOn[].readyExpr` is an optional field that can be used to define a CEL expression
```

nit: Maybe add the `[]` to indicate that `.spec.dependsOn` is a list?

docs/spec/v1/kustomizations.md
```diff
@@ -487,6 +487,51 @@ is running before deploying applications inside the mesh.
...
+        dep.metadata.generation == dep.status.observedGeneration
+```
+
+The CEL expression contains the following variables:
+
+- `dep`: The dependency Kustomization object being evaluated.
+- `self`: The Kustomization object being reconciled.
+
+**Note:** When `readyExpr` is specified, the built-in readiness check is replaced by the logic
+defined in the CEL expression. To make the CEL expression additive to the built-in readiness check,
```
matheuscscp, line 531: Same comment as in the `ReadyExpr` struct field comment

internal/controller/kustomization_controller.go
```diff
@@ -595,51 +597,123 @@ func (r *KustomizationReconciler) reconcile(
...
+		depName  = "dep"
+	)
+
+	celExpr, err := cel.NewExpression(expr,
+		cel.WithCompile(),
+		cel.WithOutputType(celtypes.BoolType),
+		cel.WithStructVariables(selfName, depName))
+	if err != nil {
+		return false, fmt.Errorf("failed to compile CEL expression: %w", err)
+	}
```
matheuscscp, line 692: Similar to healthchecks, the expressions can be compiled earlier in the reconciliation code to avoid returning unhandled errors that will cause controller-runtime to make useless retries with backoff. If an expression fails to compile this is 100% a terminal error.

internal/controller/kustomization_controller.go
```diff
@@ -595,51 +597,123 @@ func (r *KustomizationReconciler) reconcile(
...
+
+	var (
+		selfMap map[string]any
+		depMap  map[string]any
+	)
+
+	selfMap, err = runtime.DefaultUnstructuredConverter.ToUnstructured(self)
+	if err != nil {
+		return false, fmt.Errorf("failed to convert %s object to map: %w", selfName, err)
+	}
```
matheuscscp, line 702: The `self` object can be converted to unstructured just once, outside the dependency check loop.

api/v1/reference_types.go
matheuscscp: Sure, hard-coded vs built-in sounds the same to me, but I'd say AND'ed is pretty clear and additive not so much

PR #1509, opened 2025-09-18
Add a flag to populate source revision as commonAnnotation

adri1197:
Part of: https://github.com/fluxcd/flux2/issues/5510

### DIFF
--- a/internal/controller/kustomization_controller.go
+++ b/internal/controller/kustomization_controller.go
@@ -803,8 +803,24 @@ func (r *KustomizationReconciler) apply(ctx context.Context,
 		return false, nil, err
 	}
 
-	if cmeta := obj.Spec.CommonMetadata; cmeta != nil {
+	cmeta := obj.Spec.CommonMetadata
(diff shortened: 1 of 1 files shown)


internal/controller/kustomization_controller.go
```diff
@@ -812,7 +813,26 @@ func (r *KustomizationReconciler) apply(ctx context.Context,
...
+		}
+	}
+
+	cmeta := obj.Spec.CommonMetadata
+	if cmeta != nil {
+		if cmeta.Annotations == nil {
+			cmeta.Annotations = make(map[string]string)
+		}
+		for k, v := range revisionAnnotations {
+			cmeta.Annotations[k] = v
```
stefanprodan, line 834: Why are we annotating all resources? A Kubernetes deployment will never send an event to NC. This will cause drift for potentially thousands of non-Flux resources, which negates one of the core Flux design principles, we never alter the cluster state, unless a change is made in Git; in this case, no change was made in Git, you just blindly mutate all objects.

PR #1511, opened 2025-09-22
Update source-controller API to v1.7.0

stefanprodan:
Update source-controller API to v1.7.0 and refactor the metrics pkg.


matheuscscp (approved):
LGTM! :rocket:

PR #1529, opened 2025-10-03
Add x-kubernetes annotations to Kustomization CRD

m00nwtchr:
This allows `components` lists to be merged instead of replaced when applying a strategic merge patch to a flux Kustomization resource, as long as the OpenAPI schema is provided to Kustomize. (The same approach can be applied to other fields and CRDs)

Example:
ks.yaml:
```yaml
kind: Kustomization
metadata:
  name: podinfo
  namespace: default
spec:
  interval: 10m
  targetNamespace: default
  components: [ "../original" ]
  sourceRef:
    kind: GitRepository
    name: podinfo
  path: "./kustomize"
  prune: true
  timeout: 1m
```

kustomization.yaml:
```yaml
kind: Kustomization
resources:
  - ks.yaml

openapi:
  path: https://raw.githubusercontent.com/m00nwtchr/flux2-schemas/refs/heads/m00nwtchr-patch-1/_definitions.json
    
patches:
  - target:
      kind: Kustomization
      version: v1
      group: kustomize.toolkit.fluxcd.io
    patch: |-
      kind: Kustomization
      metadata:
        name: not-used
      spec:
        components:
          - "../patch"
```

### DIFF
--- a/config/crd/bases/kustomize.toolkit.fluxcd.io_kustomizations.yaml
+++ b/config/crd/bases/kustomize.toolkit.fluxcd.io_kustomizations.yaml
@@ -29,6 +29,10 @@ spec:
     name: v1
     schema:
       openAPIV3Schema:
+        x-kubernetes-group-version-kind:
+          - group: "kustomize.toolkit.fluxcd.io"
+            kind: "Kustomization"
+            version: "v1"
         description: Kustomization is the Schema for the kustomizations API.
         properties:
           apiVersion:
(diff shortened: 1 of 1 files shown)


m00nwtchr:
Just noticed that this file is auto-generated, I can see how you're supposed to add this info in kubebuilder: https://book.kubebuilder.io/getting-started but not sure when I'll be able to contribute those changes.

PR #1536, opened 2025-10-10
Reduce the mean time to recovery (MTTR) in case of failed deployments

matheuscscp:
Part of: https://github.com/fluxcd/flux2/issues/5584


stefanprodan (approved):
LGTM

Thanks @matheuscscp 🥇

PR #1539, opened 2025-10-28
Allow fetching artifacts from a local source-watcher

stefanprodan:
Add support for fetching artifacts from source-watcher running locally or in the same pod with kustomize-controller.

Followup: https://github.com/fluxcd/kustomize-controller/pull/1508

### DIFF
--- a/internal/controller/kustomization_controller.go
+++ b/internal/controller/kustomization_controller.go
@@ -326,13 +326,19 @@ func (r *KustomizationReconciler) reconcile(
 		}
 	}(tmpDir)
 
+	// Set the artifact URL hostname override for localhost access.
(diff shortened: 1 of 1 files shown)


internal/controller/kustomization_controller.go
```diff
@@ -326,13 +326,19 @@ func (r *KustomizationReconciler) reconcile(
 		}
 	}(tmpDir)
 
+	// Set the artifact URL hostname override for localhost access.
+	envSourceLocalhost := os.Getenv("SOURCE_CONTROLLER_LOCALHOST")
```
matheuscscp: nit: The `env*` idiom is for env var names, not values

```suggestion
	sourceLocalhost := os.Getenv("SOURCE_CONTROLLER_LOCALHOST")
```

matheuscscp (approved):
LGTM

PR #1549, opened 2025-11-14
Add annotation to disable decryption on a per object basis

danhubern:
As discussed in https://github.com/fluxcd/flux2/discussions/5612, this PR implements the `kustomize.toolkit.fluxcd.io/decrypt` annotation to disable decryption on a per object basis. This is necessary if decryption happens in the cluster at some later point in time, e.g. for `SopsSecrets`.

### DIFF
--- a/internal/decryptor/decryptor.go
+++ b/internal/decryptor/decryptor.go
@@ -187,6 +187,12 @@ func New(client client.Client, kustomization *kustomizev1.Kustomization, opts ..
 	return d, cleanup, nil
 }
 
+// IsDecryptionDisabled checks if the given object has the decrypt: disabled annotation set
+func IsDecryptionDisabled(annotations map[string]string) bool {
+	return annotations != nil &&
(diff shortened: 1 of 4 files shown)


internal/decryptor/decryptor.go
```diff
@@ -187,6 +187,14 @@ func New(client client.Client, kustomization *kustomizev1.Kustomization, opts ..
 	return d, cleanup, nil
 }
 
+// IsDecryptionDisabled checks if the given object has the decrypt: disabled annotation set
+func IsDecryptionDisabled(annotations map[string]string) bool {
+	if annotations != nil && annotations[fmt.Sprintf("%s/decrypt", kustomizev1.GroupVersion.Group)] == kustomizev1.DisabledValue {
+		return true
+	}
```
stefanprodan: This should be case insensitive, see how we do it here: https://github.com/fluxcd/pkg/blob/0fe2a7c7e3ddedf39d862c98df89f5878777e748/ssa/utils/meta.go#L59

matheuscscp (commented):
Some (very) important changes in the docs structure, then it looks good :+1:

stefanprodan:
Please squash all commits into one

stefanprodan (approved):

Issue #1600, opened 2026-02-26
Update Age and SOPS to enable use of hybrid post-quantum cypher


BobyMCbobs:
# What

In Age v1.3.x, support for hybrid ML-KEM-768 KEM was added, this enables post-quantum resistance. As SOPS utilises Age for this functionality, both dependencies are bumped to the latest.

# Why

Post-quantum resistance is much desired.

# Additional info

A test also also been added to test the behaviour, although based almost exactly on an existing one.

PR #1601, opened 2026-02-26
feat: update Age and SOPS for Age post-quantum cypher

BobyMCbobs:
- updates Age and SOPS dependencies to latest
- adds test for Age hybrid post-quantum cypher

fixes: https://github.com/fluxcd/kustomize-controller/issues/1600


stefanprodan:
@BobyMCbobs we'll include this in Q2 roadmap and it will be shipped in Flux v2.9. Given that this is a new feature can not go into a patch release for v2.8.

PR #1621, opened 2026-03-14
Migrate event recorder to fluxcd/pkg/runtime/events.Recorder

### LINKED ISSUE #5761: Implement Kubernetes Event API v1
Transition the Flux ecosystem from legacy `core/v1` events to the modern `events.k8s.io/v1` API. The following changes have to follow a backward compatibility approach where legacy fields  may be bridged to the new schema (e.g., keep `InvolvedObject` regardless of being renamed, add a new `RelatedObject` field) to ensure no breaking changes for users relying on current event logs.

**Action Items**
- [x] Add a new AnnotatedEventRecorder interface upstream in `client-go/tools/events` - kubernetes/kubernetes#138103
- [x] Controller-Runtime - kubernetes-sigs/controller-runtime#3509
- [x] Update `fluxcd/pkg` to use `events.k8s.io/v1` - fluxcd/pkg#1144
- [ ] Update all controllers using the changes done in `runtime/events` from `fluxcd/pkg`. Use new field `relatedObject` to include object's source metadata (as InvolvedObject does) and `Action` as mandatory field.
   - [ ] Mandatory controllers
      - [ ] Notification-Controller - fluxcd/notification-controller#1267
         - [ ]  Adapt the event receiver to handle the events.k8s.io/v1 schema.
         - [ ] Update [spec/v1beta3/events.md](https://github.com/fluxcd/notification-controller/blob/main/docs/spec/v1beta3/events.md) to document new fields.
      - [ ] Helm-Controller fluxcd/helm-controller#1510
      - [ ] Kustomize-Controller fluxcd/kustomize-controller#1621
      - [ ] Source-Controller fluxcd/source-controller#2011
   - [ ] Optional controllers
      - [ ] Source-Watcher fluxcd/source-watcher/pull/343
      - [ ] Image-Automation-Controller fluxcd/image-automation-controller#1038
      - [ ] Image-Reflector-Controller fluxcd/image-reflector-controller#889

**Reference to the new API spec and changes**
- [API Deprecation Guide](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#event-v125) ( `events.k8s.io/v1beta1` deprecated since k8s 1.25) 
- [KEP-383](https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/383-new-event-api-ga-graduation/README.md): New Event API GA Graduation

adri1197:
Migrate the controller's event recorder to `fluxcd/pkg/runtime/events.Recorder` and K8s event API from `v1beta1` to `v1`.

- Replace `kuberecorder.EventRecorder` with `events.Recorder` to support structured event metadata (source object reference and action type).
- Add an action to every event call (`Reconciled`, `Failed`, `Waiting`, `Applied`, `Deleted`, `Progressing`) and thread the source reference through `reconcile`, `apply`, `prune`, `checkHealth`, and `finalize`.
- Bump `fluxcd/pkg/runtime` to v0.113.0.

Part of - https://github.com/fluxcd/flux2/issues/5761


matheuscscp:
Hey @adri1197 please rebase all your PRs, we have upgraded to k8s 1.37 across all controllers and bumped most fluxcd/pkg packages. The only remaining dep for you to bump across your PRs is fluxcd/pkg/runtime. Please start with kustomize-controller as there's another PR depending on the runtime bump :pray:

matheuscscp (approved):
LGTM! 🚀 

Thanks @adri1197!

stefanprodan:
> Make the success event message stable across reconciliations of the same revision (drop the duration from the message; log it instead) and stamp each event with a unique token annotation, so Kubernetes can deduplicate

This is not true, deduplication will never happen now because `ResourceVersion` will always differ (due to status patching) and even if we mutate and set that to `nil`, the new client-go broadcaster [hardcodes 6m for aggregation](https://github.com/kubernetes/client-go/blob/v0.37.0/tools/events/event_broadcaster.go#L50), basically this PR reverts https://github.com/fluxcd/kustomize-controller/pull/1743

internal/controller/kustomization_controller.go
stefanprodan: This needs to be deleted as it's no longer true, also the test `TestKustomizationReconciler_EventAggregation` needs a rename, there is no event aggregation possible with this PR.

internal/controller/kustomization_controller.go
```diff
@@ -166,7 +166,8 @@ func (r *KustomizationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 			log.Info(msg,
 				"revision", obj.Status.LastAttemptedRevision,
 				"duration", time.Since(reconcileStart).String())
-			r.event(obj, obj.Status.LastAppliedRevision, obj.Status.LastAppliedOriginRevision, eventv1.EventSeverityInfo, msg,
+			r.event(obj, nil, obj.Status.LastAppliedRevision, obj.Status.LastAppliedOriginRevision, eventv1.EventSeverityInfo,
```
stefanprodan: Why we do not set the source on a success reconcile event?

matheuscscp:
We will do what's described here now: https://github.com/fluxcd/kustomize-controller/issues/1742#issuecomment-5760750001

Please rebase after that! :pray:

matheuscscp:
@adri1197 All done, please rebase! :pray:

internal/controller/kustomization_controller.go
adri1197: Removed this test case due to https://github.com/fluxcd/kustomize-controller/issues/1742#issuecomment-5760750001 revert

internal/controller/kustomization_controller.go
adri1197: Done! 😃

matheuscscp (approved):
LGTM! 🚀 

Thanks @adri1197!

stefanprodan:
Found some major blockers:

- Messages are rejected over 1024 bytes which makes all our events to fail https://github.com/kubernetes/kubernetes/blob/v1.37.0/pkg/apis/core/validation/events.go#L37 and https://github.com/kubernetes/kubernetes/blob/f54c212e3a2f75d674b717a9b29052b20b60aefc/pkg/apis/core/validation/events.go#L183

- The Series grouping does not contain the message so in the 6m window all out events for a revision will be dropped, you only see the first message https://github.com/kubernetes/client-go/blob/v0.37.0/tools/events/event_broadcaster.go#L297-L310

stefanprodan:
Given the 1024 limit of the event message, we can not ship this as is, all of Flux observability is built on the assumption that the events cary the desired state changes and errors. With 1024 limit, KC can no longer issue events with the list of applied changes, dry-run errors will no longer be surfaced to users, HC can no longer report helm upgrade errors, etc.

PR #1646, opened 2026-04-25
docs: add debugging tips to DEVELOPMENT.md

alliasgher:
## Summary
Adds a 'Debugging the controller locally' section to `DEVELOPMENT.md` covering knobs not already in 'How to run the controller locally':

- `RUNTIME_NAMESPACE` to scope the watch
- `--concurrent=1 --concurrent-ssa=1` to serialize reconciles and SSA for a clean trace
- `flux suspend kustomization` for unrelated objects sharing the cluster

Cross-controller setup with source-controller (port-forward + `SOURCE_CONTROLLER_LOCALHOST`) is already documented under 'How to run the controller locally' and is not duplicated here.

Assisted-by: Claude/claude-opus-4-7

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -92,3 +92,43 @@ make deploy
 ```
 
 Running the above will also deploy `source-controller` and its CRDs to the cluster.
+
+## Debugging the controller locally
+
+When reproducing an issue or stepping through reconciliation logic, the
+following knobs make local runs cheaper and the resulting logs easier to
+read.
+
+### Limit the watched namespace
+
+The controller watches every namespace by default. To narrow it to a single
+namespace, set the `RUNTIME_NAMESPACE` environment variable before invoking
+`make run`:
+
+```sh
+RUNTIME_NAMESPACE=flux-system make run
+```
+
+### Reduce reconcile concurrency
+
+Each `Kustomization` reconcile is processed concurrently (default
+`--concurrent=4`, plus an independent `--concurrent-ssa=4` for server-side
+apply). When debugging it is almost always easier to follow a serial
+trace; pass `--concurrent=1 --concurrent-ssa=1` so reconciles and SSA
+operations run one at a time:
+
+```sh
+go run ./main.go --concurrent=1 --concurrent-ssa=1
+```
+
+### Suspend unrelated objects
+
+If the controller is sharing a cluster with other Flux objects, suspend
+anything not relevant to the test you're running so their reconciles don't
+interleave with yours:
+
+```sh
+flux suspend kustomization <name>
+```
+
+Resume with `flux resume kustomization <name>` when you're done.


stefanprodan (commented):
The debugging info is already covered in "How to run the controller locally". The added text contains inaccurate info, you can not run SC locally as KC can not reach it, there is no mention of the ENV VAR, etc.

@alliasgher did you actually tested this? Did you used AI to generate this text, if so you need to disclose it, see  https://github.com/fluxcd/flux2/discussions/5848

alliasgher:
Thanks for the catch @stefanprodan, you are right on all three points. AI assistance disclosed (`Assisted-by: Claude/claude-opus-4-7` is now on the commit and the PR body, per https://github.com/fluxcd/flux2/discussions/5848 — sorry for missing it the first time round).

Fixed in f9b8044: I removed the 'Cross-controller interactions' paragraph entirely. The cross-controller setup is already covered correctly in 'How to run the controller locally' via `kubectl port-forward svc/source-controller 8080:80` + `SOURCE_CONTROLLER_LOCALHOST=localhost:8080`, and my paragraph was both wrong (kustomize-controller can't reach a source-controller running outside the cluster) and missed the env var.

What is left is just `RUNTIME_NAMESPACE`, `--concurrent=1 --concurrent-ssa=1`, and `flux suspend kustomization`, which I do not think the existing section covers. Happy to drop the section entirely if you'd rather not have it.

DEVELOPMENT.md
```diff
@@ -91,4 +91,44 @@ Deploy `kustomize-controller` into the cluster that is configured in the local k
...
+
+When reproducing an issue or stepping through reconciliation logic, the
+following knobs make local runs cheaper and the resulting logs easier to
+read.
+
+### Limit the watched namespace
+
+The controller watches every namespace by default. To narrow it to a single
+namespace, set the `RUNTIME_NAMESPACE` environment variable before invoking
+`make run`:
```
stefanprodan, line 106: Setting RUNTIME_NAMESPACE does not restrict the watchers, it just tells the controller in which namespace it runs. Watching only the flux-system namespace would break debugging so there is no point in having this `Limit the watched namespace` section.

DEVELOPMENT.md
```diff
@@ -91,4 +91,44 @@ Deploy `kustomize-controller` into the cluster that is configured in the local k
...
+RUNTIME_NAMESPACE=flux-system make run
+```
+
+### Reduce reconcile concurrency
+
+Each `Kustomization` reconcile is processed concurrently (default
+`--concurrent=4`, plus an independent `--concurrent-ssa=4` for server-side
+apply). When debugging it is almost always easier to follow a serial
+trace; pass `--concurrent=1 --concurrent-ssa=1` so reconciles and SSA
+operations run one at a time:
```
stefanprodan, line 118: How would you debug goroutine leaks, dead locks and other issues that only surface when concurrency is used? Also if you suspend all objects but the one you are debugging, then setting `--concurrent=1` has no effect.

DEVELOPMENT.md
```diff
@@ -91,4 +91,44 @@ Deploy `kustomize-controller` into the cluster that is configured in the local k
...
+```
+
+### Suspend unrelated objects
+
+If the controller is sharing a cluster with other Flux objects, suspend
+anything not relevant to the test you're running so their reconciles don't
+interleave with yours:
+
+```sh
+flux suspend kustomization <name>
```
stefanprodan, line 131: There is an `--all` flag

alliasgher:
Apologies for the long silence on this @stefanprodan, and thank you for the review. I went back and checked the code rather than just taking the note, and you are right on all three points.

On the namespace section, I was wrong. `RUNTIME_NAMESPACE` is only read at all when `--watch-all-namespaces` is false:

```go
watchNamespace := ""
if !watchOptions.AllNamespaces {
    watchNamespace = os.Getenv(runtimeCtrl.EnvRuntimeNamespace)
}
```

and that flag defaults to `true` (`fluxcd/pkg/runtime/controller/watch.go`). So the example I documented, `RUNTIME_NAMESPACE=flux-system make run`, changes nothing on its own. Beyond the factual error, your wider point stands too: narrowing the watch to one namespace would hide exactly the cross-namespace behaviour you often need while debugging.

On concurrency, agreed. Recommending `--concurrent=1` as general debugging advice is backwards, since goroutine leaks, deadlocks and races only show up under concurrency, and it is redundant anyway once you have suspended everything else. And `flux suspend kustomization --all` is the right incantation, not naming objects one at a time.

That removes essentially all of the substance here, so I do not think there is a useful PR left in it. I am happy to close it, or to cut it down to nothing but the `--all` mention if you think that is worth keeping. Your call, and either way sorry for the noise.

alliasgher:
Closing this. Two of the three tips were wrong rather than merely imprecise, and once they are removed there is not enough left to justify a review cycle.

For the record, so this is searchable if anyone tries the same thing later:

- **Limit the watched namespace** was wrong. `RUNTIME_NAMESPACE` is only read when `--watch-all-namespaces` is false, and that flag defaults to `true`, so the documented command had no effect. @stefanprodan also made the stronger point that narrowing the watch would hide cross-namespace behaviour you often need while debugging, so the section should not exist even in a corrected form.
- **Reduce reconcile concurrency** was backwards as general advice: goroutine leaks, deadlocks and races only surface under concurrency, and it is redundant once unrelated objects are suspended.
- **Suspend unrelated objects** was the only sound tip, and it amounts to mentioning that `flux suspend <kind> --all` exists, which is not worth a PR on its own.

Thanks @stefanprodan for the review that caught this, and apologies to the other repos for receiving the same wrong text. If a maintainer does want a debugging section in DEVELOPMENT.md, I am happy to write one from scratch against what the flags actually do.

PR #1657, opened 2026-05-28
Add instructions on using Podman

cipherboy:
make test will work just fine in Podman, provided the Docker-compatible socket is set up and the `DOCKER_HOST` environment variable is set. This allows the Docker API client libraries used by `ory/dockertest` to provision Podman containers.


matheuscscp (approved):
LGTM

stefanprodan (approved):

PR #1659, opened 2026-06-03
Introduce generic Kubernetes workload identity for SOPS OpenBao/Vault

### LINKED ISSUE #1658: Feature Request: Native OpenBao/Vault Authentication
Flux's Kustomize Controller currently takes authentication via [hard-coded Vault token](
https://fluxcd.io/flux/components/kustomize/kustomizations/#hashicorp-vault):

```yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: kustomize-controller
  namespace: flux-system
spec:
  template:
    spec:
      containers:
      - name: manager
        env:
        - name: VAULT_TOKEN
          value: <token>
```

While functional, per discussions with @stefanprodan, this leads to the unnecessary exposure of this token and results in problems rotating it. [The other form](https://fluxcd.io/flux/components/kustomize/kustomizations/#hashicorp-vault-secret-entry):

```yaml
---
apiVersion: v1
kind: Secret
metadata:
  name: sops-keys
  namespace: default
data:
  # Exemplary Hashicorp Vault Secret token
  sops.vault-token: <BASE64>
```

has a similar problem. 

The proposal is to move this to standard [OpenBao authentication](https://openbao.org/docs/auth/). This involves one or more login calls, for example:

 - Logging in [via TLS Certificate Auth](https://openbao.org/api-docs/auth/cert/#login-with-tls-certificate-method), taking the request path, the role `name`, and a mTLS client certificate on the connection.
 - Logging in [via JWT auth](https://openbao.org/api-docs/auth/jwt/#jwt-login), taking the request path, the `role`, and the `jwt`. If the JWT is not explicitly configured, this could be provisioned from the running container's service account token (`/var/run/secrets/kubernetes.io/serviceaccount/token`). 
 - Logging in [via AppRole](https://openbao.org/api-docs/auth/approle/#login-with-approle), which follows a similar pattern to all other roles: path and one or more parameters.

Only TLS cert auth requires special handling, injecting it into the client. JWT could also be special cased, to allow reading the mounted service account tokens. All remaining auth methods would probably require rotating shorter-lived token information in a similar form to the existing token rotation.

This could be accomplished entirely in this controller without requiring modifications to SOPS. Irrespective of config format, where:

https://github.com/fluxcd/kustomize-controller/blob/617fc77745509d4d9cc8aebf560613efc56bc415/internal/sops/keyservice/server.go#L50-L53

is used on line 275:

https://github.com/fluxcd/kustomize-controller/blob/617fc77745509d4d9cc8aebf560613efc56bc415/internal/sops/keyservice/server.go#L269-L280

we can replace this with a function like (pseudocode):

```go
func (ks Server) ApplyVaultToken(vaultKey *hcvault.MasterKey) error {
  data := map[string]any{}
  config := api.DefaultConfig()

  switch ks.vaultAuthType {
  case "token":
    ks.vaultToken.ApplyToMasterKey(&vaultKey)
    return nil
  case "cert":
    if err := config.ConfigureTLS(&api.TLSConfig{
      ClientCert: ks.vaultCert,
      ClientKey: ks.vaultKey,
      CACert: ks.vaultCACert, // optional
    }); err != nil {
      return err
    }
    data["name"] = ks.vaultCertRole
  case "container-token":
    data["jwt"] = os.ReadFile(KubeSAToken)
    data["role"] = ks.vaultJWTRole
  case "manual":
    data[...] = ks.vault...
  }

   client, err := api.NewClient(config)
   resp, err := client.Logical().Write(ks.vaultAuthPath, data)
   token := hcvault.Token(resp.Auth.ClientToken)
   token.ApplyToMasterKey(vaultKey)
}
```

where `api` is either the [OpenBao](https://pkg.go.dev/github.com/openbao/openbao/api/v2) or [Vault](https://pkg.go.dev/github.com/hashicorp/vault/api) clients.

This fetches the token for the subsequent SOPS call and injects it to SOPS without SOPS requiring an update to support Vault's auth methods.

OpenBao's [JWT auth engine](https://openbao.org/api-docs/auth/jwt/#createupdate-role) supports configuring the trusted audience (`bound_audience`).

matheuscscp:
Closes: #1658

Adds support for decrypting SOPS secrets encrypted with the OpenBao/Vault transit engine by authenticating with a JWT-backed auth method — e.g. the [**Kubernetes auth method**](https://openbao.org/api-docs/auth/kubernetes/) or the [**JWT auth method**](https://openbao.org/api-docs/auth/jwt/) — instead of a static token. The controller exchanges a Kubernetes ServiceAccount token for a short-lived Vault token, supporting both object-level and controller-level workload identity.

This is **purely additive and non-breaking**: the existing `sops.vault-token` (Secret) and `VAULT_TOKEN` (environment variable) paths are unchanged and take precedence. The feature is off unless the operator opts in via a flag.

Precedence per Kustomization: `sops.vault-token` → Kubernetes auth → `VAULT_TOKEN`.

### 1. Tell the controller which Vault instances to trust

Create a ConfigMap in the controller's namespace mapping each Vault address to its login path, and point the controller at it with `--sops-vault-configmap`.

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: sops-vault-config
  namespace: flux-system
data:
  config.yaml: |
    instances:
      - address: https://vault.example.com:8200
        loginPath: auth/kubernetes/login
```

The `loginPath` is used verbatim, so it supports any JWT-backed auth method (e.g. `auth/jwt/login`) and namespace-prefixed paths (e.g. `ns1/ns2/auth/kubernetes/login`), not just the Kubernetes auth method.

```yaml
# kustomize-controller Deployment
args:
  - --sops-vault-configmap=sops-vault-config
```

When the flag is empty, Kubernetes auth is disabled. Within Kubernetes auth, only addresses listed in the ConfigMap can be used (the static token paths are unaffected and work for any address).

### 2. Enable and configure the Kubernetes auth method in OpenBao/Vault

Enable the Kubernetes auth method at the mount referenced by your `loginPath` (`auth/kubernetes/login` → mount `kubernetes`; use `-path=<mount>` for a custom mount, e.g. one mount per cluster when several clusters share a Vault):

```sh
vault auth enable kubernetes
```

Vault validates incoming tokens by calling the cluster's TokenReview API, so it needs the API server address, the cluster CA, and a reviewer token. Create a reviewer ServiceAccount with the `system:auth-delegator` role and a long-lived token:

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: vault-auth
  namespace: flux-system
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: vault-auth-tokenreview
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: system:auth-delegator
subjects:
  - kind: ServiceAccount
    name: vault-auth
    namespace: flux-system
---
apiVersion: v1
kind: Secret
metadata:
  name: vault-auth
  namespace: flux-system
  annotations:
    kubernetes.io/service-account.name: vault-auth
type: kubernetes.io/service-account-token
```

Then configure the auth method with the reviewer token and CA read from that Secret:

```sh
kubectl -n flux-system get secret vault-auth -o jsonpath='{.data.token}'  | base64 -d > token
kubectl -n flux-system get secret vault-auth -o jsonpath='{.data.ca\.crt}' | base64 -d > ca.crt

vault write auth/kubernetes/config \
  kubernetes_host=https://<api-server>:6443 \
  kubernetes_ca_cert=@ca.crt \
  token_reviewer_jwt=@token
```

`kubernetes_host` must be the API server address reachable from Vault. With one Vault shared across clusters, repeat this for each cluster on its own mount (`-path`), each configured with that cluster's API server, CA, and reviewer token.

### 3. Create a role and policy per ServiceAccount

For each ServiceAccount that should decrypt, create a Kubernetes auth role named `{namespace}_{name}`, bound to that ServiceAccount and namespace, with the **audience set to the Vault address**, and a policy granting `update` on the relevant `transit/decrypt/<key>` paths. Example for ServiceAccount `sops` in namespace `apps`:

```sh
vault write auth/kubernetes/role/apps_sops \
  bound_service_account_names=sops \
  bound_service_account_namespaces=apps \
  audience=https://vault.example.com:8200 \
  token_policies=apps-decrypt
```

### 4. Point the Kustomization at a ServiceAccount

```yaml
kind: Kustomization
spec:
  decryption:
    provider: sops
    serviceAccountName: sops   # token issued for this SA; role = {namespace}_sops
```

- **Object-level** workload identity (per `serviceAccountName`) requires the `ObjectLevelWorkloadIdentity` feature gate to be enabled.
- If `serviceAccountName` is omitted, the controller's own ServiceAccount is used (controller-level workload identity), and the role is derived from it.

The token audience is always the Vault address, and the Vault role is always derived from the authenticating ServiceAccount as `{namespace}_{name}`.


matheuscscp:
This is now covered by an automated end-to-end job (`sops-openbao`) in `.github/workflows/e2e.yaml`, which on every PR spins up:

* **2 OpenBao instances** — one configured with the [Kubernetes auth method](https://openbao.org/docs/auth/kubernetes/), the other with the generic [JWT auth method](https://openbao.org/api-docs/auth/jwt/) (validating the ServiceAccount token against the cluster's public signing keys) — running as containers on the shared `kind` Docker network.
* **2 Secrets**, each encrypted with a different OpenBao instance's transit engine.
* **2 kind clusters**, each running source-controller and kustomize-controller. Both clusters decrypt **both** Secrets, i.e. each controller authenticates to **both** OpenBao instances using only its ServiceAccount token (audience-scoped per address).

It also exercises **both** workload identity levels: cluster 1 uses controller-level identity (the controller's own ServiceAccount), and cluster 2 uses object-level identity (`.spec.decryption.serviceAccountName` with the `ObjectLevelWorkloadIdentity` feature gate enabled).

Each cluster gets a `sops-vault-config` ConfigMap mapping both OpenBao addresses to that cluster's own auth mount. Cluster 1 looks like this:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: sops-vault-config
  namespace: kustomize-system
data:
  config.yaml: |
    instances:
      - address: http://172.18.0.10:8200      # Kubernetes auth method
        loginPath: auth/kubernetes-c1/login
      - address: http://172.18.0.11:8200      # JWT auth method
        loginPath: auth/jwt-c1/login
```

And in the other cluster it looks like this:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: sops-vault-config
  namespace: kustomize-system
data:
  config.yaml: |
    instances:
      - address: http://172.18.0.10:8200      # Kubernetes auth method
        loginPath: auth/kubernetes-c2/login
      - address: http://172.18.0.11:8200      # JWT auth method
        loginPath: auth/jwt-c2/login
```

The addresses (discovered at runtime) are identical across clusters — a single shared pair of OpenBao instances — but each cluster points at its own auth mount (`-c1`/`-c2`), since every OpenBao has a separate auth mount per cluster. Having one instance authenticate via the generic JWT auth method confirms the verbatim `loginPath` supports any JWT-backed auth method end-to-end, not just the Kubernetes auth method.

internal/decryptor/decryptor.go
```diff
@@ -372,7 +385,173 @@ func (d *Decryptor) SetAuthOptions(ctx context.Context) {
...
+// VAULT_* environment variables, with the address overridden by vaultAddress.
+func vaultKubernetesLogin(ctx context.Context, vaultAddress, mount, role, jwt string) (string, error) {
+	cfg := vaultapi.DefaultConfig()
+	cfg.Address = vaultAddress
+	client, err := vaultapi.NewClient(cfg)
+	if err != nil {
+		return "", fmt.Errorf("failed to create Vault client for '%s': %w", vaultAddress, err)
+	}
+	secret, err := client.Logical().WriteWithContext(ctx,
+		fmt.Sprintf("auth/%s/login", mount),
```
cipherboy: @matheuscscp This will not work with namespaces, can we move the shape to be:


```suggestion
		loginPath,
```

rather than templating `mount`?  Namespaces would be prefixed (`ns1/ns2/auth/kubernetes/login`). This will give us a fair bit of flexibility to support other JWT-backed auth methods as well and stop us from assuming that the auth method is a mount path of a specified type.

matheuscscp: Fixed, thanks! I also updated the PR description, rerun my manual e2e tests and updated the e2e comment as  well. I also tested with the agnostic `auth/jwt/login` and it works!

docs/spec/v1beta2/kustomizations.md
```diff
@@ -946,7 +946,7 @@ In order to store secrets safely in a public or private Git repository,
 you can use [Mozilla SOPS](https://github.com/mozilla/sops)
 and encrypt your Kubernetes Secrets data with [age](https://age-encryption.org/v1/)
 and [OpenPGP](https://www.openpgp.org) keys, or using provider
-implementations like Azure Key Vault, GCP KMS or Hashicorp Vault.
+implementations like Azure Key Vault, GCP KMS or OpenBao/Vault.
```
stefanprodan: Undo all changes in this doc please.

matheuscscp: Fixed, thanks!

stefanprodan (approved):
LGTM 🚀 

PS. This needs an update https://fluxcd.io/flux/guides/mozilla-sops/#encrypting-secrets-using-hashicorp-vault

matheuscscp:
I have also replaced the Vault container in the Go tests with OpenBao.

PR #1665, opened 2026-06-11
fix: keep failed-to-prune resources in status.Inventory (#1664)

### LINKED ISSUE #1664: Resources not pruned after a rapid series of commits where intermediate `kustomize build` output is broken
### Describe the bug

When a `Kustomization`'s source repo is updated through a fast series of commits that:

1. introduce/keep a kustomize overlay producing resource `X`,
2. then break `kustomize build` (e.g., delete leaf files while keeping the parent `resources:` reference), and
3. then later commits remove the dangling reference so the build starts producing the new desired set without `X`,

`X` can remain on the cluster forever as an orphan: it is no longer in git, no longer in the `Kustomization`'s `status.inventory`, but still labeled `kustomize.toolkit.fluxcd.io/name=<kustomization>` and `kustomize.toolkit.fluxcd.io/namespace=<ns>` and still has the `kustomize-controller` listed in `managedFields`. Subsequent successful reconciles do not detect it as prunable, because the inventory comparison no longer sees `X` in the "old" set — the snapshot has already advanced past the state where `X` was present.

Aggravating factor in our specific case: `X` had a finalizer from another controller (Karpenter's `karpenter.k8s.aws/termination`). If Flux did attempt a delete during one of the intermediate reconciles, it would have been accepted (`deletionTimestamp` set) but stuck on the finalizer; inventory still moved on regardless. We did not catch this in flight; we only observed the orphan state hours later, by which point logs and Events had rotated out.

### Steps to reproduce

Repository layout under a `Kustomization` with `prune: true`, `path: ./infra/nodes`:

```
infra/nodes/
  kustomization.yaml          # lists: apps, application
  apps/
    kustomization.yaml        # lists: application-nodes
    application-nodes/
      kustomization.yaml      # lists: general
      general/
        kustomization.yaml    # lists: 1a, 1b, 1c
        1a/kustomization.yaml # patches base name to application-nodes-1a
        1b/...
        1c/...
        base/nodes.yaml       # CR(s) named application-nodes
  application/
    node-class.yaml           # CR application-nodes (the new consolidated resource)
    node-pool.yaml
```

Commits land on `main` within a short window (faster than the reconcile interval in some cases, slower in others — i.e. some intermediate states ARE reconciled):

1. Add `infra/nodes/application/` and reference it from top-level `kustomization.yaml`. Both `apps` and `application` are listed simultaneously. Flux reconciles → emits both old `application-nodes-1a/1b/1c` AND new `application-nodes`.
2. Delete the leaf files under `infra/nodes/apps/application-nodes/general/{1a,1b,1c}/` but do not remove the `resources: [1a, 1b, 1c]` entry in `general/kustomization.yaml`. `kustomize build infra/nodes` now fails ("accumulating resources: ... no such file or directory").
3. Several more commits land that eventually clean up `apps/application-nodes/` references in `apps/kustomization.yaml` and remove `apps` from the top-level. Build succeeds again, output excludes `application-nodes-1a/1b/1c`.

Observed: `application-nodes-1a/1b/1c` remain on the cluster indefinitely after step 3 is reconciled. `kubectl get <cr> application-nodes-1a -o jsonpath='{.metadata.labels}'` still shows the Flux ownership labels. The `Kustomization`'s `status.inventory.entries` does not list them. `flux reconcile kustomization infra-nodes --with-source` is a no-op against them.

### Expected behavior

Either:

(a) Flux should not advance inventory past a state where a prune did not complete, so the next successful reconcile still sees the resource and retries the prune; or
(b) Flux should fall back to a label-based GC scan to catch resources labeled `kustomize.toolkit.fluxcd.io/name=<kustomization>` that aren't in the current inventory.

The current behavior silently leaves cluster state diverged from git with no surfaced error condition.

### Versions

- `kustomize-controller` v1.8.2
- Flux v2.8.3 (managed by `flux-operator` v0.45.0)
- Kubernetes EKS v1.35
- Resources involved: `karpenter.k8s.aws/EC2NodeClass` (carrying `karpenter.k8s.aws/termination` finalizer)

### Related prior art

- #409 (closed 2021) — same category of failure ("Orphan objects after failed reconciliation"), fixed for the `validation: none` cause in #426 but the general "early reconcile exit → no GC snapshot" risk surface still exists in different forms.
- #1510 / #1513 (Sep 2025) — separate inventory-bookkeeping bug.
- fluxcd/flux2 #5646 (open) — adjacent symptoms of orphans surviving partial-failure reconciles, focused on parent/child Kustomizations.

### Workaround

Manual `kubectl delete <kind> <name>` for each orphan. With no remaining references, finalizers (in our case Karpenter's) are removed within seconds.

### Possible root cause direction

Worth tracing in `kustomize-controller`:

- the order of operations between inventory write and prune execution in `Reconcile()`,
- whether prune failures (delete-request errors, or successful delete but stuck-on-finalizer with no follow-up) cause the new inventory snapshot to be rolled back, and
- whether `kustomize build` errors during an intermediate revision can cause the "last known good" inventory used for diffing to be stale relative to the snapshot the controller thinks it has.

Happy to provide more detail from our cluster if useful (the orphans are still in place on one cluster, untouched, for further inspection).

gecube:
## Summary

Fixes #1664. When `status.Inventory` is advanced to the new build output before the prune step succeeds, and prune subsequently fails because the apiserver rejects the DELETE (admission webhook denial, in-use resource, etc.), the stale resources are silently lost from Flux's tracking. They:

- are no longer in `status.Inventory`,
- still carry the `kustomize.toolkit.fluxcd.io/name=…` label,
- are not surfaced as prunable by the next reconcile's old-vs-new diff (the "old" snapshot has already advanced past them),
- cannot be re-pruned by `flux reconcile` — only manual `kubectl delete` removes them.

Result: long-lived orphans labeled by the `Kustomization` but invisible to its GC.

## Behaviour change

`prune()` now also returns the slice of *survivor* objects whose DELETE wasn't confirmed (no `DeletedAction` or `SkippedAction` `ChangeSetEntry`). The reconcile path merges those survivors back into `status.Inventory` immediately before recording `PruneFailedReason` and returning the error. The next reconcile then sees them in the diff again and retries the prune.

Treating `SkippedAction` entries as settled means resources explicitly opted out (`kustomize.toolkit.fluxcd.io/prune: disabled`) are **not** re-tracked — keeping them would create an infinite retry loop for opt-outs.

Paths NOT affected:
- `spec.prune: false` — `prune()` short-circuits before any of the new logic runs.
- The finalizer path (Kustomization deletion) uses `deleteObjects` directly, not `prune()`.
- Successful prunes — `survivors` is empty, no merge occurs.

## Reproducer

I built a minimal kind-based reproducer at https://github.com/gecube/flux-kc-1664-repro. The relevant variant — `repro-restart-matrix.sh` scenario `webhook-deny-delete` — reproduces the exact production state from #1664 (3/3 consecutive runs): per-AZ ConfigMaps survive past the consolidation commit, `deletionTimestamp` not set, `status.Inventory` already advanced past them. With this patch applied to a locally-built `kustomize-controller`, the same reproducer should leave the orphans in inventory and retry prune on the next reconcile (I have not yet rebuilt-and-loaded a custom controller image into kind to verify end-to-end; happy to do so if useful before merge).

## Tests

- `internal/inventory`: new subtest `merge_re-adds_objects_whose_prune_failed` plus a nil-safety subtest.
- `internal/controller`: new file `kustomization_prune_survivors_test.go` with unit-level coverage of `pruneSurvivors` across the `DeletedAction` / `SkippedAction` / `UnknownAction` / missing-entry cases.
- Existing tests pass locally:
  - `go test ./internal/inventory/...`
  - `go test ./internal/controller/... -run "Prune|Inventory|DeletionPolicy"`
- `go vet ./...` clean, `gofmt -l` clean.

## Notes for maintainers

Marking this **draft** because:

1. I haven't yet end-to-end-tested with a custom-built controller image against the kind reproducer (only unit-tested the new functions). Happy to do so before merge.
2. There may be an alternative design you'd prefer — e.g. tracking the failed-prune subset separately in `status` (under `staleResources` or similar) rather than merging back into inventory; or surfacing an additional `OrphanedResourcesDetected` condition; or a different stance on `SkippedAction` re-tracking semantics. I picked the smallest change that restores GC convergence.
3. The `SkippedAction` case is the subtle one — could you confirm my reading that an explicit `prune: disabled` opt-out should never end up in `survivors`? If you'd rather treat skipped-as-survivor, the change is a one-line swap.

Reviews welcome. Test plan after rebuild verification:

- [x] unit tests for `pruneSurvivors` and `inventory.Merge`
- [x] existing prune / inventory / deletion-policy tests pass locally
- [ ] full local repro with rebuilt image (will run if useful)
- [ ] CI green

### DIFF
--- a/internal/controller/kustomization_controller.go
+++ b/internal/controller/kustomization_controller.go
@@ -498,7 +498,11 @@ func (r *KustomizationReconciler) reconcile(
 	}
 
 	// Run garbage collection for stale resources that do not have pruning disabled.
-	if _, err := r.prune(ctx, resourceManager, obj, revision, originRevision, staleObjects); err != nil {
+	// On failure, re-track the objects whose DELETE wasn't confirmed so that the
+	// next reconcile retries — otherwise status.Inventory advances past them
+	// and they leak as untracked orphans (issue #1664).
+	if _, survivors, err := r.prune(ctx, resourceManager, obj, revision, originRevision, staleObjects); err != nil {
+		inventory.Merge(obj.Status.Inventory, survivors)
 		obj.Status.History.Upsert(checksum, time.Now(), time.Since(reconcileStart), meta.PruneFailedReason, historyMeta)
 		conditions.MarkFalse(obj, meta.ReadyCondition, meta.PruneFailedReason, "%s", err)
 		return err
@@ -1080,31 +1084,69 @@ func (r *KustomizationReconciler) checkHealth(ctx context.Context,
 	return nil
 }
 
+// prune issues delete requests for the given stale objects. In addition to the
+// changed/error return, it returns the slice of objects whose deletion was NOT
+// confirmed by the apiserver (e.g. rejected by an admission webhook). Callers
+// must merge those survivors back into status.Inventory; otherwise the next
+// reconcile's old-vs-new diff will not surface them again and they become
+// untracked orphans labeled with this Kustomization. See issue #1664.
 func (r *KustomizationReconciler) prune(ctx context.Context,
 	manager *ssa.ResourceManager,
 	obj *kustomizev1.Kustomization,
 	revision string,
 	originRevision string,
-	objects []*unstructured.Unstructured) (bool, error) {
+	objects []*unstructured.Unstructured) (bool, []*unstructured.Unstructured, error) {
 	if !obj.Spec.Prune {
-		return false, nil
+		return false, nil, nil
 	}
 
 	log := ctrl.LoggerFrom(ctx)
 
 	changeSet, err := deleteObjects(ctx, obj, manager, objects)
 	if err != nil {
-		return false, err
+		// Identify objects whose DELETE wasn't confirmed (apiserver rejected,
+		// transient error, etc.) so the caller can re-track them. SkippedAction
+		// entries are intentional opt-outs (kustomize.toolkit.fluxcd.io/prune:
+		// disabled) and are treated as settled, otherwise we'd cause an endless
+		// retry of a delete the operator explicitly disabled.
+		return false, pruneSurvivors(objects, changeSet), err
 	}
 
 	// emit event only if the prune operation resulted in changes
 	if changeSet != nil && len(changeSet.Entries) > 0 {
 		log.Info(fmt.Sprintf("garbage collection completed: %s", changeSet.String()))
 		r.event(obj, revision, originRevision, eventv1.EventSeverityInfo, changeSet.String(), nil)
-		return true, nil
+		return true, nil, nil
 	}
 
-	return false, nil
+	return false, nil, nil
+}
+
+// pruneSurvivors returns the subset of objects whose DELETE was not confirmed
+// by the apiserver — i.e. objects without a corresponding DeletedAction or
+// SkippedAction entry in the returned ChangeSet. These should be re-merged into
+// status.Inventory so subsequent reconciles retry their prune. See #1664.
+func pruneSurvivors(objects []*unstructured.Unstructured, changeSet *ssa.ChangeSet) []*unstructured.Unstructured {
+	if len(objects) == 0 {
+		return nil
+	}
+	settled := make(map[string]bool)
+	if changeSet != nil {
+		for _, entry := range changeSet.Entries {
+			switch entry.Action {
+			case ssa.DeletedAction, ssa.SkippedAction:
+				settled[entry.ObjMetadata.String()] = true
+			}
+		}
+	}
+	var survivors []*unstructured.Unstructured
+	for _, obj := range objects {
+		id := object.UnstructuredToObjMetadata(obj).String()
+		if !settled[id] {
+			survivors = append(survivors, obj)
+		}
+	}
+	return survivors
 }
 
 // finalizerShouldDeleteResources determines if resources should be deleted
--- a/internal/controller/kustomization_prune_survivors_test.go
+++ b/internal/controller/kustomization_prune_survivors_test.go
@@ -1,0 +1,87 @@
+/*
+Copyright 2026 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"testing"
+
+	"github.com/fluxcd/cli-utils/pkg/object"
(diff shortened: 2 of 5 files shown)


gecube:
### E2E verification against the original repro

I rebuilt the controller with this PR (`make docker-build IMG=local/kustomize-controller:fix-1664`), loaded it into the kind cluster from https://github.com/gecube/flux-kc-1664-repro, and re-ran the `webhook-deny-delete` scenario that previously reproduced the orphan state 3/3 times on `v1.8.8`. Two checks:

**Check 1 — orphans stay tracked while the webhook blocks DELETE**

After stage C with a `ValidatingAdmissionPolicy` denying DELETE on `app-1a/1b/1c`:

```
$ kubectl -n flux-system get kustomization kc-1664 \
    -o jsonpath='{range .status.inventory.entries[*]}{.id}{"\n"}{end}'
kc-1664_app-flat__ConfigMap
kc-1664_app-1c__ConfigMap
kc-1664_app-1b__ConfigMap
kc-1664_app-1a__ConfigMap
```

The three resources that failed to prune stayed in `status.Inventory` — exactly what this PR promises. (On vanilla `v1.8.8` the inventory would contain only `app-flat` and the per-AZ ConfigMaps would be permanently orphan.)

Status surface is also actionable now:

```
Ready=False  reason=PruneFailed
message: delete failed, errors: ConfigMap/kc-1664/app-1c delete failed:
  configmaps "app-1c" is forbidden: ValidatingAdmissionPolicy
  'kc-1664-block-delete' with binding 'kc-1664-block-delete-binding'
  denied request: kc-1664-block-delete: refuse DELETE for app-1a/1b/1c; …
```

The operator now sees exactly which webhook is blocking which DELETE.

**Check 2 — Flux recovers automatically when the obstruction is removed**

```
$ kubectl delete validatingadmissionpolicybinding kc-1664-block-delete-binding
$ kubectl delete validatingadmissionpolicy kc-1664-block-delete
$ flux reconcile kustomization kc-1664 -n flux-system
✔ applied revision main@sha1:350b6b0a…
$ kubectl -n kc-1664 get cm
NAME       DATA   AGE
app-flat   1      4m
$ kubectl -n flux-system get kustomization kc-1664 \
    -o jsonpath='{range .status.inventory.entries[*]}{.id}{"\n"}{end}'
kc-1664_app-flat__ConfigMap
$ kubectl -n flux-system get kustomization kc-1664 \
    -o jsonpath='Ready={.status.conditions[?(@.type=="Ready")].status}'
Ready=True
```

So the fix doesn't just expose the orphans — it lets the controller converge on its own once the cause is fixed. No manual `kubectl delete` step is required.

I haven't touched the `--with-finalizer` variant (where `deletionTimestamp` was getting set + a controller-owned finalizer blocked completion), since that path was already visible to operators in current Flux releases (`Ready=False` flips, the resource is `Terminating`). The bug surface this PR fixes is the silent one — admission rejections that never even reach the deletion phase.

Ready for review whenever you have a moment.

matheuscscp (approved):
LGTM! 🚀 

@gecube Nice one!

PR #1671, opened 2026-06-16
Enable `StrictPostBuildSubstitutions` by default

matheuscscp:
You always want kustomize-controller to alert you if you forget to add a value for a substitution variable, even if the value must be explicitly set to the empty string. Tests were added here to confirm that variables explicitly set to empty string behave the same as omitted variables: https://github.com/fluxcd/pkg/pull/1246/changes/2cd36cba113a8f5a569eea7285c1e3f3c9458e23


stefanprodan (approved):
LGTM

PR #1672, opened 2026-06-16
Introduce `substituteStrategy: Always`

### LINKED ISSUE #1011: Postbuild substitution should substitute when kustomization have no substitution defined
Postbuild substitution is not working when there is no substitution in kustomization : 

For example, the following kustomization 

```yaml
---
kind: Kustomization
metadata:
  name: my-app
  namespace: flux-system
spec:
  path: ./apps/staging/my-app
  targetNamespace: my-app
  interval: 1m
  prune: true
  sourceRef:
    kind: GitRepository
    name: flux-system
  healthChecks:
    - apiVersion: helm.toolkit.fluxcd.io/v2beta1
      kind: HelmRelease
      name: my-app
      namespace: my-app
```

Generate me the following ressource

```yaml
---
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  labels:
    kustomize.toolkit.fluxcd.io/name: my-app
    kustomize.toolkit.fluxcd.io/namespace: flux-system
  name: my-app
  namespace: my-app
spec:
  chart:
    ...
  values:
    metricsServer:
      dnsPolicy: ${dnsPolicy:=ClusterFirst}
      useHostNetwork: ${hostnetwork:="false"}
...
```

Which result on problem with my release installation
I need to add at least one postBuild argument to make the substitution occurs, even If it is not referred in my ressources
So doing like this :

```yaml
---
kind: Kustomization
metadata:
  name: my-app
  namespace: flux-system
spec:
  path: ./apps/staging/my-app
  targetNamespace: my-app
  interval: 1m
  prune: true
  sourceRef:
    kind: GitRepository
    name: flux-system
  healthChecks:
    - apiVersion: helm.toolkit.fluxcd.io/v2beta1
      kind: HelmRelease
      name: my-app
      namespace: my-app
  postBuild:
    substitute:
      foo: bar
```

will work well :

```yaml
---
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  labels:
    kustomize.toolkit.fluxcd.io/name: my-app
    kustomize.toolkit.fluxcd.io/namespace: flux-system
  name: my-app
  namespace: my-app
spec:
  chart:
    ...
  values:
    metricsServer:
      dnsPolicy: ClusterFirst
      useHostNetwork: "false"
...
```

matheuscscp:
Closes: #1011

Depends on: https://github.com/fluxcd/pkg/pull/1246


api/v1/kustomization_types.go
```diff
@@ -271,9 +271,35 @@ type Decryption struct {
...
 // PostBuild describes which actions to perform on the YAML manifest
 // generated by building the kustomize overlay.
 type PostBuild struct {
+	// SubstituteStrategy defines the strategy for substituting variables in the YAML manifests.
+	// Valid values are:
+	//
+	//  * WithVariables (the default): require at least one variable to be defined,
+	//    either through the inline map or through the resolved references to ConfigMaps
+	//    and Secrets.
+	//  * Always: perform the substitution even if no variables are defined.
```
stefanprodan: Use `-` instead of `*`

matheuscscp: Fixed, thanks!

stefanprodan (approved):
LGTM

PS. We need to make sure `flux build ks` is aware of this new feature

PR #1681, opened 2026-06-23
Add categories to the Kustomization CRD

josesiqueira:
This adds the standard Flux resource categories to the Kustomization CRD in this
repository so users can list Flux resources with kubectl, for example:

```
kubectl get fluxcd          # all Flux resources
kubectl get fluxcd-appliers   # all resources in this category
kubectl get all             # now includes Flux resources
```

Each CRD gets exactly three categories:

```
all
fluxcd
fluxcd-appliers
```

The CRDs were regenerated with `make manifests`.

Part of fluxcd/flux2#5947


matheuscscp (approved):
LGTM! 🚀

PR #1693, opened 2026-07-01
Fix CRD going through variable substitution

### LINKED ISSUE #1692: StrictPostBuildSubstitutions
With https://github.com/fluxcd/kustomize-controller/pull/1671 `StrictPostBuildSubstitutions` is now set to `true` by default, which was released with Kustomization version [v1.9.0](https://github.com/fluxcd/kustomize-controller/blob/v1.9.1/CHANGELOG.md#190) resp. FluxCD version [v2.9.0](https://github.com/fluxcd/flux2/releases/tag/v2.9.0). This causes problems if FluxCD itself is bootstrapped with FluxCD with a Kustomization that has variable substitution enabled. The problem is that the Kubernetes CustomResourceDefinition (CRD) `kustomizations.kustomize.toolkit.fluxcd.io` contains the following `description` field, which contains `${var...`:

```yaml
additionalProperties:
  type: string
description: |-
  Substitute holds a map of key/value pairs.
  The variables defined in your YAML manifests that match any of the keys
  defined in the map will be substituted with the set value.
  Includes support for bash string replacement functions
  e.g. ${var:=default}, ${var:position} and ${var/substring/replacement}.
type: object
``` 

Which then results in the below error:

```
post build failed for 'CustomResourceDefinition.v1.apiextensions.k8s.io/kustomizations.kustomize.toolkit.fluxcd.io': envsubst error: variable substitution failed: variable not set (strict mode): "var"
```

The offending sources are:

- https://github.com/fluxcd/kustomize-controller/blob/eaca23cc2fddd9d9d497e75af3dcedf35ec4a4a7/api/v1/kustomization_types.go#L307
- https://github.com/fluxcd/kustomize-controller/blob/eaca23cc2fddd9d9d497e75af3dcedf35ec4a4a7/api/v1beta1/kustomization_types.go#L179
- https://github.com/fluxcd/kustomize-controller/blob/eaca23cc2fddd9d9d497e75af3dcedf35ec4a4a7/api/v1beta2/kustomization_types.go#L199

# Proposed Workaround or Fixes

- Workaround: Explicitly set `var` to a dummy value to avoid the error
- Fix: Remove the example (`e.g. ${var:=default}, ${var:position} and ${var/substring/replacement}`) from the source/CDR

matheuscscp:
Fixes: #1692

### DIFF
--- a/api/v1/kustomization_types.go
+++ b/api/v1/kustomization_types.go
@@ -454,6 +454,7 @@ func (in *Kustomization) SetConditions(conditions []metav1.Condition) {
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description=""
 // +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=".status.conditions[?(@.type==\"Ready\")].status",description=""
 // +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.conditions[?(@.type==\"Ready\")].message",description=""
+// +kubebuilder:metadata:annotations="kustomize.toolkit.fluxcd.io/substitute=disabled"
 
 // Kustomization is the Schema for the kustomizations API.
 type Kustomization struct {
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

stefanprodan:
@matheuscscp IMO we should do this for all our CRDs. We could do it at distro level in flux2 with a patch also.

matheuscscp:
> We could do it at distro level in flux2 with a patch also.

I think the authoritative place is on the CRDs, as people could be using other things to bootstrap Flux. I will file PRs for the other controllers

PR #1705, opened 2026-07-09
ci: parallelize e2e job to speed it up

### LINKED ISSUE #1678: ci: parallelize e2e job to speed it up
The workflow is now taking 20+ minutes to run. We could easily break it down in 4 jobs to speed the CI up. This would require the basic environment setup, kind cluster etc. to be done on the beginning anyway for all jobs. One of these jobs should only run the controller tests and other static analysis that we run before the actual e2e tests using kind.

Iam-Karan-Suresh:
Closes: https://github.com/fluxcd/kustomize-controller/issues/1678

happy to iterate based on your suggestions @matheuscscp 🙏

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -20,6 +20,11 @@ jobs:
           cache-dependency-path: |
             **/go.sum
             **/go.mod
+      - name: Enable integration tests
+        # Only run integration tests for main branch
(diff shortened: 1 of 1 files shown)


matheuscscp (approved):
LGTM! 🚀 

Thanks very much @Iam-Karan-Suresh!

PR #1718, opened 2026-07-27
Drop controller-runtime dependency from API module

alvaroaleman:
This change removes the usage of the deprecated controller-runtime schemebuilder from the api package and with it the entire controller-runtime dependency.


matheuscscp (approved):
LGTM! 🚀

PR #1720, opened 2026-07-31
Optimize inventory.Diff with map lookup

manoj-1407:
## What does this change do?

`inventory.Diff` used a `versionOf` closure that performed a linear
scan through all inventory entries on every call — once per item in
the diff list. For an inventory with `n` entries and a diff list of
`m` items, this is O(n*m).

This change builds a version map once upfront (O(n)), then uses O(1)
map lookups per diff item, reducing overall complexity to O(n+m).

On clusters with large inventories this avoids redundant linear scans
on every reconciliation cycle where pruning is needed.

## Testing

Existing `Test_Inventory/diff_objects_in_inventory` passes without
modification — behavior is identical, only the lookup strategy changed.


stefanprodan (approved):
LGTM

Thanks @manoj-1407 

PS. Please open a PR with this optimization in [flux-operator](https://github.com/controlplaneio-fluxcd/flux-operator/blob/be7f5449343970407d3abbd43b7caa2ff82dcf32/internal/inventory/inventory.go#L113).

PR #1729, opened 2026-08-31
fix: purge tmp dirs at startup

### LINKED ISSUE #1723: Kustomize-controller leaks /tmp/kustomization-* dirs after ungraceful exit
### Describe the bug
kustomize-controller creates a temp dir per reconcile and removes it with a defer. Deferred functions do not run when the process exits ungracefully, and nothing cleans stale dirs on startup. Any os.Exit or SIGKILL therefore abandons the untarred artifacts of every in-flight reconcile in /tmp, and they persist for the lifetime of the pod.

We found 1.5G of orphaned artifacts sitting in a pod's /tmp five days after the crash that created them.

### What happens
The mechanism, on current `main` (`186fb3b`):

1. A reconcile creates `/tmp/kustomization-<random>` and registers a deferred `os.RemoveAll` ([`internal/controller/kustomization_controller.go#L336-L347`](https://github.com/fluxcd/kustomize-controller/blob/186fb3b29138ce82ff61ba1a9370b96a3dd598fe/internal/controller/kustomization_controller.go#L336-L347)).
2. The controller untars the source artifact into that dir.
3. The process exits without a graceful shutdown, so the deferred cleanup never runs ([`main.go#L389-L392`](https://github.com/fluxcd/kustomize-controller/blob/186fb3b29138ce82ff61ba1a9370b96a3dd598fe/main.go#L389-L392):
   `mgr.Start` returns and `os.Exit(1)` runs, which skips the defers of the reconcile goroutines).
4. The new process does not sweep the temp root, so the dirs stay.

Two exit paths trigger this:

- **Lost leader lease.** controller-runtime v0.24.1 sets the graceful shutdown timeout to zero
  and pushes `leader election lost` ([`pkg/manager/internal.go#L627-L637`](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.24.1/pkg/manager/internal.go#L627-L637)).
- **OOMKill.** The kernel sends SIGKILL. No defer runs.

### Expected behavior

Stale `kustomization-*` dirs left by a previous process are removed when the controller starts, so leaked artifacts do not accumulate across restarts.

### Environment
- kustomize-controller v1.6.1, Flux 2.6.4
- EKS 1.35, Amazon Linux 2023 nodes, cgroup v2
- --concurrent at default (4)
- /tmp mounted as emptyDir: {medium: Memory} rather than the upstream emptyDir: {}. The leak is independent of the medium; see Impact.

### Additional context

Crash logs, lease renewal failing and the manager exiting:

```console
19:03:29 error  Failed to update lock optimistically: Put ".../leases/kustomize-controller-leader-election?timeout=15s": context deadline exceeded, falling back to slow path
19:03:44 error  error retrieving resource lock flux-system/kustomize-controller-leader-election: Get ...: context deadline exceeded
19:03:44 info   failed to renew lease flux-system/kustomize-controller-leader-election: context deadline exceeded
19:03:55 error  Failed to release lock: Put ...: http2: client connection lost
19:03:55 error  problem running manager  {"error": "leader election lost"}
```

The pod's `/tmp` volume, inspected from the node. Four dirs, matching `--concurrent=4`:

```console
# ls -la /var/lib/kubelet/pods/2165bd2d-.../volumes/kubernetes.io~empty-dir/temp
drwxrwsrwt. 6 root   1337 120 Aug 11 16:08 .
drwxr-xr-x. 3 root   root  18 Jul  7 10:15 ..
drwx--S---. 5 nobody 1337 160 Aug 11 16:08 kustomization-1931908412
drwx--S---. 5 nobody 1337 160 Aug 11 16:08 kustomization-252548195
drwx--S---. 5 nobody 1337 160 Aug 11 16:08 kustomization-3947919638
drwx--S---. 5 nobody 1337 160 Aug 11 16:08 kustomization-698526464

# du -sh /var/lib/kubelet/pods/2165bd2d-.../volumes/kubernetes.io~empty-dir/temp
1.5G    /var/lib/kubelet/pods/2165bd2d-.../volumes/kubernetes.io~empty-dir/temp
```

### Impact 

On the upstream default disk-backed `emptyDir`, leaked dirs consume node ephemeral storage until the pod is replaced. 

In our case `/tmp` was `emptyDir: {medium: Memory}`, which turns the same leak into an OOM loop: the pages stay charged to the pod cgroup, so each OOMKill leaves less headroom for the next start.

```console
# findmnt -T .
TARGET   ...volumes/kubernetes.io~empty-dir/temp   SOURCE tmpfs   FSTYPE tmpfs   OPTIONS rw,relatime,seclabel,size=3145728k
```

### Suggested fix

 Sweep leftover `kustomization-*` dirs in the temp root at startup, before the manager begins reconciling. 

Happy to open a PR if you agree with the approach.

stefanprodan:
This PR makes the controller remove the tmp dirs left behind by a previous process that exited without running its deferred cleanup. The dirs are listed synchronously, before the reconcilers start creating new ones, and removed in the background with a timeout of 2m so that the manager startup is not delayed.

Fixes: #1723
Supersedes: #1726

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,6 +34,7 @@ require (
 	github.com/fluxcd/pkg/testserver v0.14.0
 	github.com/fluxcd/source-controller/api v1.9.0
 	github.com/getsops/sops/v3 v3.13.2
+	github.com/go-logr/logr v1.4.3
 	github.com/google/cel-go v0.26.1
 	github.com/hashicorp/vault/api v1.23.0
 	github.com/onsi/gomega v1.42.1
@@ -142,7 +143,6 @@ require (
 	github.com/getsops/gopgagent v0.0.0-20241224165529-7044f28e491e // indirect
 	github.com/go-errors/errors v1.5.1 // indirect
 	github.com/go-jose/go-jose/v4 v4.1.4 // indirect
-	github.com/go-logr/logr v1.4.3 // indirect
 	github.com/go-logr/stdr v1.2.2 // indirect
 	github.com/go-logr/zapr v1.3.0 // indirect
 	github.com/go-openapi/jsonpointer v0.21.1 // indirect
--- a/internal/controller/kustomization_controller.go
+++ b/internal/controller/kustomization_controller.go
@@ -335,7 +335,7 @@ func (r *KustomizationReconciler) reconcile(
 	}
 
 	// Create tmp dir.
-	tmpDir, err := MkdirTempAbs("", "kustomization-")
+	tmpDir, err := MkdirTempAbs("", TempDirPrefix)
 	if err != nil {
 		err = fmt.Errorf("tmp dir error: %w", err)
 		conditions.MarkFalse(obj, meta.ReadyCondition, sourcev1.DirCreationFailedReason, "%s", err)
--- a/internal/controller/tmpdir.go
+++ b/internal/controller/tmpdir.go
@@ -1,0 +1,71 @@
+/*
+Copyright 2026 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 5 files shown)


matheuscscp (approved):
LGTM

PR #1734, opened 2026-09-01
fix: requeue at dependency interval for missing object refs

Bisman-Singh:
Part of fluxcd/flux2#5809

When the referenced Source object does not exist, the controller requeued at the retry interval; when a non-optional `substituteFrom` ConfigMap/Secret does not exist, the failure fell into the generic reconcile-error path, also requeueing at the retry interval. Both are inconsistent with how not ready dependencies, nil source artifacts and missing artifact files are handled, all of which requeue at the `--requeue-dependency` interval, and they delay bootstrap ordering when the referenced objects are created shortly after the Kustomization (e.g. via a ResourceSet).

This change:

- requeues at `--requeue-dependency` when the source object is not found;
- tags build errors caused by a missing referenced Kubernetes object (e.g. `substituteFrom` ConfigMap/Secret) with a sentinel `errBuildRefNotFound` and requeues them at `--requeue-dependency`, alongside the existing `fetch.ErrFileNotFound` handling. The Ready=False condition with `BuildFailedReason` and the full error message is still recorded.

The check is deliberately scoped to the build step rather than a blanket `apierrors.IsNotFound(reconcileErr)`: `fluxcd/pkg/ssa` wraps apply errors with `%w`, so a blanket check would also reroute apply-time NotFound failures (e.g. a namespace that does not exist) away from their error events.

Tests: new `TestKustomizationReconciler_RequeueOnMissingRefs` asserts the returned `RequeueAfter` equals the dependency interval for both scenarios; both subtests were verified to fail against the unfixed code (they observed the 10m retry interval).

Signed-off-by: Bisman-Singh <bismanmadaan1@gmail.com>

### DIFF
--- a/internal/controller/kustomization_controller.go
+++ b/internal/controller/kustomization_controller.go
@@ -83,6 +83,11 @@ import (
 // +kubebuilder:rbac:groups="",resources=serviceaccounts/token,verbs=create
 // +kubebuilder:rbac:groups="",resources=events,verbs=create;patch
 
+// errBuildRefNotFound is returned when a Kubernetes object referenced in
+// the build process of a Kustomization, e.g. a substituteFrom ConfigMap or
+// Secret, does not exist.
+var errBuildRefNotFound = errors.New("build reference not found")
+
 // KustomizationReconciler reconciles a Kustomization object
 type KustomizationReconciler struct {
 	client.Client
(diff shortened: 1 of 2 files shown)


Bisman-Singh:
Flagging a scope question raised on fluxcd/flux2#5809: unlike helm-controller, this controller has no exponential backoff on these paths. Every reconcile failure already returns `ctrl.Result{RequeueAfter: obj.GetRetryInterval()}`, and the source-not-found branch used `obj.GetRetryInterval()` before this change too, so this PR moves a user-tunable `retryInterval` onto the global `--requeue-dependency` flag rather than fixing a backoff problem. Holding this one for @matheuscscp's call ([discussion](https://github.com/fluxcd/flux2/issues/5809#issuecomment-5492095581)) — happy to close it or cut it down to just the `substituteFrom` case. fluxcd/helm-controller#1577 is unaffected by this question.

Bisman-Singh:
Closing per the discussion on fluxcd/flux2#5809.

PR #1741, opened 2026-09-11
Extend SOPS redaction to postBuild substitutions

dipti-pai:
postBuild.substituteFrom values are substituted verbatim into manifests before server-side apply. When the resulting value causes a Kubernetes API validation error that echoes the value back (e.g. a label value over the 63 byte limit), the raw error was copied unmodified into the Kustomization status conditions and events.

Load the postBuild.substituteFrom variables once per reconcile, using the same loader generator.SubstituteVariables relies on internally, and mask their values out of apply errors before they are attached to the Ready condition or emitted as an event. This extends the existing SOPS decryption error redaction (safeDecrypt) to also cover postBuild substitutions.

Assisted-by: GitHub Copilot/Claude Sonnet 5


internal/controller/kustomization_controller.go
```diff
@@ -433,6 +434,15 @@ func (r *KustomizationReconciler) reconcile(
...
 	}
 
+	// Load the postBuild.substituteFrom variables once (this is the same
+	// loader that generator.SubstituteVariables uses internally in
+	// r.build()), so their values can be redacted from any apply error
+	// surfaced below. This extends the existing SOPS decryption redaction
+	// (see safeDecrypt) to also cover postBuild substitutions. Errors are
+	// ignored here: a failure to load is surfaced again, and handled, by
+	// r.build() below.
+	substituteVars, _ := generator.LoadVariables(ctx, r.Client, unstructured.Unstructured{Object: k})
```
matheuscscp: Loading twice, then?

Karthik-Chowdary (commented):
I reviewed the redaction path against the substitution load/build sequencing. The separate best-effort load introduces a TOCTOU gap that can still expose the value this change intends to protect; I left details inline.

matheuscscp:
@dipti-pai Looks like the agent still implemented this in a way that the data is being loaded twice

dipti-pai:
> @dipti-pai Looks like the agent still implemented this in a way that the data is being loaded twice

Hi @matheuscscp. Had to update pkg code to avoid the additional call to `LoadVariables` from kustomize-controller. The associated PR is here - https://github.com/fluxcd/pkg/pull/1298 . Temporarily updated the go.mod to point to the pkg fork. 

Could you take a look? Thanks.

matheuscscp:
Thanks @dipti-pai, I merged pkg 1298. I will cut a `kustomize` release together with all the other pkgs after pkg 1144 is also merged

dipti-pai:
> Thanks @dipti-pai, I merged pkg 1298. I will cut a `kustomize` release together with all the other pkgs after pkg 1144 is also merged

Thanks @matheuscscp : )

